PR #8, opened 2015-01-25
README.rst: Code blocks

msabramo:
Will result in color syntax highlighting on PyPI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,9 @@ What?
 =====
 
 ``smart_open`` is a Python library for **efficient streaming of very large files** from/to S3, HDFS or local (compressed) files.
-It is well tested (using `moto <https://github.com/spulec/moto>`_), well documented and sports a simple, Pythonic API::
+It is well tested (using `moto <https://github.com/spulec/moto>`_), well documented and sports a simple, Pythonic API:
+
+.. code-block:: python
 
   >>> # stream lines from an S3 object
   >>> for line in smart_open.smart_open('s3://mybucket/mykey.txt'):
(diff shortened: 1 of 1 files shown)


piskvorky:
Thanks!

Is this standard .rst, would it work with e.g. Sphinx / other tools?

msabramo:
Yeah so docutils + Pygments supports it and will render it colored. 

GitHub will show it without color today but my PR has already been merged to make it work. They just need to deploy the change. 

I can't think of any downside.

piskvorky:
I just [tested it](https://github.com/piskvorky/smart_open/tree/rst_test) and GitHub seems to support `.. code-block:: python` (not `pycon`).

Would `python` work on PyPI too?

msabramo:
Yes. The `pycon` one is slightly better because it renders the `>>>` a little differently. However it doesn't work on GitHub yet, whereas the `python` one does. If you're impatient, you could use the `python` one and it will work pretty well on both PyPI and GitHub today.

https://github.com/atom/language-python/pull/48 is the PR I referenced before. I can't wait for Github to deploy it!

msabramo:
I fixed the merge conflict and updated it to use `python`. It looks pretty good I think: https://github.com/msabramo/smart_open/tree/patch-2

piskvorky:
Excellent, much thanks :)

piskvorky:
Made a new release, the [PyPI page](https://pypi.python.org/pypi/smart_open) looks much better indeed, thanks again Marc!

msabramo:
You're welcome! Glad to help!

PR #10, opened 2015-02-01
faster travis builds using their container infra

theanalyst:
Using containers[1] & caching the dependant packages (using wheels) so that subsequent
travis builds cache this.

As a result was able to bring down the build time to about 52 seconds[2]

[1] http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/ 
[2] https://travis-ci.org/theanalyst/smart_open/builds/49096656

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,18 @@
+sudo: false
 language: python
 notifications:
   email: false
 python:
 - "2.7"
 - "2.6"
+cache:
+  directories:
+    - $PWD/wheelhouse  # cache the dependencies
+env:
+  global:
+    - export PIP_FIND_LINKS=$PWD/wheelhouse
 install:
-- pip install mock moto
-- python setup.py install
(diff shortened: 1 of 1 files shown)


piskvorky:
Thanks, but does 71s (before) vs. 52s (now) spent in Travis really make a difference? Can you give more motivation for this optimization?

I'm not a big fan of fixing things that are not broken. Especially when the fix brings in more complexity (as caching invariably does).

Not to sound ungrateful but I'm afraid I'm -1 on this one. I'll leave the decision to @VincTheSecond.

theanalyst:
Yeah atm doesn't seem worth it.. feel free to close this.

piskvorky:
Ping @VincTheSecond : either merge or close, don't leave PRs stale.

VincTheSecond:
Thank you for your proposal, but for this small project it is not really necessary. Let's keep things as simple as possible.

PR #14, opened 2015-02-26
 Auto create folder , if the file's folder is not exit.

texib:
When use s3 opener , that AWS S3 will auto create file's folder.
So i add the mechanism  at smart_ope when use local file system


piskvorky:
Thanks @texib , but I don't think that's a good idea. We try to follow the principle of least surprise = keep "local filesystem" functionality as close to the original behaviour as possible.

Also, I think that boto will try to get the AWS credentials from environment automatically. No need to pull them explicitly from smart_open -- or does that not work for you?

PR #16, opened 2015-04-09
Debugging weird Travis fails on 3.4

piskvorky:
Tests work fine when run locally on py3.4. But they stall (hang) at random places when run on Travis. Travis reports a failure because it times out after 10 minutes. 

This only happens for Python 3.4 (2.6 and 2.7 and 3.3 work, even on Travis).
 WTF?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ install:
 - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install ordereddict; fi
 - pip install mock moto
 - python setup.py install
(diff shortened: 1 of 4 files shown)


ziky90:
It seems that the problem is related to https://github.com/boto/boto/issues/3011, unfortunately updated version of boto in setup.py did not help.

piskvorky:
Good catch! Maybe we should specify the required version of HTTPretty in `setup.py`, rather than just `.travis.yml`?

Let's keep an eye on gabrielfalcao/HTTPretty#228 .

PR #19, opened 2015-04-24
fixed boto and httpretty dependencies for travis

ziky90:
This PR simply summarises and repairs errors that were found in the #16 .
It was needed to switch for the newest version of boto and force travis to test with httpretty==0.8.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ python:
 - "2.7"
 - "2.6"
 - "3.3"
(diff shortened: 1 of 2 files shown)


piskvorky:
Can we merge? Or anything else missing?

ziky90:
I've just moved the `httpretty==0.8.6` to setup.py and I'm waiting for results of travis tests.
Then we can merge.

PR #20, opened 2015-06-04
- smart_open will try to re-upload the last chunk that caused an except…

andreycizov:
…ion when trying to close a connection when exiting S3OpenWrite context

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -378,13 +378,20 @@ class S3OpenWrite(object):
     def __enter__(self):
         return self
 
+    def _termination_error(self):
+        logger.exception("encountered error while terminating multipart upload; attempting cancel")
+        self.outbucket.cancel_multipart_upload(self.mp.key_name, self.mp.id)
+        logger.info("cancel completed")
+
     def __exit__(self, type, value, traceback):
+        if type is not None:
+            self._termination_error()
+            return False
+        
         try:
             self.close()
         except:
-            logger.exception("encountered error while terminating multipart upload; attempting cancel")
-            self.outbucket.cancel_multipart_upload(self.mp.key_name, self.mp.id)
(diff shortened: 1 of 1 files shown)


piskvorky:
Thanks!

nvie:
@piskvorky I was relying on this old behaviour where the log stream was still written out _up until_ the moment the unexpected exception occurs. At least this gives you a clue when debugging. What is the desired behaviour here?

Was this patch indeed deliberately removing this behaviour and indeed cancelling the entire upload when an uncaught exception happens inside this context? To me this means a job emitting logs for over 10 minutes and then suddenly hitting an uncaught exception means I lose all the logs.

If this was not intentional, you should probably revert this commit.

andreycizov:
This should be fixed in some other way. In my case getting an unexpected
exception most certainly implied that the upload could not be complete in
the first place and trying to retry it would fail it any way. Basically it
would try it again and fail anyway, spending double the time to detect the
problem.

I get the context of your problem, but my guess it only worked because the
exception gets a single retry on re-uploading the last piece of file and my
reasoning behind this commit is that this should at least be generalised to
any amount of possible retries for a given upload.

piskvorky:
@nvie but it was NOT written out... unless the multipart upload is completed, the parts uploaded so far are neither here nor there... just dangling in the space...

And by the way, will S3 still charge you for this "dangling data", even if the objects are not normally visible. You'd have to abort the multipart upload manually.

I guess the question here is, what is the intended/best behaviour?

nvie:
I'm sure the following used to work for me:

```python
with smart_open('s3://...', 'w') as f:
    f.write('Hello world!\n')
    1/0
```

This throws a ZeroDivisionError, but the log was written to S3 and contained the line. Is this a special case because the output is relatively small so the multipart may only consist of a single chunk?

PR #22, opened 2015-06-17
Return the key in s3_iter_bucket

salilb:
Returns the entire key instead of only the key name. Breaks compatibility with older code!!!

### DIFF
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -480,7 +480,10 @@ class S3IterBucketTest(unittest.TestCase):
         self.assertEqual(len(result), min(len(expected), 10))
 
         for workers in [1, 4, 8, 16, 64]:
-            self.assertEqual(dict(smart_open.s3_iter_bucket(mybucket, workers=workers)), expected)
+            result = {}
+            for k, c in smart_open.s3_iter_bucket(mybucket):
+                result[k.name] = c
+            self.assertEqual(result, expected)
 
 
 PY2 = sys.version_info[0] == 2
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -465,14 +465,12 @@ class S3IterBucketTest(unittest.TestCase):
         result = {}
         for k, c in smart_open.s3_iter_bucket(mybucket):
             result[k.name] = c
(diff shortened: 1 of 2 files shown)


piskvorky:
Travis build fails -- will require a change to unit tests as well, I suppose.

salilb:
Do you want to make changes to the unit tests? Else, I'll try to fix it later.

Could returning the whole key have an impact on the performance?

piskvorky:
Is the key object pickleable (can be `pickle.dump()`ed)?

salilb:
I fixed the tests. Not the most elegant solution but it passes the tests.

smart_open/tests/test_smart_open.py
```diff
@@ -462,11 +462,17 @@ def test_s3_iter_bucket_moto(self):
                 expected[key_name] = content
 
         # read all keys + their content back, in parallel, using s3_iter_bucket
-        result = dict(smart_open.s3_iter_bucket(mybucket))
+        result = {}
+        for k, c in smart_open.s3_iter_bucket(mybucket):
+            result[k.name] = c
+        #result = dict(smart_open.s3_iter_bucket(mybucket))
```
piskvorky: We don't want to keep dead code around, even as comments. The line is preserved in git history, if we ever need it again.

smart_open/tests/test_smart_open.py
```diff
@@ -462,11 +462,17 @@ def test_s3_iter_bucket_moto(self):
...
         self.assertEqual(expected, result)
 
         # read some of the keys back, in parallel, using s3_iter_bucket
-        result = dict(smart_open.s3_iter_bucket(mybucket, accept_key=lambda fname: fname.endswith('4')))
+        result = {}
+        for k, c in smart_open.s3_iter_bucket(mybucket, accept_key=lambda fname: fname.endswith('4')):
+            result[k.name] = c
+        #result = dict(smart_open.s3_iter_bucket(mybucket, accept_key=lambda fname: fname.endswith('4')))
```
piskvorky: ditto

piskvorky:
Looks elegant enough to me :) Please fix the dead code and we can merge :+1:

salilb:
done! :)

piskvorky:
Merged, thanks @salilb !

PR #25, opened 2015-07-14
Hdfs __iter__ implementation fix

### LINKED ISSUE #24: Bug (maybe feature) in HdfsOpenRead
On line 249 I have noticed a minor bug in reading from HDFS

There is used command `hadoop fs -cat` which based on the documentation http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/FileSystemShell.html reads any file mounted to Hadoop (HDFS, FS, HFTP FS, S3 FS, and others)

Correctly to meet what smart_open claims to do there should be used `hdfs dfs -cat`.
http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/HDFSCommands.html#dfs

I can do a PR with fix if this is really a bug.

If this is expected behaviour then I would change the documentation + I would not limit files schema to only HDFS.

ziky90:
Fixed HDFS read to be able to read both `hdfs://` and `hdfs:///` (both should be equivalent and it depends on hdfs distribution)

Also fixed to use `hdfs dfs -cat` which is recommended.

Closes #24

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -246,7 +246,7 @@ class HdfsOpenRead(object):
         self.parsed_uri = parsed_uri
 
     def __iter__(self):
-        hdfs = subprocess.Popen(["hadoop", "fs", "-cat", self.parsed_uri.uri_path], stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


piskvorky:
Merging. Thanks @ziky90 !

PR #28, opened 2015-08-12
Specify encoding when reading file. Otherwise fails if default encoding is set, for instance, to cp1251.

kalaidin:
Python >=2.6 friendly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@
 # from the MIT License (MIT).
 
 
+import io
 import os
 import sys
 
@@ -18,7 +19,7 @@ if sys.version_info < (2, 6):
 from setuptools import setup, find_packages
 
 def read(fname):
-    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+    return io.open(os.path.join(os.path.dirname(__file__), fname), encoding='utf-8').read()
 
 
 setup(


piskvorky:
Thanks, but what exactly fails?

`read()` is a general method returning a binary string, I don't see how encoding (text) comes into it?

setup.py
```diff
@@ -18,7 +19,7 @@
 from setuptools import setup, find_packages
 
 def read(fname):
-    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+    return io.open(os.path.join(os.path.dirname(__file__), fname), encoding='utf-8').read()
```
piskvorky: standard way to do this would be with `codecs.open` module; why `io`?

kalaidin: Been thinking about codecs.open too, but then came across:
> While the builtin open() and the associated io module are the recommended approach for working with encoded text files, this module provides additional utility functions and classes that allow the use of a wider range of codecs when working with binary files
https://docs.python.org/3/library/codecs.html#codecs.open

Anyway, I can send an update with codecs if needed - please let me know.

piskvorky: OK. I don't know much about the `io` module, but as long as all tests pass and it solves your cp1251 problems, it's good enough :)

piskvorky:
Merging, thanks @kalaidin !

kalaidin:
Thanks much!

PR #31, opened 2015-08-26
Added writing to webhdfs support

ziky90:
This PR adds ability to write directly to webhdfs.

**NOTE:**
Please do not merge yet

**TODO:**
 - tests
 - ~~investigation of `close`,` __enter__` and `__exit__` methods~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,6 +45,11 @@ It is well tested (using `moto <https://github.com/spulec/moto>`_), well documen
   ...     for line in ['first line', 'second line', 'third line']:
   ...          fout.write(line + '\n')
 
+  >>> # stream content *into* WebHDFS (write mode):
+  >>> with smart_open.smart_open('webhdfs://host:port/user/hadoop/my_file.txt', 'wb') as fout:
+  ...     for line in ['first line', 'second line', 'third line']:
+  ...          fout.write(line + '\n')
+
   >>> # stream from/to local compressed files:
   >>> for line in smart_open.smart_open('./foo.txt.gz'):
   ...    print line
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -27,7 +27,10 @@ import os
 import subprocess
 import sys
 import requests
-import httplib
+if sys.version_info[0] == 2:
+    import httplib
+elif sys.version_info[0] == 3:
+    import http.client as httplib
 
 from boto.compat import BytesIO, urlsplit, six
 import boto.s3.key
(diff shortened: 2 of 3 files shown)


ziky90:
Implementation of writing to WebHDFS is basically done. I'll get back to tests probably at the end of the next week.

I have question regarding cached writing. System of writing to WebHDFS is quiet comparable to writing to S3, but caching is not necessary here. Should I add there some caching mechanism or should I let this on user?
Without caching, writing of one line == one http request.

piskvorky:
I don't know of anyone else requesting WebHDFS, so your use case is the only one we have and care about for now. So... whatever works for you, I guess.

One HTTP request per line sounds pretty horrible... but if it works for you, it must cover at least one use case, so it cannot be completely useless. There's always room for improvement later.

piskvorky:
@ziky90 Let me know when you're ready to merge -- I'll make a new release after that.

ziky90:
I'd like to do caching for writing (one request per line does not fit our performance requirements) and some unit tests before merging. I'm just not sure if I'll be able to make it during this week.

val314159:
it'd be nice to generalize this.  if you could do it by block, you could
actually mount this thing as a filesystem.

anyway, just thinking out loud :)

ziky90:
@val314159 I think, that I don't fully understand your idea. Do you mean the `buffersize` parameter from here?

val314159:
Well, my idea was only half-formed, I didn't even look at the underlying
source, but i'm just saying this could be the core of a mounted file
system.  But my experience is that network filesystem are mostly terrible
(unless you have a guaranteed fast network) unless you have a really good
caching layer.

(TL;DR caching is cool / block caches allow for filesystem mounts / don't
worry about it too much)

ziky90:
I'm kind of stacked with writing test.
Unfortunately I have not found some any elegant solution for mocking the webhdfs (for reading there was simple workaround using just responses library, but it's getting more complicated for writing).

piskvorky:
@ziky90 we'll release `smart_open` now (today/tmr). The "read WebHDFS support is in"; we can add this "write support" in the next release.

ziky90:
I'm working on it right now and I'd like to finish it today.

piskvorky:
OK, so we'll wait :)

ziky90:
Some, at least minimal necessary tests were added, so from my point of view it is ready to merge, but additional tests for webhdfs (in general) are welcomed and might be also added later on.

piskvorky:
All right. Tests pass, and I'm assuming this works for you, so let's merge.

Thanks @ziky90 !

PR #32, opened 2015-09-16
make smart_open a no-op if passed a file-like

gojomo:
If rather than a string URI/path, `smart_open()` receives an object with a `read` attribute, pass it through: it's arguably already 'open'.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
         'boto >= 2.32',
         'httpretty==0.8.6',
         'bz2file',
+        'requests',
     ],
 
     test_suite="smart_open.tests",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -47,6 +47,7 @@ def read(fname):
         'boto >= 2.32',
         'httpretty==0.8.6',
         'bz2file',
+        'requests',
```
piskvorky: What is this for? I can't see any such dependency in `smart_open`.

gojomo: https://github.com/piskvorky/smart_open/blob/master/smart_open/smart_open_lib.py#L29 & multiple subsequent references?

piskvorky: Oh I see, it's one of the recent additions by @ziky90 , I didn't have that pulled locally yet. Cheers!

Issue #35, opened 2015-11-04
Using a proxy to access S3


asieira:
The boto S3 connection allows `proxy`, `proxy_port`, `proxy_user` and `proxy_pass` to be set explicitly. It would be nice to have a way to propagate those options from `smart_open` to the opening of the connection.

Maybe the best way to go about this is to allow the boto connection object to be passed to `smart_open`, instead of a new one being created each time.

piskvorky:
Hmm, couldn't we do this using by extending the functionality added in #30 ?

asieira:
Not sure if I understand what you have in mind, but I don't think we can use the same argument here. 

On S3OpenWrite, the current usage for `**kw` is to pass it along to `initiate_multipart_upload`, not `connect_s3`: https://github.com/piskvorky/smart_open/blob/master/smart_open/smart_open_lib.py#L402

We could, however, add a new argument to `smart_open` that is a dict with keyword arguments to be passed along to `connect_s3` in a similar way.

asieira:
Ok, I though of something simpler. How about `smart_open` properly handles the case where `uri` is actually an instance of `boto.s3.key.Key` and uses its data instead of creating new connection and bucket objects?

piskvorky:
Hmm, interesting idea. As long as it's backward compatible (strings work the same way), we could "add" support for other types of input, why not.

Any comments @ziky90 @VincTheSecond ?

ziky90:
I like the idea. It would be useful for me as well to be able to use `proxy` parameter.

asieira:
The challenge is that for this to work, I would change the constructors for both `S3OpenRead` and `S3OpenWrite` to receive a `boto.s3.key.Key` instance, which would not be strictly backwards compatible if anyone was using them directly.

asieira:
I just submitted a pull request with a working version of this idea. I'll still need some help fixing one of the tests, which I temporarily commented out.

tmylk:
Merged in #38

PR #38, opened 2015-11-05
Allowing boto keys to be passed directly to smart_open (#35)

asieira:
* Updated `smart_open` to accept an instance of boto.s3.key.Key as the
`uri` parameter.
* Refactored `S3OpenRead` and `S3OpenWrite` to require only a
boto.s3.key.Key instance to work, and get the bucket directly from its
properties. *This is not backwards compatible*.
* Commented out test `SmartOpenReadTest.test_s3_boto` temporarily.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -52,16 +52,23 @@ def smart_open(uri, mode="rb", **kw):
 
     The `uri` can be either:
 
-    1. local filesystem (compressed ``.gz`` or ``.bz2`` files handled automatically):
+    1. a URI for the local filesystem (compressed ``.gz`` or ``.bz2`` files handled automatically):
        `./lines.txt`, `/home/joe/lines.txt.gz`, `file:///home/joe/lines.txt.bz2`
-    2. Amazon's S3 (can also supply credentials inside the URI):
+    2. a URI for HDFS: `hdfs:///some/path/lines.txt`
+    3. a URI for Amazon's S3 (can also supply credentials inside the URI):
        `s3://my_bucket/lines.txt`, `s3://my_aws_key_id:key_secret@my_bucket/lines.txt`
-    3. HDFS: `hdfs:///some/path/lines.txt`
+    4. an instance of the boto.s3.key.Key class.
 
     Examples::
 
       >>> # stream lines from S3; you can use context managers too:
       >>> with smart_open.smart_open('s3://mybucket/mykey.txt') as fin:
+      ...     for line in fin:
+      ...         print line
+
+      >>> # you can also use a boto.s3.key.Key instance directly:
+      >>> key = boto.connect_s3().get_bucket("my_bucket").get_key("my_key")
+      >>> with smart_open.smart_open(key) as fin:
       ...     for line in fin:
       ...         print line
 
@@ -221,7 +228,8 @@ class S3OpenRead(object):
 
     """
     def __init__(self, read_key):
-        if not isinstance(read_key, boto.s3.key.Key):
+        if not hasattr(read_key, "bucket") and not hasattr(read_key, "name") and not hasattr(read_key, "read") \
+                and not hasattr(read_key, "close"):
             raise TypeError("can only process S3 keys")
         self.read_key = read_key
         self.line_generator = s3_iter_lines(self.read_key)
@@ -405,6 +413,9 @@ class S3OpenWrite(object):
         is 5MB.
 
         """
+        if not hasattr(outkey, "bucket") and not hasattr(outkey, "name"):
+            raise TypeError("can only process S3 keys")
+
         self.outkey = outkey
         self.min_part_size = min_part_size
 
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -124,6 +124,8 @@ def smart_open(uri, mode="rb", **kw):
                 return WebHdfsOpenRead(parsed_uri, **kw)
             else:
                 return WebHdfsOpenWrite(parsed_uri, **kw)
+        else:
+            raise NotImplementedError("scheme %r is not supported", parsed_uri.scheme)
     elif isinstance(uri, boto.s3.key.Key):
         # handle case where we are given an S3 key directly
         if mode in ('r', 'rb'):
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -85,41 +85,56 @@ def smart_open(uri, mode="rb", **kw):
       ...    fout.write("good bye!\n")
 
     """
-    # simply pass-through if already a file-like
-    if not isinstance(uri, six.string_types) and hasattr(uri, 'read'):
-        return uri
 
-    # this method just routes the request to classes handling the specific storage
-    # schemes, depending on the URI protocol in `uri`
-    parsed_uri = ParseUri(uri)
+    # validate mode parameter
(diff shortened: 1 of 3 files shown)


asieira:
I commented out one of the tests since I'm not very familiar with moto and I'm probably doing something wrong. This is the output I get:

```
Error
  File "/Users/asieira/.virtualenvs/smart_open/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/Users/asieira/Repositorios/smart_open/smart_open/tests/test_smart_open.py", line 139, in test_s3_boto
    smart_open_object = smart_open.smart_open("s3://mybucket/mykey")
  File "/Users/asieira/Repositorios/smart_open/smart_open/smart_open_lib.py", line 111, in smart_open
    return S3OpenRead(key, **kw)
  File "/Users/asieira/Repositorios/smart_open/smart_open/smart_open_lib.py", line 222, in __init__
    if not isinstance(read_key, boto.s3.key.Key):
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types
```

piskvorky:
Thanks @asieira !

That's a weird error, not sure what's going on. Will have to check in more detail.

asieira:
:bow:

smart_open/tests/test_smart_open.py
```diff
@@ -131,26 +131,26 @@ def test_webhdfs_read(self):
...
-        smart_open_object = smart_open.S3OpenRead(smart_open.ParseUri("s3://access_id:access_secret@mybucket/mykey"))
-        smart_open_object.__iter__()
-        mock_boto.connect_s3.assert_called_with(aws_access_key_id="access_id", aws_secret_access_key="access_secret")
-
-        # lookup bucket, key; call s3_iter_lines
-        smart_open_object = smart_open.S3OpenRead(smart_open.ParseUri("s3://access_id:access_secret@mybucket/mykey"))
-        smart_open_object.__iter__()
-        mock_boto.connect_s3().get_bucket.assert_called_with("mybucket")
-        mock_boto.connect_s3().get_bucket().lookup.assert_called_with("mykey")
-        self.assertTrue(mock_s3_iter_lines.called)
```
tmylk: Please uncomment the test as we need to merge them in.

smart_open/tests/test_smart_open.py
```diff
@@ -131,26 +131,26 @@ def test_webhdfs_read(self):
...
+    #     # with credential
+    #     smart_open_object = smart_open.smart_open("s3://access_id:access_secret@mybucket/mykey")
+    #     smart_open_object.__iter__()
+    #     mock_boto.connect_s3.assert_called_with(aws_access_key_id="access_id", aws_secret_access_key="access_secret")
+    #
+    #     # lookup bucket, key; call s3_iter_lines
+    #     smart_open_object = smart_open.smart_open("s3://access_id:access_secret@mybucket/mykey")
+    #     smart_open_object.__iter__()
+    #     mock_boto.connect_s3().get_bucket.assert_called_with("mybucket")
+    #     mock_boto.connect_s3().get_bucket().lookup.assert_called_with("mykey")
```
tmylk: Please update the test as you are using get_key instead of lookup 
lookup is deprecated and you switched to get_key now.

smart_open/smart_open_lib.py
```diff
@@ -203,25 +220,16 @@ class S3OpenRead(object):
...
-            raise TypeError("can only process S3 files")
-        self.parsed_uri = parsed_uri
-        s3_connection = boto.connect_s3(
-            aws_access_key_id=parsed_uri.access_id,
-            aws_secret_access_key=parsed_uri.access_secret)
-        self.read_key = s3_connection.get_bucket(parsed_uri.bucket_id).lookup(parsed_uri.key_id)
-        if self.read_key is None:
-            raise KeyError(parsed_uri.key_id)
+    def __init__(self, read_key):
+        if not isinstance(read_key, boto.s3.key.Key):
```
tmylk: This check is redundant as you have already checked the type in the constructor. It needs to be removed.
If you remove it then the error in test_s3_boto will disappear. Mock library is designed for duck-typing so this is why this fails.

smart_open/smart_open_lib.py
```diff
@@ -85,41 +85,58 @@ def smart_open(uri, mode="rb", **kw):
       ...    fout.write("good bye!\n")
```
tmylk: Add the new boto key usage example to the docstring

smart_open/smart_open_lib.py
```diff
@@ -85,41 +85,58 @@ def smart_open(uri, mode="rb", **kw):
...
-            raise NotImplementedError("read mode not supported for %r scheme", parsed_uri.scheme)
-    elif mode in ('w', 'wb'):
-        if parsed_uri.scheme in ("s3", "s3n"):
+    # validate mode parameter
+    if not isinstance(mode, six.string_types):
+        raise TypeError('mode should be a string')
+    if not mode in ('r', 'rb', 'w', 'wb'):
+        raise NotImplementedError('unknown file mode %s' % mode)
+
+    if isinstance(uri, six.string_types):
```
tmylk: Rename uri input parameter into uri_or_key as it can be either now

tmylk:
@asieira 

See comments in code. Refer to #45 for examples.

Please add the change description to the CHANGELOG and then it will be ready to merge

smart_open/tests/test_smart_open.py
asieira: Nice catch, thank you!

smart_open/smart_open_lib.py
asieira: I added this because `S3OpenRead` is a class that could be used directly by users of the library, so I don't think we can assume the type checks in `smart_open` will have been performed already. Isn't there a way to check if this was an instance of `boto.s3.key.Key` or the corresponding mock library class?

smart_open/smart_open_lib.py
asieira: Done.

smart_open/smart_open_lib.py
asieira: That could create a rather serious backwards compatibility problem, wouldn't it? If any existing code is calling `smart_open(uri='...')` this renaming will stop it working. So however logically consistent your suggestion is, unless we do a major version bump for `smart_open` I would recommend against that.

What do you think?

smart_open/smart_open_lib.py
tmylk: The mock object is of class MagicMock so isinstance won't work.
For Python duck typing you can use hasattrs to check if read_key has name,close and other methods that are actuallyused.

asieira: Ok, so I added the `hasattr` tests and now the tests are passing. Thank you!

smart_open/smart_open_lib.py
tmylk: Sure!

tmylk:
@asieira Thanks for the quick response. Waiting for the commit - going to release today.

asieira:
There you go, @tmylk !

asieira:
I see the new version was released as 1.3.1. 

I would suggest that in the future the project follows the [semantic versioning](http://semver.org) guidelines, which is what most developers expect these days. Under these guidelines, since new functionality was introduced by my PR, I believe the release version would have been 1.4.0.

No crisis, just a suggestion for future releases. :) Thanks again for the wonderful project!

PR #39, opened 2015-11-05
disable multiprocessing if unavailable

jsphpl:
I hacked something up, please check if you're okay with it. Couldn't get the tests running, so this code was tested only manually – not sure if these code changes require the tests to be updated.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -22,7 +22,6 @@ The main methods are:
 """
 
 import logging
-import multiprocessing.pool
 import os
 import subprocess
 import sys
@@ -36,6 +35,18 @@ from boto.compat import BytesIO, urlsplit, six
 import boto.s3.key
 
 logger = logging.getLogger(__name__)
+
+# Multiprocessing is unavailable in App Engine (and possibly other sandboxes).
+# The only method currently relying on it is s3_iter_bucket, which is instructed
+# not to use it by the NO_MULTIPROCESSING flag.
+try:
+    import multiprocessing.pool
+except ImportError:
+    logger.warning("multiprocessing could not be imported and won't be used")
+    NO_MULTIPROCESSING = True
+    from itertools import imap
+else:
+    NO_MULTIPROCESSING = False
 
 S3_MIN_PART_SIZE = 50 * 1024**2  # minimum part size for S3 multipart uploads
 WEBHDFS_MIN_PART_SIZE = 50 * 1024**2  # minimum part size for HDFS multipart uploads
@@ -578,7 +589,8 @@ def s3_iter_bucket(bucket, prefix='', accept_key=lambda key: True, key_limit=Non
     If `key_limit` is given, stop after yielding out that many results.
 
     The keys are processed in parallel, using `workers` processes (default: 16),
-    to speed up downloads greatly.
+    to speed up downloads greatly. If multiprocessing is not available, thus
+    NO_MULTIPROCESSING is True, this parameter will be ignored.
 
     Example::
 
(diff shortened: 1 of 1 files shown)


smart_open/smart_open_lib.py
```diff
@@ -611,7 +628,11 @@ def s3_iter_bucket(bucket, prefix='', accept_key=lambda key: True, key_limit=Non
         if key_limit is not None and key_no + 1 >= key_limit:
             # we were asked to output only a limited number of keys => we're done
             break
-    pool.terminate()
+
+    try:
```
piskvorky: Cleaner to check against `NO_MULTIPROCESSING`, same way as above, rather than try-catch.

piskvorky:
The py2.6 tests pass too, it was just some Travis glitch.

Can you add a test around this `NO_MULTIPROCESSING` functionality too though? Refactoring could easily break this later on and we'd never know.

smart_open/smart_open_lib.py
jsphpl: true

jsphpl:
@piskvorky i could implement it, but have no clue how to approach this. Any ideas?
* Check if `NO_MULTIPROCESSING` is set, well… but how to force `import multiprocessing` to throw an exception?
* How to check, that `s3_iter_bucket()` isn't using multiprocessing? Do we even need to, as it would fail given that multiprocessing is unavailable?

Would be great if you could point me in the right direction, as i have zero experience with mock & moto… Thanks!

piskvorky:
Sure, no problem. I was thinking something simple:

1. multiprocessing cannot be imported => flag `NO_MULTIPROCESSING` is set
2. force `NO_MULTIPROCESSING` flag and run all the existing `s3_iter_bucket` tests. maybe even in a loop (`for multiprocessing in [True, False]: do test body...`)

jsphpl:
Sounds good. But how, again, would you simulate the import error for multiprocessing, in order to check if the flag is set? I couldn't find anything about that in pyunit related docs, nor can i think of a creative solution to this…

piskvorky:
I didn't even think of testing that.

I just thought we'd set the flag manually (True/False) and see that both code paths work.

jsphpl:
Great, let’s do it like that…



--
Joseph Paul
Sent with Airmail

Am 12. November 2015 bei 13:40:52, Radim Řehůřek (notifications@github.com) schrieb:

I didn't even think of testing that.

I just thought we'd set the flag manually (True/False) and see that both code paths work.

smart_open/smart_open_lib.py
piskvorky: @jsphpl Can you fix this? Then we're ready to merge (the tests we can add later). Cheers!

jsphpl: PR updated. But still the 2.6 tests fail due to a timeout when testing s3_iter_bucket() - any idea about that?

piskvorky:
Thanks for the `if`  fix  @jsphpl -- merged :)

tmylk:
@jsphpl  Could we set the flag in the unit tests? (both True and False)

jsphpl:
@tmylk i'm having trouble with the test environment on my machine and don't want to push untested tests ;). So if you feel like writing the tests, please feel free! I'd be thankful if you could help! Otherwise i'll have to get the tests running…

tmylk:
@jsphpl Feel free to push "untested tests". They only take 30 sec to run and Travis is free. :)

jsphpl:
@tmylk True, i'll give it a shot.

smart_open/smart_open_lib.py
```diff
@@ -37,6 +36,18 @@
...
 
+# Multiprocessing is unavailable in App Engine (and possibly other sandboxes).
+# The only method currently relying on it is s3_iter_bucket, which is instructed
+# not to use it by the NO_MULTIPROCESSING flag.
+try:
+    import multiprocessing.pool
+except ImportError:
+    logger.warning("multiprocessing could not be imported and won't be used")
+    NO_MULTIPROCESSING = True
+    from itertools import imap
```
shaform: I think this might break Python 3. imap is only available in Python 2.

piskvorky: @tmylk how come the unit tests have been passing for a year then?

Issue #40, opened 2015-11-14
Cannot install if `LC_ALL=C`


nikicc:
When the system environment variable `LC_ALL=C` I cannot install `smart_open`. The problem is in the dependency `httpretty`, since setup.py requires the version `httpretty==0.8.6` which is [know not to work with](https://github.com/gabrielfalcao/HTTPretty/issues/223) `LC_ALL=C`. The error I get is this:

    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 133: ordinal not in range(128)

`httpretty` fixed this error in version `0.8.8`, so I am wondering if it would be possible to to relax the requirement to `httpretty>=0.8.6`? 

I actually discovered this when trying to install `gensim`, which also did not work since it requires `smart_open`.

piskvorky:
Oh, wow, not good.

A fix will be definitely welcome. Nobody's reported this yet.

tmylk:
Fixed by #41

nikicc:
Can this be reopened, since this does not work in version `1.3.1`?

The problem is that in #41 (that should fix this) the version was set incorrectly to `httpretty>=0.8.8`. It should actually be set to `httpretty>=0.8.12`, since `httpretty` version `0.8.12` is the first version that can be installed with `LC_ALL=C`.

piskvorky:
Yes, if the fix doesn't work, please reopen.

nikicc:
I would, but cannot. Can you?

piskvorky:
Sorry :)

piskvorky:
@tmylk : some progress in https://github.com/gabrielfalcao/HTTPretty/issues/278 .

nikicc:
After a quick check `httpretty 0.8.14` seems to work ok with python3 and `LC_ALL=C` (at least locally). Is there any other reason why we still require httpretty to be fixed? Could we relax the condition to allow `httpretty 0.8.14`?

tmylk:
Thanks for the update. I tried `httpretty 0.8.14` on with Python 3.4 and it failed same way as in https://github.com/piskvorky/smart_open/pull/59 . Raised  https://github.com/gabrielfalcao/HTTPretty/issues/291

nikicc:
OK, I see. I just test the installation part (which seems to be fixed), but forgot to run the tests. Thanks for the effort!

tmylk:
Update: Both httpretty==0.8.12 and httpretty==0.8.14 give test errors.

[...]

PR #41, opened 2015-11-16
Upgraded httpretty version; fixing #40

nikicc:
httpretty==0.8.6 cannot be installed with `LANG=C` and consequently neither can be `smart_open` (and `gensim`). This pull relaxes the version constrains since the bug was fixed in 0.8.8.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
 
     install_requires=[
         'boto >= 2.32',
-        'httpretty==0.8.6',
+        'httpretty>=0.8.8',
         'bz2file',
         'requests',
     ],


piskvorky:
Thanks!

nikicc:
Awesome! BTW, approximately when can we expect new version on pypi?

piskvorky:
I think after #39 is merged (only tests missing). CC @tmylk

nikicc:
Great, thanks!

nikicc:
Just figured out that this was bad. Actually, the first version of `httpretty` that work with `LC_ALL=C` is `0.8.12` and not `>=0.8.8` as was proposed by this PR. To work with `LC_ALL=C` this should be set to `>=0.8.12`.

Issue #36, opened 2015-11-04
Add support for scp(ssh) protocol


val314159:
This would be pretty useful to copy from generic shell accounts to S3 and vice versa.

piskvorky:
Yes, I can see how that is useful.

Python's [fabric](https://github.com/fabric/fabric) library has good support for this kind of thing; perhaps we could build on top of that?

Are you up for a pull request @val314159 ?

val314159:
Yeah, i hadn't thought of the exact mechanism, I figured the most dead simple way would be just to shell out a command.  

fabric might be overkill, there's also this: https://pypi.python.org/pypi/scpclient 
They're both front ends to the paramiko library.

Let me play around and see what I can do - it should be fairly straightforward.

piskvorky:
Perfect -- the simpler the dependencies, the better!

Thanks for offering the exploration / help. Feel free to open a PR for review once you have something reviewable.

ziky90:
This is a good idea and it would remove my need of issue #33 which I was about to implement.

PR #44, opened 2015-12-11
First attempt to implement reading via SSH/SCP

ziky90:
Goal of this PR is to add the functionality of reading files via SSH/SCP from remote machines part of what was discussed in the issue #36 .

PLEASE DO NOT MERGE YET:
TODO:

- [x] ~~implement tests~~
- [x] update README.rst

What should be discussed:
- authentication (should we care of authentication or can we estimate that everybody has copied SSH keys to the machine before using smart_open at least for the beginning)

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -67,7 +67,9 @@ def smart_open(uri, mode="rb", **kw):
        `./lines.txt`, `/home/joe/lines.txt.gz`, `file:///home/joe/lines.txt.bz2`
     2. Amazon's S3 (can also supply credentials inside the URI):
        `s3://my_bucket/lines.txt`, `s3://my_aws_key_id:key_secret@my_bucket/lines.txt`
-    3. HDFS: `hdfs:///some/path/lines.txt`
+    3. SSH: `ssh://ubuntu@ip_address:/some/path/lines.txt`
+    4. HDFS: `hdfs:///some/path/lines.txt`
+    5. WEBHDFS: `webhdfs://some/path/lines.txt`
 
     Examples::
 
@@ -116,6 +118,8 @@ def smart_open(uri, mode="rb", **kw):
             return HdfsOpenRead(parsed_uri, **kw)
         elif parsed_uri.scheme in ("webhdfs", ):
             return WebHdfsOpenRead(parsed_uri, **kw)
+        elif parsed_uri.scheme in ("ssh", ):
+            return SSHOpenRead(parsed_uri, **kw)
         else:
             raise NotImplementedError("read mode not supported for %r scheme", parsed_uri.scheme)
     elif mode in ('w', 'wb'):
@@ -205,6 +209,9 @@ class ParseUri(object):
 
             if not self.uri_path:
                 raise RuntimeError("invalid file URI: %s" % uri)
+        elif self.scheme == "ssh":
+            self.uri_path = parsed_uri.path
+            self.netloc = parsed_uri.netloc.rstrip(":")
         else:
             raise NotImplementedError("unknown URI scheme %r in %r" % (self.scheme, uri))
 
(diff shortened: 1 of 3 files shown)


ziky90:
There seems to be some bug in travis, based on the log there is problem with packages installation via pip.

Second question is that if we can assume for reading from ssh that everybody has his/her public key in `authorized_keys`? 
For my purposes it's sufficient for now, but I can imagine that someone might need some config file with username and password or setting these in some comparable way to s3 credentials.

piskvorky:
I'd say, implement what you need, cleanly. If someone needs more functionality, they can open a new PR to justify that extra complexity.

I don't understand the question with `authorized_keys`.

It may also make sense to pass additional options to SSH, so that it doesn't trigger interactive prompts ("known host checking" etc). But it's a slippery slope, I wouldn't want `smart_open` to end up taking responsibility for SSH setup and management.

CC @gojomo opinion?

piskvorky:
By the way @ziky90 , @tmylk is planning a new `smart_open` release this week -- if you want this functionality in there, now is the best time.

ziky90:
I've added that smart_open will not check for known hosts.

With `authorized_keys` I mean that this PR currently does not support any possibility to attach keys via -i parameter.

smart_open/smart_open_lib.py
```diff
@@ -269,6 +276,34 @@ def __exit__(self, type, value, traceback):
...
+
+    """
+    def __init__(self, parsed_uri):
+        if parsed_uri.scheme not in ("ssh"):
+            raise TypeError("can only process remote files via ssh")
+        self.parsed_uri = parsed_uri
+
+    def __iter__(self):
+        ssh = subprocess.Popen("ssh -o StrictHostKeyChecking=no " + self.parsed_uri.netloc + " 'cat " + self.parsed_uri.uri_path + "'",
+            stdout=subprocess.PIPE, shell=True)
```
gojomo: Choices on this line introduce risks that, while perhaps justified in many/most usage situations, should be clearly documented and perhaps require a user's explicit opt-in. (1) Users should know (perhaps by choosing a non-default option) that there's no host-key-checking; (2) composing a locally- and remotely- shelled execution string from the URI parts could introduce arbitrary execution risks at both endpoints, so extra care must be taken where such paths originate.

piskvorky:
@tmylk is working on fixing the travis tests.

Regarding SSH options -- let's make that a configurable list, via an object attribute (e.g. `obj.ssh_opts = ['-o "StrictHostKey=no"']`), with some sensible defaults.

Regarding the documentation, I agree with @gojomo that we should feature prominently the fact that `smart_open` is only meant for running on known, controllable inputs.

Also I'm not sure the `cat` escaping is right -- did you test this on remote filenames with spaces or special characters in them?

smart_open/smart_open_lib.py
```diff
@@ -269,6 +276,34 @@ def __exit__(self, type, value, traceback):
...
+    Implement streamed reader from remote server using SSH
+
+    """
+    def __init__(self, parsed_uri):
+        if parsed_uri.scheme not in ("ssh"):
+            raise TypeError("can only process remote files via ssh")
+        self.parsed_uri = parsed_uri
+
+    def __iter__(self):
+        ssh = subprocess.Popen("ssh -o StrictHostKeyChecking=no " + self.parsed_uri.netloc + " 'cat " + self.parsed_uri.uri_path + "'",
```
piskvorky: Probably better (more robust) to pass a list of arguments to `Popen` ctr, rather than a single hand-assembled string.

val314159:
Here's a pure-python (no shell) solution using Parimiko (the low level lib used by fabric)

It also has the side-benefit of maintaining connections so you only have to suffer through setup time once per connection :)

[https://github.com/piskvorky/smart_open/pull/58](https://github.com/piskvorky/smart_open/pull/58)

ziky90:
@val314159 Solution using Parimiko looks better. 
So I'll close this PR.

PR #45, opened 2015-12-16
pr_38_with_tests

tmylk:
Fixing tests that are failing in #38


tmylk:
Removed the second isinstance check for an S3 key as it is redundant.

Changed tests to be in sync with us switching to get_key instead of deprecated lookup.

PR #50, opened 2016-01-02
Fix #49 Mode restriction applied only to the relevant types.

### LINKED ISSUE #49: Get rid of file mode restrictions
Currently `smart_open` is limiting file modes to `('r', 'rb', 'w', 'wb')` [here](https://github.com/piskvorky/smart_open/blob/master/smart_open/smart_open_lib.py#L110).

I don't think there is any reason for this -- remove the restriction and allow whatever mode is supported by the underlying storage ([such as `wb+`](http://stackoverflow.com/questions/34396300/gensim-installation-on-yosemite-using-anaconda)).

tmylk:
Mode restriction only applied to S3, HDFS, WebHDFS modes

@piskvorky @gojomo Would appreciate your opinion before bug fix release in 8 hours.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -137,14 +137,14 @@ def smart_open(uri, mode="rb", **kw):
             if mode in ('r', 'rb'):
                 return HdfsOpenRead(parsed_uri, **kw)
             else:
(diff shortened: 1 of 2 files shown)


piskvorky:
Looks good to me :+1:

gojomo:
Looks good to me – the unit test seems to cover the case that was failing for gensim. (But has that been separately tested to succeed with this branch?)

PR #61, opened 2016-01-26
Set key to empty when S3OpenWrite closed writing any data

petedmarsh:
S3 does not support zero-content multi-part file uploads, i.e. if you try to
multi-part upload no data then you'll get an error. This meant that if, for
some reason, you opened an S3 key for writing but actually didn't write to it
at all then nothing would be written to that key.

This changes S3OpenWrite to fallback to set_contents_from_string to set a key
to have an empty (zero byte) in this situation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+* HEAD
+
+  - Write an empty key to S3 even if nothing is written to S3OpenWrite (PR #61, @petedmarsh)
+
 * 1.3.2, 3rd January 2016
 
   - Bug fix release to enable 'wb+' file mode (PR #50)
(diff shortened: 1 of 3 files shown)


piskvorky:
Thanks for the fix and also for the unit test, @petedmarsh !

Can you add a [CHANGELOG](https://github.com/piskvorky/smart_open/blob/master/CHANGELOG.rst) entry on top?

@tmylk please review and merge.

petedmarsh:
@piskvorky, done

petedmarsh:
Build failure looks like a build issue rather than a failing test, re-running should get everything passing

tmylk:
Green on re-run

PR #63, opened 2016-02-24
Passing extra parameters when writing to S3 keys (#35)

asieira:
Passing along extra parameters to `initiate_multipart_upload` when
writing S3 keys directly. This is necessary, for example, to allow
server-side encryption to be used through the `encrypt_key` parameter.


tmylk:
Thanks for the PR. That would be useful.
Just need a small test and a note in the CHANGELOG, then ready to merge.

tmylk:
@asieira Would you have time to add the finishing touches for the April release?

asieira:
Argh, I am so sorry for disappearing like that. Things have been truly hectic around here.

Issue #55, opened 2016-01-12
Not able to run tests


AndreaCrotti:
For some weird reason I can't seem to run tests with "python setup.py test", the error here below.
I'm getting this both with Python 2.7.10 and Python 3.4, any idea why?

```
    $> python setup.py test
    running test
    running egg_info
    writing requirements to smart_open.egg-info/requires.txt
    writing smart_open.egg-info/PKG-INFO
    writing top-level names to smart_open.egg-info/top_level.txt
    writing dependency_links to smart_open.egg-info/dependency_links.txt
    reading manifest file 'smart_open.egg-info/SOURCES.txt'
    writing manifest file 'smart_open.egg-info/SOURCES.txt'
    running build_ext
    Traceback (most recent call last):
      File "setup.py", line 66, in <module>
        'Topic :: Database :: Front-Ends',
      File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/home/andrea/.virtualenvs/smart_open/local/lib/python2.7/site-packages/setuptools/command/test.py", line 142, in run
        self.with_project_on_sys_path(self.run_tests)
      File "/home/andrea/.virtualenvs/smart_open/local/lib/python2.7/site-packages/setuptools/command/test.py", line 122, in with_project_on_sys_path
        func()
      File "/home/andrea/.virtualenvs/smart_open/local/lib/python2.7/site-packages/setuptools/command/test.py", line 163, in run_tests
        testRunner=self._resolve_as_ep(self.test_runner),
      File "/usr/lib/python2.7/unittest/main.py", line 94, in __init__
        self.parseArgs(argv)
      File "/usr/lib/python2.7/unittest/main.py", line 149, in parseArgs
        self.createTests()
      File "/usr/lib/python2.7/unittest/main.py", line 158, in createTests
        self.module)
      File "/usr/lib/python2.7/unittest/loader.py", line 130, in loadTestsFromNames
        suites = [self.loadTestsFromName(name, module) for name in names]
      File "/usr/lib/python2.7/unittest/loader.py", line 103, in loadTestsFromName
        return self.loadTestsFromModule(obj)
[...]

PR #65, opened 2016-03-23
Remove explicit version pins

nvie:
It's generally good practice to be as tolerant as possible in libs like this. Unless there is a very good reason not to, don't put pins on dependencies. I'm assuming this uses specific httpretty>=0.8.10 features and boto>=2.32 features. Also, it's good practice to only ever use `>=`, and never `<=` (as this will prevent these dependencies from being upgraded).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,9 +45,9 @@ setup(
 
     install_requires=[
         'boto >= 2.32',
-        'httpretty==0.8.10',
+        'httpretty >= 0.8.10',
         'bz2file',
-        'requests<=2.8.1',
+        'requests',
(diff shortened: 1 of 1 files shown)


piskvorky:
The reason is in #40 and #55.

tmylk:
Requests version updated in #67
(FYI the reason tests pass in #65 is because versions are still pinned in .travis.yml)

nvie:
Thanks. Could you release this new version to PyPI?

tmylk:
What is the advantage of having a newer version of requests?

PR #68, opened 2016-03-29
Support profile_name as a keyword argument.

robcowie:
Allow passing `profile_name` to optionally specify a boto credentials profile.

`aws_access_key_id` and `aws_secret_access_key` are still supported and in fact will take precedence if provided. See http://boto.cloudhackers.com/en/latest/boto_config_tut.html#credentials for auth source precedence.

```python
with smart_open.smart_open('s3://bucket/prefix/', profile_name='name') as infile:
    pass
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
+* 1.4.0, 29th March 2016
+
+  - Allow passing boto `profile_name` to smart_open (PR #68, @robcowie)
+
+
 * 1.3.2, 3rd January 2016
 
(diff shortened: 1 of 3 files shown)


robcowie:
I think the tests failed for unrelated reasons (see https://github.com/piskvorky/smart_open/issues/62).

tmylk:
Thanks for the contribution and the tests.
Please add a short description in the CHANGELOG and we are good to go.

robcowie:
Happy with version `1.4.0`? I guess I'd call this a "backward compatible change of functionality" which semver.org would regard as a minor version bump.

PR #73, opened 2016-05-24
add relative path support for file

yupbank:
for convenience of relative file paths

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -248,21 +248,14 @@ class ParseUri(object):
                 raise RuntimeError("invalid S3 URI: %s" % uri)
         elif self.scheme == 'file':
             self.uri_path = parsed_uri.netloc + parsed_uri.path
-            self.uri_path = expand_full_path(self.uri_path)
+
+            # '~/tmp' may be expanded to '/Users/username/tmp'
+            self.uri_path = os.path.expanduser(self.uri_path)
 
             if not self.uri_path:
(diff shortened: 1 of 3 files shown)


tmylk:
Thanks for the PR! 
Can we make it optional(on by default)? 

The tests fail as they don't expect an expansion yet. Need tests for both.

yupbank:
I think on by default would be nice, since it is not conflicting with other schema :)

And i have updated the abspath which is not needed for non expanding case

tmylk:
@yupbank Are you planning to add tests for this new functionality? And a note in Changelog?

yupbank:
hey, i have added a few tests and modified the change log :)

smart_open/smart_open_lib.py
```diff
@@ -241,13 +248,23 @@ def __init__(self, uri, default_scheme="file"):
...
             self.uri_path = parsed_uri.netloc + parsed_uri.path
+            self.uri_path = expand_full_path(self.uri_path)
 
             if not self.uri_path:
                 raise RuntimeError("invalid file URI: %s" % uri)
         else:
             raise NotImplementedError("unknown URI scheme %r in %r" % (self.scheme, uri))
 
 
+def expand_full_path(path):
```
piskvorky: What is this for? Why create an alias to a built-in function?

tmylk:
@yupbank  Thanks for the PR!

piskvorky:
I don't see any changes to CHANGELOG in this PR.

Issue #12, opened 2015-02-14
Unable to read gz files on s3


chuinker:
It reads them, but it the data remains compressed, thus defeating line iteration.

piskvorky:
Transparent (de)compression is only supported for local files now.

Should be possible to do it transparently for S3 files too, using Python's `zlib`. Let me know if you want to tackle this -- a low-hanging, extremely useful feature!

brianmingus:
+1 for this issue, I have some example code here: https://gist.github.com/brianmingus/a47f26760d244ba7e9d1

asieira:
It's interesting how smart_open is evolving into something similar for the Python world to what [Apache Commons VFS](https://commons.apache.org/proper/commons-vfs/) is to the JVM world. Maybe there's some inspiration of the abstractions they used to build something more general.

asieira:
This is useful for me as well, as soon as https://github.com/piskvorky/smart_open/pull/38 gets merged I could write this one.

piskvorky:
Thanks for the link @asieira -- I didn't know about VFS.

I'm all for learning from other people's mistakes -- what abstractions and designs in particular do you think would be useful?

asieira:
The first thing they did was to create the abstraction for a *file system*, not opening a single file. Going the single file route as smart_open has gone so far is great for formats like xz, gzip and bzip... mas if you want to handle archives that themselves contain internal structure (like zip, tar, etc) that won't work as well. So in VFS, opening a ZIP file is akin to traversing a virtual folder to access the content inside.

Plus, they worked in an arbitrary number of layers. You can build a URI like gz://zip://ftp://ftp.example.com/file/blah.zip!/zipfolder1/file.gz and access a GZIP file, inside a ZIP file, read from an FTP server.

Plus, it's an architecture that can be extended. They defined a set of abstract classes that you can implement to define a new type of filesystem in addition to the built-in ones.

All great ideas, but maybe too complicated for smart_open and worthy of a separate independent project that emulates those ideas in Python. 

At the very least, thinking of those ideas I would be inspired to implement support for gzip and bzip compression / decompression is a way that is general for all supported file types in smart_open.

AndreaCrotti:
+1 on this, I might be able to implement it soon since we'll probably need it..thanks

piskvorky:
Sounds great @AndreaCrotti ... that would be really sueful!

mpenkov:
[...]

Issue #62, opened 2016-01-29
Travis tests sometimes hang


piskvorky:
The Travis CI tests sometimes hang, and this may be indicative of a bug in `s3_iter_bucket` routine, which uses multiprocessing.

The failing tests mess up unrelated PRs (need manual test restarts, like in #61), so we want this fixed ASAP.

menshikh-iv:
@piskvorky I have not seen hangs lately, it's still relevant?

menshikh-iv:
Yes, relevant, reproduced with Travis, problem in `test_s3_iter_bucket_moto (smart_open.tests.test_smart_open.S3IterBucketTest)`

mpenkov:
We don't use Travis anymore.

PR #75, opened 2016-06-09
Resolve issue #12 (unable to read gz from S3)

mpenkov:


tmylk:
Thanks for the PR!

Can we make it an optional dependency until `gzipstream` appears on PyPi? i.e. try/catch around import 

Several people requested for `gzipstream` to be added to PyPi but t is not likely to happen. Git install is ok for us.  https://github.com/Smerity/gzipstream/issues/3

piskvorky:
Does bundling `gzipstream` make sense? Are the licenses compatible?

tmylk:
We can re-license the MIT-licensed `gzipstream` as we see fit.

mpenkov:
@tmylk Gzipstream is optional already. If it's not installed, ImportError is caught and gzipstream-related functionality is disabled.

@piskvorky I wouldn't recommend bundling. If someone improves it e.g. by adding write capability we'd have to re-bundle. It isn't a required component, and is not that hard to install, anyway.

piskvorky:
Thanks for feedback @mpenkov . I'm actually still +1 on bundling, if it's a simple, pure-Python script. 

Potential "rebundling" seems straightforward, the added functionality very neat for us to rely on, and asking users to git-install unnecessarily demanding.

But I'll defer the decision to @tmylk .

Are there any performance implications?

tmylk:
I would like to bundle it as there has been no changes to it for 2 years. We are in better position to add write capability than current maintaners.

mpenkov:
Ok, those are good points. I'll do it later today.

mpenkov:
@tmylk Just to double-check, gzipstream is GNU GPL **not** MIT. If we bundle it, what do we do about their license? Do we have to do anything to our MIT license? Or do the two just coexist side-by-side? I'm not a lawyer, so I thought I'd check :)

tmylk:
Their github is confusing. It says MIT in Readme and GPL in the license file.
If it is GPL then we can't use it.

tmylk:
Let's hope the maintainer gets back quickly to my query in https://github.com/Smerity/gzipstream/issues/7

mpenkov:
Not sure why the Travis CI build failed, it looks like it could be issue #62.

mpenkov:
Included by https://github.com/piskvorky/smart_open/pull/76

Issue #13, opened 2015-02-14
readline() method


chuinker:
This would be useful for extracting the header from a csv.

piskvorky:
I agree. Can you implement it?

coreyatmoat:
I'll give it a shot.

piskvorky:
@coreyatmoat what's your progress?

coreyatmoat:
Sadly no. A while back a co-worker wrote https://gist.github.com/coreyatmoat/c961f99dfb0cfabdda54 as a sort of work around, but I haven't had time to figure out how it'd fit into your framework.

Mgutjahr:
+1

mpenkov:
I'm having a look at this now, since one of my projects needs this to work too :)

Is there any reason why we need to maintain independent pointers for `read` and `__iter__`? The default implementation of `readline` shares the file pointer with `read` (and other I/O operations), to the best of my knowledge.

piskvorky:
The reason was related to difficulties with buffering IIRC, but was only technical. That behaviour is not a part of the API contract (I don't think it's even documented).

If we can make a single pointer work, that's fine, even preferable.

mpenkov:
@piskvorky OK. I think I've got it working. Have a look here: https://github.com/mpenkov/smart_open/commit/aca3a18358afb5f42b5f7e5c1355e8ff93a9bcea

I made a separate branch (`readline`) for it, since an open pull request (https://github.com/piskvorky/smart_open/pull/75) is currently using the clone's master.

What do you think is the cleanest way of getting this into your repo? I can think of:

1. You guys merge https://github.com/piskvorky/smart_open/pull/75 whenever you're ready
2. I update my clone and merge the branch
3. I make a new pull request for the readline stuff

Let me know if there's a simpler or more convenient way.

piskvorky:
Great stuff! @tmylk please review.

Re. branches: it's your call. If splitting the changes into two separate PRs is too complicated, doing both in one branch is fine with me. We appreciate your work, this is some much needed functionality.

mpenkov:
@piskvorky @tmylk I've authored a request to pull from my `readline` to your `master` (https://github.com/piskvorky/smart_open/pull/76). Let me know if you need anything else from me.

mpenkov:
@tmylk I think we can close this now. 78c461e resolved this.

PR #76, opened 2016-06-12
Resolve issues #12 (gzipped S3) and #13 (readline)

mpenkov:
This pull request includes the changes from https://github.com/piskvorky/smart_open/pull/75 (which I have now closed).


mpenkov:
Looks like Python 3.x builds are failing. I'm new to Python 3.x, but will see if I can fix these issues during the week.

On a side note, I'm not sure why these problems have come up now as opposed to https://github.com/piskvorky/smart_open/issues/75 - the code is mostly the same. If anyone can shed light onto this, I'd appreciate it greatly.

gzipstream/gzipstreamfile.py
```diff
@@ -0,0 +1,89 @@
+import io
```
piskvorky: We definitely want to attribute and give credit to the original author / repository / license here!

bin/test_s3_gzip.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+import smart_open
+
+URL = "s3://aws-publicdatasets/common-crawl/crawl-data/CC-MAIN-2014-35/\
+segments/1408500800168.29/warc/\
+CC-MAIN-20140820021320-00000-ip-10-180-136-8.ec2.internal.warc.gz"
+
+with smart_open.smart_open(URL) as fin:
```
piskvorky: Doesn't look like a unit test. If it's some internal test you're using, no need for it to be in repo.

If it's meant to be a unit test, the right place for it is under `tests`, such as inside `tests/test_smart_open.py`.

gzipstream/gzipstreamfile.py
mpenkov: Done.

bin/test_s3_gzip.py
mpenkov: Done.

tmylk:
Looks good. Just waiting for Travis builds to come back after we moved project to another org.

Adding a couple of lines in the CHANGELOG would be good too.

mpenkov:
@tmylk What date and version should I use in the changelog? Are placeholders acceptable?

gzipstream/tests/test_gzipstreamfile.py
```diff
@@ -0,0 +1,23 @@
+from .. import gzipstreamfile
+import unittest
+
+GzipStreamFile = gzipstreamfile.GzipStreamFile
+
+
+class TestSequenceFunctions(unittest.TestCase):
```
tmylk: Just noticed these tests are not run in Travis. What is the best way to make the test runner pick them up?

tmylk:
@mpenkov Thanks for updating the changelog! Once gzipstream tests run, then will merge in.

gzipstream/tests/test_gzipstreamfile.py
mpenkov: @tmylk I think moving it into the same subdirectory with the other tests is the easiest way to go. Let me know if this is acceptable for you.

tmylk: That is acceptable. Thanks!

tmylk:
Thanks a lot for this great PR!

mpenkov:
You're welcome. Thank you for such a useful module!

PR #79, opened 2016-06-24
Invert NO_MULTIPROCESSING flag

Janrain-Colin:


tmylk:
Hi, may I ask what is the purpose of this change?

Janrain-Colin:
Certainly!  It's confusing to maintain booleans that are described as a negative.  So instead of `not_found = True` it usually makes more sense to use `found = False`.  In a way of thinking, it folds the constant `NOT_` into the boolean the variable contains, simplifying the expressions.  In this case, it removes the somewhat confusing `if not NO_MULTIPROCESSING:`.

tmylk:
Thanks.Agree that it is ncer. I am very concerned though that this cosmetic change breaks  backwards compatibility so I would rather not merge it in.

Janrain-Colin:
How's that?

tmylk:
Actually, thinking about it... It is an internal param so no back-compatibility is needed. Will merge on Sunday without the last commit. Thanks!

tmylk:
Thanks for the PR!
Merged and squashed in https://github.com/RaRe-Technologies/smart_open/commit/cde325845886a5eec6547a89c5dbf829964fb1a7

PR #80, opened 2016-06-29
Bump version in setup.py to match CHANGELOG.rst

mpenkov:


piskvorky:
@tmylk what's going on with the versions?

Also, let's remove the "download count" until PyPI is fixed.

tmylk:
@mpenkov Thanks for this. I usually update `setup.py` when I do the release.

PR #81, opened 2016-07-02
Fixing readline() behavior when EOF has been reached.

mpenkov:
readline returns empty string once EOF has been reached.
It shouldn't _ever_ return None, because it breaks code like this:

    buf += reader.readline()

with a TypeError.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -339,6 +339,15 @@ class S3ReadStream(io.BufferedReader):
         self.stream = _S3ReadStream(key)
         super(S3ReadStream, self).__init__(self.stream)
 
+        #
+        # Generator behavior changed in Python 3.
+        # Instead of g.next(), Python 3.x expects next(g) or g.__next__().
+        # http://stackoverflow.com/questions/1073396/is-generator-next-visible-in-python-3-0
+        # This is a patch to keep things working in Python 3.x.
+        #
+        if sys.version_info[0] == 3:
+            self.next = self.__next__
+
     def read(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


piskvorky:
Is this a fix for the recent GzipStream PR merge, or has the buggy version already been released?

mpenkov:
@piskvorky Yes, this is a fix for the recent merge. Previous tests didn't catch this bug, so I've also added some unit tests to cover it. I'm trying to figure out why these new tests are failing on Travis CI 3.x builds. Everything passes in my local Python 3.5.1 environment.

To answer your second question, I'm not sure whether the buggy version has been released. From what I can see, the [current version on pip](https://pypi.python.org/pypi/smart_open) is still 1.3.3 - the previous version before the merge.

smart_open/smart_open_lib.py
```diff
@@ -339,6 +339,15 @@ def __init__(self, key):
         self.stream = _S3ReadStream(key)
         super(S3ReadStream, self).__init__(self.stream)
 
+        #
+        # Generator behavior changed in Python 3.
```
piskvorky: I think we're using `six` in smart_open anyway, so `six.next` simpler?

mpenkov: @piskvorky Check out the next commit, it's possible to avoid the whole generator business entirely.

mpenkov:
@piskvorky @tmylk Are you waiting on anything from me? Wanted to double-check, just in case.

tmylk:
@mpenkov Thanks for the bugfix. I like the very good translation of 'how are you' into Russian. :)

PR #96, opened 2016-11-15
Add retry to get_contents to address SSLError that can occur

bbbco:
I have been experiencing https://github.com/boto/boto/issues/2409 while using `smart_open`. This update allows us to retry three times before finally failing because of a read timeout.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -744,7 +744,7 @@ class WebHdfsOpenWrite(object):
         self.close()
 
 
-def s3_iter_bucket_process_key(key, retries = 3):
+def s3_iter_bucket_process_key(key, retries=3):
     """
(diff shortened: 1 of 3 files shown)


smart_open/smart_open_lib.py
```diff
@@ -743,13 +744,23 @@ def __exit__(self, type, value, traceback):
         self.close()
 
 
-def s3_iter_bucket_process_key(key):
+def s3_iter_bucket_process_key(key, retries = 3):
```
piskvorky: PEP8: No spaces around method arguments.

tmylk:
@bbbco A test would be very welcome.
It can be tested by setting `s3_iter_bucket_process_key_mock.side_effect = SSLError`

bbbco:
Updated with PR items addressed.

tmylk:
Thanks for the PR and great tests!

PR #97, opened 2016-11-21
Add encoding keyword argument to file_smart_open.

notnami:
This adds the possibility to specify the file encoding in `file_smart_open`. This is necessary under Python 3 on Windows, where files are opened with cp-1251 by default, which results in errors for files that contain Unicode characters.


notnami:
I am not sure why the build fails with Python 2.7, given that it passes with all other versions of Python, including Python 2.6.

Issue #98, opened 2016-11-21
cant use s3 bucket names with dots


balihoo-gens:
```
 with smart_open.smart_open('s3://balihoo.fulfillment.dev/retain_7_0/test.txt', 'wb') as f:
    f.write('test')
```
results in:
`ssl.CertificateError: hostname 'balihoo.fulfillment.dev.s3.amazonaws.com' doesn't match either of '*.s3.amazonaws.com', 's3.amazonaws.com'`

I tried to fix this with the 'host' parameter, but I can't get it to write the file.

tmylk:
Hi @balihoo-gens, Does adding `calling_format` fix it as in your previous PR https://github.com/RaRe-Technologies/smart_open/pull/83 ?

balihoo-gens:
ha, will try. That PR is not mine, but by one of my coworkers, @balihoo-dengstrom
I actually ended up using s3fs: https://github.com/dask/s3fs

tmylk:
Thanks for the link. You are not using a proxy, right? Or did s3fs get around https://github.com/boto/boto3/issues/338 ? That is the issue that is blocking us from moving to boto3.

balihoo-gens:
I am using this from an AWS Lambda function

mpenkov:
This looks like a duplicate of Issue #26 @menshikh-iv

menshikh-iv:
@mpenkov yep
@balihoo-gens already fixed in current master

PR #99, opened 2016-11-22
Adding seek method to S3ReaderStream implementation

tejinderss:
If I use the underlying S3ReaderStream implementation directly without using S3OpenRead, it will throw attribute error exceptions, I have followed the same path django uses for their base files, silencing that error.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -377,6 +377,12 @@ class S3ReadStream(io.BufferedReader):
         except ValueError:
             return ''
 
+    def seek(self, position, whence=0):
+        try:
+            super(S3ReadStream, self).seek(position, whence)
+        except AttributeError:
+            pass
+
 
 class S3OpenRead(object):
     """


piskvorky:
I didn't realize this was even possible with S3 -- being able to seek into the middle of objects would be awesome.

How was this tested?

Silencing exceptions is not a good idea, what exactly is the problem there?

tejinderss:
The problem is if there is no buffer generated yet or the position is out
greater than the content length, it will raise attribute error. Try
instantiating s3streamreader and call seek without calling read first or
calling seek with out of bounds position after read it will raise attribute
error. Silencing the exception does not have any aide effect, since the
underlying s3 key object does not have seek method implemented, if the
buffer is not generated yet, it will try to call seek on the original
stream object hence the attribute error. I can add a comment but not sure
how we test it? Any suggestions ?

tejinderss:
This is not behaving correctly, sorry closing it out :(, will further investigate and open another pull request

piskvorky:
Alright. This is a highly desired functionality, so let me know how this goes.

There was also a link to another implementation of an "S3 filesystem abstraction" in the other issue here: https://github.com/RaRe-Technologies/smart_open/issues/98#issuecomment-262271495 . Perhaps worthwhile inspiration?

Issue #95, opened 2016-11-15
For s3 file, readline doesn't work after seek(0)


tejinderss:
If we do .seek(0) on an open s3 file instance and calls the readline on it after, it just keeps on giving the empty string '' back and also the iterator is consumed at that point of time. Please find the code snipped link below:
https://dpaste.de/VXTi

EDIT: Updated with code snippet link to dpaste

tmylk:
hmm. could you add a small code snippet to reproduce?

tejinderss:
Added

agent-jay:
@tejinderss The link is dead. Could you please re-share or embed here?

agent-jay:
I've managed to reproduce the error. Code embedded

```
with smart_open.smart_open('s3://key_id:key_secret@smartopentest/valid_text.txt') as fin :
    fin.read() 
    fin.seek(0)
    fin.readline()
```

The readline() outputs a ' ' instead of the first line

agent-jay:
@tmylk I've tested out the following bit of code, is it a valid solution? I'm not sure why the read_key.close() was there to begin with, and how seek has managed to pass the unit tests all this time
    
def seek(self, offset, whence=0):
        """
        Seek to the specified position.

        Only seeking to the beginning (offset=0) supported for now.

        """
        if whence != 0 or offset != 0:
            raise NotImplementedError("seek other than offset=0 not implemented yet")
        
        # self.read_key.close(fast=True) 

        self._open_reader()
```

menshikh-iv:
Works correctly with current master fbc82cc04bf92e8c588c49c43ff5aff8234ea87d

PR #100, opened 2016-11-23
Fixing bug #95

tejinderss:
Don't think key.close(fast=True) is doing what it is supposed to do, I could not find anything on s3 object that could possible do what is needed in this case, so I just delete and create the key object again, fixing the bug

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -422,8 +422,22 @@ class S3OpenRead(object):
         """
         if whence != 0 or offset != 0:
             raise NotImplementedError("seek other than offset=0 not implemented yet")
-        self.read_key.close(fast=True)
+        self._regenerate_key()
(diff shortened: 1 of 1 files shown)


piskvorky:
I don't understand -- what is the actual bug? How does it manifest as exception/unwanted behaviour?

Documentation for `close(fast=True)` is in the [boto docs](http://boto.cloudhackers.com/en/latest/ref/s3.html#boto.s3.key.Key.close).

tejinderss:
I didn't dig into the boto code but the current seek(0) is broken, I might dig into the boto code later, meanwhile if you get a chance, please have a look as well. I will monkey patch this for the time being, since I need this functionallity

menshikh-iv:
Already fixed in current master fbc82cc04bf92e8c588c49c43ff5aff8234ea87d

PR #101, opened 2016-11-29
Better support for custom S3 servers.

robottwo:
This patch adds support for custom S3 servers in the connection string.
It also adds explicit support for setting the server port, and whether
or not to use SSL, both as paramaters to the smart_open function as
well as within the connection string.

These changes are neccessary to be able to connect to s3proxy and
other custom s3 servers which don't run on the default port,
or neccessarily use SSL.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -33,6 +33,7 @@ elif sys.version_info[0] == 3:
     import http.client as httplib
 
 from boto.compat import BytesIO, urlsplit, six
+import boto.s3.connection
 import boto.s3.key
 from ssl import SSLError
 
@@ -126,11 +127,18 @@ def smart_open(uri, mode="rb", **kw):
             # local files -- both read & write supported
             # compression, if any, is determined by the filename extension (.gz, .bz2)
             return file_smart_open(parsed_uri.uri_path, mode)
-        elif parsed_uri.scheme in ("s3", "s3n"):
+        elif parsed_uri.scheme in ("s3", "s3n", "s3u"):
+            kwargs = {}
             # Get an S3 host. It is required for sigv4 operations.
-            host = kw.pop('host', None)
-            if not host:
-                host = boto.config.get('s3', 'host', 's3.amazonaws.com')
+            host = kw.pop('host', parsed_uri.host)
+            port = kw.pop('port', parsed_uri.port)
+            if port != 443:
+                kwargs['port'] = port
+
+            if not kw.pop('is_secure', parsed_uri.scheme != 's3u'):
+                kwargs['is_secure'] = False
+                # If the security model docker is overridden, honor the host directly.
+                kwargs['calling_format'] = boto.s3.connection.OrdinaryCallingFormat()
 
             # For credential order of precedence see
             # http://boto.cloudhackers.com/en/latest/boto_config_tut.html#credentials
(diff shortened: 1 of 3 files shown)


tmylk:
Thanks for the improvements! 
Could you please add tests for the new modes and extend README.md

robottwo:
I don't know where to find a public server which provides the new modes, in order to add tests for them :(

robottwo:
Pushed a README.rst update.

tmylk:
@robottwo just mock tests would do

robottwo:
@tmylk , I added a new unit test to cover the new calling convention. 

Thanks!
-Dan

smart_open/smart_open_lib.py
```diff
@@ -126,19 +127,27 @@ def smart_open(uri, mode="rb", **kw):
...
             return file_smart_open(parsed_uri.uri_path, mode)
-        elif parsed_uri.scheme in ("s3", "s3n"):
+        elif parsed_uri.scheme in ("s3", "s3n", "s3u"):
+            kwargs = {}
             # Get an S3 host. It is required for sigv4 operations.
-            host = kw.pop('host', None)
-            if not host:
-                host = boto.config.get('s3', 'host', 's3.amazonaws.com')
+            host = kw.pop('host', parsed_uri.host)
+            port = kw.pop('port', parsed_uri.port)
```
piskvorky: Does this mean host and port are now mandatory in URIs?

That wouldn't be an acceptable change (old behaviour must continue to work the same).

robottwo: No, it doesn't mean that. parsed_uri sets the same default host and port as before, it just happens in another part of the code. Previous behavior is preserved.

smart_open/tests/test_smart_open.py
```diff
@@ -483,6 +483,22 @@ def test_s3_mode_mock(self, mock_write, mock_boto):
...
 
+    @mock.patch('smart_open.smart_open_lib.boto')
+    @mock.patch('smart_open.smart_open_lib.S3OpenWrite')
+    def test_s3_unsecured_mode_mock(self, mock_write, mock_boto):
+        """Are s3u:// open modes passed correctly?"""
+        # Configure the mock boto.config.get to return default host
+        smart_open.smart_open_lib.boto.config.get.return_value = 'example.com'
+
+        # correct write mode, correct s3 URI
+        smart_open.smart_open("s3u://user:secret@example.com@mybucket/mykey", "w")
```
tmylk: what about port?

smart_open/tests/test_smart_open.py
```diff
@@ -483,6 +483,22 @@ def test_s3_mode_mock(self, mock_write, mock_boto):
...
+        mock_boto.connect_s3.assert_called_with(aws_access_key_id='user', aws_secret_access_key='secret',
```
piskvorky: Prefer hanging indent, not vertical alignment ([pep8 examples](https://www.python.org/dev/peps/pep-0008/#indentation)).

smart_open/tests/test_smart_open.py
robottwo: Added another test to include the port.

smart_open/tests/test_smart_open.py
robottwo: Fixed.

robottwo:
The Ci error looks like an unrelated timeout in the unit test

smart_open/tests/test_smart_open.py
```diff
@@ -483,6 +483,40 @@ def test_s3_mode_mock(self, mock_write, mock_boto):
         mock_boto.connect_s3().get_bucket.assert_called_with("mybucket")
```
tmylk: Can we assert for default port here?

robottwo: Not sure how to do that? Not a mock expert.

tmylk: Just need to check that the default port is right. That nothing has changed

robottwo: OK, added a test which makes sure that the 'port' argument isn't even passed in the default case.

tmylk:
Thanks a lot for the PR!

piskvorky:
Thanks -- deserves a CHANGELOG entry too!

PR #102, opened 2016-12-03
make it possible to change # of retries

shaform:


smart_open/smart_open_lib.py
```diff
@@ -744,6 +744,10 @@ def __exit__(self, type, value, traceback):
         self.close()
 
 
+def s3_iter_bucket_process_key_with_args(args):
```
piskvorky: I'd prefer `kwargs` -- too easy to make a mistake in the order of arguments otherwise. Better use named arguments (beyond the mandatory `key`, which can be positional).

tmylk:
Thanks for the PR!
May I ask to add a test of this new feature, similar to https://github.com/RaRe-Technologies/smart_open/pull/96/files#diff-36c1f33ff2d12844dc387270ca11d040R694

shaform:
Thanks for the reply. I've added the test cases.

shaform:
It appears the test sometimes hangs (#62) possibly because mock_s3 does not go well with multiprocessing. So I've fixed it as well with a new commit.

tmylk:
@shaform Not sure about the last commit. It lowers our test coverage signifcantly. Will merge only the first one.

shaform:
I think that's because `smart_open.smart_open("", 'wb')` is not being
tested anywhere else. Maybe we could add a new test for that.

shaform:
@tmylk All right, now I look at the code, I find that there are actually other places that have `smart_open.smart_open("", 'wb')` being tested. But then I couldn't find out why the coverage would be decreased. Could share the coverage report?

shaform:
@tmylk Ok. I've decided to revert the second commit. This makes the test more likely to fail due to #62 though.

smart_open/smart_open_lib.py
```diff
@@ -984,15 +988,15 @@ def s3_iter_bucket(bucket, prefix='', accept_key=lambda key: True, key_limit=Non
...
     """
     total_size, key_no = 0, -1
-    keys = (key for key in bucket.list(prefix=prefix) if accept_key(key.name))
+    keys = ({'key': key, 'retries': retries} for key in bucket.list(prefix=prefix) if accept_key(key.name))
 
     if MULTIPROCESSING:
         logger.info("iterating over keys from %s with %i workers" % (bucket, workers))
         pool = multiprocessing.pool.Pool(processes=workers)
-        iterator = pool.imap_unordered(s3_iter_bucket_process_key, keys)
+        iterator = pool.imap_unordered(s3_iter_bucket_process_key_with_kwargs, keys)
```
tmylk: Please use `partial`  instead of defining a new function `s3_iter_bucket_process_key_with_kwargs`. The number of retries is the same for each worker.

tmylk:
Thanks. Just a decorative change requested.

smart_open/smart_open_lib.py
shaform: @tmylk 
Surely, but as mentioned earlier, this would cause errors in Python 2.6. Specifically, it's because partial conflicts with multiprocessing in Python 2.6: https://bugs.python.org/issue5228.

In order to make it work on Python 2.6, I've made some changes according to this thread:
https://mail.python.org/pipermail/python-dev/2009-February/086039.html

tmylk:
Apologies, didn't notice the discussion above. Thanks for this new feature!

smart_open/smart_open_lib.py
```diff
@@ -28,11 +28,23 @@
...
 
+from functools import partial
+
 
 IS_PY2 = (sys.version_info[0] == 2)
 
 if IS_PY2:
     import httplib
+    if sys.version_info[1] == 6:
+        import copy_reg
```
piskvorky: @shaform What is this cleverness?

We try not to write code that is too clever, and if we must, we comment it heavily.

Please include detailed comments about the motivation and behaviour of these blocks of code.

shaform: @piskvorky 
How about this: https://github.com/shaform/smart_open/commit/3154b060ef6e6388b76848ccb23ac029d0b183a4.

piskvorky: Isn't it easier and infinitely more readable to just create a named function?

Named functions (at the module level) are pickleable without a problem.

shaform: @piskvorky @tmylk I've put the earlier implementation here https://github.com/shaform/smart_open/tree/retry_alt

I could send another pull request if you like.

tmylk: Thanks @shaform. Simplified version merged in #118

PR #105, opened 2017-01-20
BUG: include tests data in sdist + install them.

cournape:
This allows running the tests from an installed smart_open without referring to the sources at all (e.g. wheel).


tmylk:
Thanks for the fix!

PR #106, opened 2017-02-06
HdfsOpenWrite implementation similar to read

skibaa:
I used the same idea which works for HdfsOpenRead to implement the write operation.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -167,7 +167,7 @@ def smart_open(uri, mode="rb", **kw):
             if mode in ('r', 'rb'):
                 return HdfsOpenRead(parsed_uri, **kw)
             else:
-                raise NotImplementedError("file mode %s not supported for %r scheme", mode, parsed_uri.scheme)
+                return HdfsOpenWrite(parsed_uri, **kw)
         elif parsed_uri.scheme in ("webhdfs", ):
             if mode in ('r', 'rb'):
                 return WebHdfsOpenRead(parsed_uri, **kw)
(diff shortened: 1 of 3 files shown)


tmylk:
Looks promising. 
May I ask you to [update](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/tests/test_smart_open.py#L458) existing tests and add new ones like for [read with mocking]
(https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/tests/test_smart_open.py#L141)

skibaa:
Done

smart_open/tests/test_smart_open.py
```diff
@@ -486,6 +485,19 @@ def test_s3_mode_mock(self, mock_write, mock_boto):
         mock_boto.connect_s3().get_bucket.assert_called_with("mybucket")
         self.assertTrue(mock_write.called)
 
+    @mock.patch('smart_open.smart_open_lib.subprocess')
+    def test_hdfs(self, mock_subprocess):
+        """Is HDFS write called correctly"""
+        smart_open_object = smart_open.HdfsOpenWrite(smart_open.ParseUri("hdfs:///tmp/test.txt"))
```
tmylk: Woud it be possible to check on ` smart_open.smart_open("hdfs:///some/file.txt", "wb")`?

smart_open/smart_open_lib.py
```diff
@@ -490,6 +492,33 @@ def __exit__(self, type, value, traceback):
         pass
 
 
+class HdfsOpenWrite(object):
+    """
+    Implement streamed reader from HDFS, as an iterable & context manager.
```
tmylk: You mean `Implement streamed writer to HDFS, as a context manager.`?

smart_open/tests/test_smart_open.py
skibaa: I took example from the HdfsOpenRead, and asked myself why it was called directly and not via smart_open.smart_open. And answered to myself that in the unit test it is preferable to check the most direct API - the smart_open.smart_open, the call to ParseUri and branching according to the schema is tested elsewhere.

tmylk: Yes, agree that HdfsOpenRead doesn't have this test. There is benefit in adding a test like that though to replace the exception test. As an aside, an ideal would be to mock hdfs similarly to [s3 tests]( https://github.com/skibaa/smart_open/blob/0fe9e521b5f7cf3589f6c5b2c537acec875a765d/smart_open/tests/test_smart_open.py#L251)

skibaa: I did not understand - what is the outcome. Do you want to leave it as is or to replace the call to HdfsOpenWrite ctor to the smart_open call?

tmylk: I would like to add tests for `smart_open.smart_open("hdfs:///some/file.txt", "wb")` and `smart_open.smart_open("hdfs:///some/file.txt", "rb")`

skibaa: It is a good idea to add these tests. But IMO it's better to leave this pull request focus on the write operation alone, as it was from the beginning and add new tests for smart_open.smart_open on a separate pull request.

tmylk:
Ok, Thanks for the PR!

Issue #109, opened 2017-03-16
Passing File Object Not Filename To GzipFile and BZ2File


simonseed:
Here

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L626

and here

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L630

Should be passing in the filename not the file object.

tmylk:
Reproduced. Working on a fix.

For completeness, could you paste a code snippet that breaks for you in this version.

cs48-project:
Hi.

I am also seeing this bug when attempting to use gensim, which uses smart_open to open gz compressed files... here is a minimal reproduction of the issue:
```

➜  /tmp virtualenv venv
New python executable in venv/bin/python
Installing setuptools, pip...done.
➜  /tmp source venv/bin/activate
(venv)➜  /tmp pip install -U pip smart_open
You are using pip version 6.0.8, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting pip from https://pypi.python.org/packages/b6/ac/7015eb97dc749283ffdec1c3a88ddb8ae03b8fad0f0e611408f196358da3/pip-9.0.1-py2.py3-none-any.whl#md5=297dbd16ef53bcef0447d245815f5144
  Using cached pip-9.0.1-py2.py3-none-any.whl
Collecting smart-open
  Using cached smart_open-1.5.0.tar.gz
Collecting boto>=2.32 (from smart-open)
  Using cached boto-2.46.1-py2.py3-none-any.whl
Collecting bz2file (from smart-open)
  Using cached bz2file-0.98.tar.gz
Collecting requests (from smart-open)
  Using cached requests-2.13.0-py2.py3-none-any.whl
Installing collected packages: requests, bz2file, boto, smart-open, pip

  Running setup.py install for bz2file

  Running setup.py install for smart-open
  Found existing installation: pip 6.0.8
    Uninstalling pip-6.0.8:
      Successfully uninstalled pip-6.0.8

Successfully installed boto-2.46.1 bz2file-0.98 pip-9.0.1 requests-2.13.0 smart-open-1.5.0
(venv)➜  /tmp 
(venv)➜  /tmp echo 'test text' | gzip > test_text.gz
(venv)➜  /tmp zcat test_text.gz
test text
(venv)➜  /tmp python
[...]

PR #112, opened 2017-03-17
Re-use the open filehandle rather than open a new one.

robottwo:
This is a permanent fix to the issue in #109 and #110.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 - "3.3"
 - "3.4"
 - "3.5"
+- "3.6"
 install:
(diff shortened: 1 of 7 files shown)


tmylk:
New unit tests for compressed files via http are needed.

robottwo:
OK -- the lastest version add unit tests for .gz and .bz2 over HTTP. It also fixes a couple of issues discovered by those tests.

tmylk:
Thanks for the fix. Will merge as is. 
Will create another wishlist item to test for the absence of file handle leak.

PR #115, opened 2017-03-18
Fixing bug #95

agent-jay:
The read_key.close() was unnecessary, can be removed. More importantly, how did the unit test for seek pass before this

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -465,7 +465,6 @@ class S3OpenRead(object):
         """
         if whence != 0 or offset != 0:
             raise NotImplementedError("seek other than offset=0 not implemented yet")
-        self.read_key.close(fast=True)
         self._open_reader()
 
     def __enter__(self):


tmylk:
Thanks. Could you please first submit a fix to the unit tests? They should fail with current code and pass with this fix.

agent-jay:
So this doesn't fix the original issue- that *readline()* doesn't work after seek(0). It is true however that the read_key.close(fast=True) isn't doing anything. Documenting some interesting behavior that expands on the original bug: 
```
g= smart_open.smart_open('s3://key_id:key_secret@smartopentest/valid_text.txt')
g.read() #works

g.seek(0)
g.read() #works

g.seek(0)
g.read(-1)#works

g.seek(0)
g.readline() #doesn't work

g.seek(0)
g.read(40) #doesn't work

```
I guess the question is what makes reading the entire file in one go different from reading bytes/lines at a time. This [PR](https://github.com/RaRe-Technologies/smart_open/pull/100) doesn't answer the question but seems to get the job done.
Then unit test for seek did not cover the above cases, only read(0) and read (-1) (the two are equivalent) after seek(0)

tmylk:
Please submit a PR that contais only the new unit tests that you wrote above that fail. Then add a second commit that fixes them.

agent-jay:
Dug into the S3ReadStream code so I could expand the test coverage if needed. Found the bug! Below is the fix

```
    def read(self, size=None):
        if not size or size < 0:
            return self.read_from_buffer(
                len(self.unused_buffer)) + self.read_until_eof()

        # Use unused data first
        if len(self.unused_buffer) >= size:
            return self.read_from_buffer(size)
##############################################################
        # If the stream is finished and no unused raw data, return what we have
        #if self.stream.closed or self.finished:
        #    self.finished = True
##############################################################
         if self.finished:
            return self.read_from_buffer(size)

        # Consume new data in chunks and return it.
        while len(self.unused_buffer) < size:
            raw = self.stream.read(io.DEFAULT_BUFFER_SIZE)
            if len(raw):
                self.unused_buffer += raw
            else:
                self.finished = True
                break

        return self.read_from_buffer(size)
```
self.stream.closed doesn't mean what the code would like it to mean. I wasn't able to find an equivalent in [boto's documentation](http://boto.cloudhackers.com/en/latest/ref/s3.html#module-boto.s3.key); any recommendations would be welcome. I've removed it for now, so it should at least work.
 
Also there is an attribute self.closed that's unreferenced, it's possible that it was the variable to condition on instead of self.stream.closed.

menshikh-iv:
Already implemented in current master fbc82cc04bf92e8c588c49c43ff5aff8234ea87d

PR #116, opened 2017-03-19
Fix bug #95 and expand test coverage to catch it

agent-jay:
More info [here](https://github.com/RaRe-Technologies/smart_open/pull/115) and [here](https://github.com/RaRe-Technologies/smart_open/issues/95)

A couple things that should be changed in the future:
- An unreferenced attribute self.closed should be removed or,
- self.stream.closed (removed in this commit) needs to be replaced with an alternative that actually does what it was intended to do. And maybe it's functioning is tied to the above attribute

Also, this will be my first (successful :crossed_fingers:)  PR :smile:

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -353,8 +353,7 @@ class S3ReadStreamInner(object):
             return self.read_from_buffer(size)
 
         # If the stream is finished and no unused raw data, return what we have
-        if self.stream.closed or self.finished:
-            self.finished = True
+        if self.finished:
             return self.read_from_buffer(size)
 
         # Consume new data in chunks and return it.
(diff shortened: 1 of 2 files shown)


agent-jay:
Spoke too soon, build needs to be restarted for some reason :p

tmylk:
Do the unit tests really fail before the fix? Can't see travis build for it

agent-jay:
I tested it locally, it did. I've never used travis before, would have made more sense to make two separate PRs

tmylk:
The `closed` attribute should function the same way as in `IOBase` https://docs.python.org/3.6/library/io.html

agent-jay:
The `closed` parameter returns True if the stream is closed, so it should poll the boto.s3.key object in some way? Basically what `stream.closed` was trying to do?

tmylk:
Yes

tmylk:
Please add tests similar to #117  that assert that files still are closed even after your changes to `seek` .  and that read finishes when stream is closed.

menshikh-iv:
Ping @agent-jay, what's a status of this PR?

agent-jay:
Sorry haven't had time to work on it.

menshikh-iv:
I close this PR because this is abandoned.

PR #121, opened 2017-04-25
Remove URL parameters before evaluating the file name.

### LINKED ISSUE #120: Gzipped URLS with parameters don't match the .endswith()
I have a pre-signed S3 URL for a gzipped xml file. 

https://bucket.s3.amazonaws.com/file.xml.gz?AWSAccessKeyId=asdf&Expires=2493144554&Signature=qasdf

The if fname.endswith('.gz') won't evaluate to True because of the URL parameters. 

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L759 

I suggest removing all query parameters when finding the file name.

mcrowson:
Closes #120

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -1034,3 +1034,4 @@ class WebHdfsException(Exception):
     def __init__(self, msg=str()):
         self.msg = msg
         super(WebHdfsException, self).__init__(self.msg)
+
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -754,7 +754,10 @@ def HttpOpenRead(parsed_uri, mode='r', **kwargs):
 
     response = HttpReadStream(url, **kwargs)
 
-    fname = url.split('/')[-1]
+    last_sep = url.split('/')[-1]
+
+    # Remove any URL parameters
+    fname = last_sep.split('?')[0]
 
     if fname.endswith('.gz'):
         #  Gzip needs a seek-able filehandle, so we need to buffer it.
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -1034,4 +1034,3 @@ class WebHdfsException(Exception):
     def __init__(self, msg=str()):
         self.msg = msg
         super(WebHdfsException, self).__init__(self.msg)
-
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -754,14 +754,11 @@ def HttpOpenRead(parsed_uri, mode='r', **kwargs):
 
     response = HttpReadStream(url, **kwargs)
 
-    last_sep = url.split('/')[-1]
-
-    # Remove any URL parameters
-    fname = last_sep.split('?')[0]
+    fname = urlsplit(url, allow_fragments=False).path.split('/')[-1]
 
     if fname.endswith('.gz'):
         #  Gzip needs a seek-able filehandle, so we need to buffer it.
-        buffer =  make_closing(io.BytesIO)(response.binary_content())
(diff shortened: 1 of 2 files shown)


tmylk:
Thanks for your contribution @mcrowson !

Would it be possible to add a test for this fix?

Also, the string manipulation is not needed as `parsed_uri.uri_path` already contains the path without parameters. For an unknown reason it is overwritten [here](https://github.com/mcrowson/smart_open/blob/c3a925a9329c7e6cd66bf5a8068722625332f1af/smart_open/smart_open_lib.py#L307), but that should probaby be removed without consequences.

mcrowson:
Sure, I'll write a test and go with that approach. Commits should be incoming later today.

mcrowson:
It looks like it is being overwritten, because the full URI needs to be passed to the HttpOpenRead function. My presigned file for example wouldn't work without the parameters. Because the function exists outside of the smart_open class we don't have access to self either to see the parsed_uri object.

I think just parsing the URL again within that function is the cleanest and most consistent option. Either that or the string function, or finding a way to pass the parsed URL from [here](https://github.com/mcrowson/smart_open/blob/c3a925a9329c7e6cd66bf5a8068722625332f1af/smart_open/smart_open_lib.py#L246) to the function.

smart_open/smart_open_lib.py
```diff
@@ -754,11 +754,11 @@ def HttpOpenRead(parsed_uri, mode='r', **kwargs):
 
     response = HttpReadStream(url, **kwargs)
 
-    fname = url.split('/')[-1]
+    fname = urlsplit(url, allow_fragments=False).path.split('/')[-1]
```
tmylk: May I ask about the reason for the `.split('/')[-1]`?
Would just a `if split_url.path.endswith('.gz')` check be enough?

mcrowson:
No, the issue this is related to is because there are URL parameters which means the .gz is in the middle of the string.

smart_open/smart_open_lib.py
mcrowson: No, the issue this is related to is because there are URL parameters which means the .gz is in the middle of the string.

mcrowson:
Not sure why travis timed out at Py3.6, works locally on 3.6. Closing and reopening for it to try the build again.

smart_open/smart_open_lib.py
tmylk: Sorry, would need an example. In what I tried path is the URL without the parameters.

menshikh-iv: Why you use `allow_fragments=False` ? I get the same result with/without this flag

```python
In [13]: urlsplit('http://127.0.0.1/data.gz', allow_fragments=False)
Out[13]: SplitResult(scheme='http', netloc='127.0.0.1', path='/data.gz', query='', fragment='')

In [14]: urlsplit('http://127.0.0.1/data.gz', allow_fragments=True)
Out[14]: SplitResult(scheme='http', netloc='127.0.0.1', path='/data.gz', query='', fragment='')

In [15]: urlsplit('http://127.0.0.1/data.gz?a=3,b=4', allow_fragments=False)
Out[15]: SplitResult(scheme='http', netloc='127.0.0.1', path='/data.gz', query='a=3,b=4', fragment='')

In [16]: urlsplit('http://127.0.0.1/data.gz?a=3,b=4', allow_fragments=True)
Out[16]: SplitResult(scheme='http', netloc='127.0.0.1', path='/data.gz', query='a=3,b=4', fragment='')
```

mcrowson: I was trying to stay consistent with the previous use of the urlsplit function 

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L246

piskvorky:
Thanks a lot @mcrowson ! Clean job.

Issue #64, opened 2016-03-07
S3OpenRead doesn't work with TextIOWrapper


mjwillson:
Currently only binary-mode read is supported, which would be OK if the stream you returned worked with `TextIOWrapper` -- but it doesn't, at least not `S3OpenRead`. `TextIOWrapper` expects readable, writable, seekable, closed etc methods. If you make S3OpenRead inherit (or quack like) `IOBase` that should fix it.

piskvorky:
I'm not familiar with hacking these classes. If you send a PR, we can add that support.

Is it possible to add it globally to all readers? (not just `S3OpenRead`)

tmylk:
Ping @mjwillson

mjwillson:
Hey, sorry I'm unlikely to get around to this as it stands -- we've worked around it like so, although this won't work for everyone:

```python
            # smart_open's S3OpenRead doesn't support TextIOWrapper
            # and smart_open doesn't support text-mode open. So
            # we return a poor-man's line-by-line TextIOWrapper:
            return (str(line, "utf-8") for line in text_stream)
```

To be honest smart_open feels like quite a leaky abstraction due to patchy / partial support for various features, meaning you need extra special-cased wrapping code after parsing the URL yourself. E.g. gzip only working for local files, other compression formats requiring manual wrapping, text mode not working for S3, https issues. Maybe some of those are fixed now, but if I was to revisit I'd be tempted to just manually wrap the streams from the underlying libraries myself.

If you wanted to address that in a deeper way, I'd suggest making sure any stream implementations here respect the stream interfaces from the standard library ( https://docs.python.org/3/library/io.html ), which would make it easier to write general and composable code that builds on them. Perhaps some of that would be easier if you give up on Python 2. Then try and implement features like compression and text mode support in a general way where possible via wrapping streams, and make the dispatch mechanisms (for protocols, compression formats etc) extensible.

tmylk:
Thanks for the idea of stream interfaces. Leaving open as it is an interesting direction to explore.

mpenkov:
This is related to #91.

mpenkov:
@menshikh-iv This is also done (PR #131 fixed it).

menshikh-iv:
great @mpenkov, fixed by #131

Issue #43, opened 2015-12-01
Migrate to boto3


piskvorky:
Check if boto3 is production-ready and [migrate all code from boto to boto3](http://boto3.readthedocs.org/en/latest/guide/migration.html).

anupamme:
I tried running smart_open on ubunty (15.10) and when I run python3.5 setup.py test I get following error message

pkg_resources.DistributionNotFound: The 'httpretty>=0.8.8' distribution was not found and is required by the application

And I am not able to install httpretty 0.8.8 or higher version on this machine (I can get 0.8.6 version running by directly downloading the deb package). I opened a ticket (https://github.com/gabrielfalcao/HTTPretty/issues/269)

So I am not sure how to go about fixing this. Any help will be appreciated.

output of uname -a:

Linux instance-8 4.2.0-18-generic #22-Ubuntu SMP Fri Nov 6 18:25:50 UTC 2015 x86_64 x86_64 x86_64 GNU/Linu

asieira:
I would suggest that event if boto3 is supported, existing implementation using boto 2 is kept as well.

There are many things boto3 hasn't implemented yet. To give one example, the lack of [explicit proxy support](https://github.com/boto/boto3/issues/338) would prevent me from continuing to use smart_open if it went boto3-only. See #35 for supporting evidence. :)

piskvorky:
Thanks @asieira , that's the kind of practical feedback we need!

I don't have experience with boto3, and certainly wouldn't want to inconvenience users by a hasty switch.

brunoalano:
+1

vinayan3:
+1. I'm having issues with Python 3.5 and Boto 2.39. From what I've read Python 3x is better with boto3

tmylk:
@vinayan3 What kind of issues are you having? Would be great to check they are resolved in boto3.

tmylk:
Update on Boto3 : Proxy support still not implemented in https://github.com/boto/boto3/issues/338

tmylk:
Our test dependeny moto has moved to boto3 unfortunately in ver 1.0.0

> * The normal @mock_<service> decorators will no longer work with boto. It is suggested that you upgrade to boto3 or use the standalone-server mode. If you would still like to use boto, you must use the @mock_<service>_deprecated decorators which will be removed in a future release.

combinatorist:
+1

- I came in at boto3. 
- Currently, I use this library by string interpolating keys into paths
[...]

Issue #42, opened 2015-11-24
s3 seek


perrygeo:
If I try to use smart open to seek/read parts of an s3 file, I get `NotImplementedError: seek other than offset=0 not implemented yet`.

Arbitrary seeking, especially when the seek was specified relative to the beginning of the file (`seek(..., whence=0)`, should be possible through the `Range` HTTP header

```
>>> import boto
>>> s3 = boto.connect_s3()
>>> bucket = s3.lookup('bucket')
>>> key = bucket.lookup('key')
>>> parts = key.get_contents_as_string(headers={'Range' : 'bytes=12-24'})
```

`seek` could establish a pointer to the starting byte and subsequent `read`s would define the end.

Are there any technical limitation or design restrictions that would prevent this?

piskvorky:
No, I think it's doable, and useful. Only no one did it yet.

perrygeo:
`get_contents_as_string` *should* respect the HTTP Range header but it doesn't always behave that way. In particular, I found that the first call read the entire contents while subsequent calls (with the exact same args and kwargs) pulled in only the requested bytes. I believe this to be a bug in boto. Unfortunately I couldn't find another way to implement this in boto2.

However, switching to boto3 I was able to put together a working s3 reader using the object abstractions. I wrapped it in a file handle interface that does arbitrary seeks and reads: https://gist.github.com/perrygeo/9239b9ab64731cacbb35#file-s3reader-py . It's very effective and allowed me to read TIF tags off 2000 x 1.1 GB files stored on S3 in just a few minutes.

I haven't yet considered how something like this could integrate with smart_open but I figured it might be useful.

piskvorky:
Ah, great. `get_contents_as_string` didn't work when I tried, but your analysis is clearly more exhaustive :)

As long as it supports streamed reads and writes (large files), it's a welcome addition! I think we'd have to implement buffering internally (for the `for line in x:` line-streaming interface, looking for newlines), but that shouldn't be too difficult. The only tricky part is interaction between stream buffering and `read`: user does `seek`, then `read` vs. user does `seek`, then iterates a few lines, then does `read`...

What does switching to boto3 entail? I vaguely remember issues with that library (unfinished), but that was some time ago.

tmylk:
Blocked by #43

mpenkov:
@menshikh-iv I think this is done. We can seek S3 files now.

menshikh-iv:
[...]

Issue #74, opened 2016-06-02
S3OpenRead ignores the mode of `'r'` opens always in `'rb'`


vinayan3:
The `S3OpenRead` always returns data as `bytes` in Python 3 which breaks things expecting `str`. The `S3OpenRead` needs to respect `'r'` instead of returning everything as `bytes`.

appierys:
It appears that now `smart_open` supports `'r'` mode in 1.5.4?
However, I found that the speed of reading files drops significantly after smart_open 1.5.4.

For smart_open 1.5.3, the speed is:
`2868923it [01:05, 43987.90it/s]`

while smart_open 1.5.4 is:
`41588it [01:35, 434.05it/s]`
for the same S3 file.

menshikh-iv:
Already fixed (doesn't reproduce with current master fbc82cc04bf92e8c588c49c43ff5aff8234ea87d

PR #131, opened 2017-08-17
Create separate compression-specific layer to enable writing gzipped files

mpenkov:
This PR addresses issue #91. There are two major parts to this PR.

First, I rewrote all the lower-level S3 stuff using boto3 in a separate module, smart_open/s3.py. he reader and writer inherit from the [io classes in the standard library](https://docs.python.org/3/library/io.html), meaning that you can pretty much treat them like local file streams and pass them to codecs like gzip, bz2, etc. 

Second, I integrated this module into smart_open_lib.py. This part required a rebase: I started working on this over a year ago, and much has happened since then. The part I couldn't handle well during the rebase was: there was a new `s3u` mode added, with code that depends on the old boto library. I couldn't port this part to boto3 - this functionality may be [discontinued](https://github.com/boto/boto3/issues/334). The code is isolated in the `setup_unsecured_mode` function and is currently commented out.

This takes care of #91 for S3 only. In the future, we should apply the same approach to other protocols like HDFS, more specifically: make sure readers and writers inherit from stdlib io classes.


mpenkov:
I forgot to mention that this also resolves #42 (s3 seek), and partially resolves #43 (starts boto3 migration).

I'm not 100% sure, but it should also resolve #64 and #74.

piskvorky:
Epic! @menshikh-iv let's review ASAP.

We should probably clean up and merge the smaller PRs first, this will be a major update.

@mpenkov what kind of backward incompatibilities do you expect here?

mpenkov:
@piskvorky I don't really understand the s3u stuff that was added in the meanwhile. Since that's currently effectively disabled in this branch, there is a chance of an incompatibility there.

Other than that, I don't expect these changes to break anything - they still pass pretty much the same tests and work the same way.

smart_open/s3.py
```diff
@@ -0,0 +1,373 @@
...
+
+import six
+
+
+_LOGGER = logging.getLogger(__name__)
+
+#
+# https://github.com/matthewwithanm/django-imagekit/commit/0d5bfe37517955cca2284735769fe1dffb38ed37
+#
+try:
```
menshikh-iv: Need to remove (we dropped 2.6 support)

smart_open/s3.py
```diff
@@ -0,0 +1,373 @@
...
+"""Implements file-like objects for reading and writing from/to S3."""
+import boto3
+
+import io
+import logging
+
+import six
+
+
+_LOGGER = logging.getLogger(__name__)
```
menshikh-iv: please use `logger` (everywhere)

smart_open/s3.py
```diff
@@ -0,0 +1,373 @@
...
+            raw = self._raw_reader.read(size=io.DEFAULT_BUFFER_SIZE)
+            if len(raw):
+                self._buffer += raw
+            else:
+                _LOGGER.debug('reached EOF while filling buffer')
+                self._eof = True
+
+        return self._read_from_buffer(size)
+
+    def read1(self, size=-1):
```
menshikh-iv: Why this needed here?

smart_open/s3.py
```diff
@@ -0,0 +1,373 @@
...
+        """This is the same as read()."""
+        return self.read(size=size)
+
+    def readinto(self, b):
```
menshikh-iv: Very strange, why we need this method with mutable `b`?

smart_open/s3.py
```diff
@@ -0,0 +1,373 @@
...
+        """Read up to len(b) bytes into b, and return the number of bytes
+        read."""
+        data = self.read(len(b))
+        if not data:
+            return 0
+        b[:len(data)] = data
+        return len(data)
+
+    def terminate(self):
+        """Do nothing."""
```
menshikh-iv: Add `pass`

smart_open/s3.py
mpenkov: It's part of the interface for [BufferedIOBase](https://docs.python.org/2/library/io.html#io.BufferedIOBase.read1).

smart_open/s3.py
mpenkov: Same as above. We try to behave like a [BufferedIOBase](https://docs.python.org/2/library/io.html#io.BufferedIOBase.readinto), which defines this method.

Some application downstream of smart_open may depend on this functionality.

mpenkov:
@menshikh-iv I made changes in response to your review. I left two items as they are because they are required to satisfy interface contracts. Let me know if I need to do anything else. Thanks!

menshikh-iv:
Thanks for clarification about `BufferIOBase`, great work, big thanks for your contribution @mpenkov :fire: :fire: :fire:

PR #140, opened 2017-10-28
Fix warnings

horpto:


smart_open/tests/test_smart_open.py
```diff
@@ -785,20 +787,23 @@ def write_read_assertion(self, test_file):
...
-        self.write_read_assertion(test_file)
+        with tempfile.NamedTemporaryFile('wb', suffix='.gz', delete=False) as file:
+            test_file_name = file.name
+        self.write_read_assertion(test_file_name)
 
     def test_write_read_bz2(self):
         """Can write and read bz2?"""
-        test_file = tempfile.NamedTemporaryFile('wb', suffix='.bz2', delete=False).name
-        self.write_read_assertion(test_file)
+        with tempfile.NamedTemporaryFile('wb', suffix='.bz2', delete=False) as file:
```
menshikh-iv: small nitpick - can you rename `file` (because this is type in python), for example, use `infile` (here and anywhere)

smart_open/smart_open_lib.py
```diff
@@ -497,19 +497,32 @@ def __exit__(self, type, value, traceback):
...
+
+    __exit__ = getattr(base, '__exit__', None)
+
+    def exit_attr(self, type, value, traceback):
+        if __exit__ is not None:
+            __exit__(self, type, value, traceback)
+        else:
+            self.close()
+        if under_stream is not None:
+            under_stream.close()
```
menshikh-iv: this shouldn't be called twice ? (connected with comment after)

smart_open/smart_open_lib.py
```diff
@@ -530,11 +543,11 @@ def compression_wrapper(file_obj, filename, mode):
             from bz2file import BZ2File
         else:
             from bz2 import BZ2File
-        return make_closing(BZ2File)(file_obj, mode)
+        return make_closing(BZ2File, file_obj)(file_obj, mode)
```
menshikh-iv: looks slightly strange ` file_obj)(file_obj`

smart_open/smart_open_lib.py
menshikh-iv: oh, my bad, sorry, you are correct.

smart_open/smart_open_lib.py
horpto: Tried redo as just classes

Issue #145, opened 2017-11-20
Add tests around encoding handling for various readers


mpenkov:
smart_open supports a variety of reader protocols:

- s3
- hdfs
- webhdfs
- http
- any others?

Currently, each of those implements its own logic for handling character encoding. This logic isn't being tested anywhere.

We should add tests to test that this logic works. In the future, we may refactor such that the encoding logic gets handled in a single place for all readers, so these tests should face the public API, as opposed to something internal.

Issue #146, opened 2017-11-24
HDFS, WebHDFS, HTTP I/O is not encoding-aware


mpenkov:
The HDFS and Http subsystems does not do any text encoding or decoding.

- HdfsOpenRead returns the stdout from a Popen, so it will be in whatever encoding Popen uses for pipes (could be UTF-8, or raw bytes)
- HdfsOpenWrite writes bytes directly into the stdin of a Popen
- WebHdfsOpenRead reads bytes directly (requests.get returns bytes when `stream=True`)
- HttpOpenRead also reads bytes directly (uses requests in the same manner as above).

mpenkov:
I think this is resolved by #185

menshikh-iv:
@mpenkov :+1:

PR #147, opened 2017-11-24
Resolve Issue #145, add warnings for Issue #146

mpenkov:
Issue 145: Added tests for reading and writing encoded text files on S3. Originally planned to add the same tests for the other subsystems (HDFS, WebHDFS, HTTP), but discovered that encoding is not handled at all there (it is ignored).

Issue 146: added warnings for cases where the encoding is specified and quietly ignored.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -28,6 +28,7 @@ import subprocess
 import sys
 import requests
 import io
+import warnings
 
 from boto.compat import BytesIO, urlsplit, six
 import boto.s3.connection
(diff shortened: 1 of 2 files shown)


smart_open/smart_open_lib.py
```diff
@@ -157,20 +168,26 @@ def smart_open(uri, mode="rb", **kw):
         elif parsed_uri.scheme in ("s3", "s3n", 's3u'):
             return s3_open_uri(parsed_uri, mode, **kw)
         elif parsed_uri.scheme in ("hdfs", ):
+            if kw.get('encoding') is not None:
+                warnings.warn('Issue #146: encoding is not supported for HDFS, ignoring')
```
piskvorky: Let's be a little more verbose. Why is not supported -- is it a fault of the user, or a functionality gap in `smart_open`? What can the user do about this? (how to fix the warning?)

mpenkov: OK, done. Please have a look.

menshikh-iv:
Thank you @mpenkov :+1:

PR #149, opened 2017-11-26
Added support for passing custom s3 client

creatorrr:


mpenkov:
@creatorrr Thank you for making the pull request. While in general the change looks OK, I have a few questions:

- What is the use case for passing a custom S3 client?
- Is there a benefit to including this as a parameter, as opposed to mocking it?
- I notice that the top-level interface has not changed. How does the user pass this new parameter?

menshikh-iv:
Ping @creatorrr

yganti87:
Hello @mpenkov - One case I ran into where a custom s3 client needed to be passed was for cross account access using STS. I couldn't find a way of doing this in the current version.

eschwartz:
Looks like this PR has gone stale, but I also need this functionality (see #199)
Give me a couple days, but I'm going to see if I can get this finished up.

piskvorky:
Fixed by #201.

Issue #152, opened 2017-12-01
Reading S3 files becomes slow after 1.5.4


appierys:
As mentioned earlier in #74, it appears that the reading speed is very slow after 1.5.4.

```bash
$ pyvenv-3.4 env
$ source env/bin/activate
$ pip install smart_open==1.5.3 tqdm ipython
$ ipython
```
```python
from tqdm import tqdm
from smart_open import smart_open
for _ in tqdm(smart_open('s3://xxxxx', 'rb')):
    pass
```
2868923it [00:53, 53888.94it/s]

```bash
$ pyvenv-3.4 env
$ source env/bin/activate
$ pip install smart_open==1.5.4 tqdm ipython
$ ipython
```
```python
from tqdm import tqdm
from smart_open import smart_open
for _ in tqdm(smart_open('s3://xxxxx', 'rb')):
    pass
```
8401it [00:18, 442.64it/s] (too slow so I could not wait for it to finish.)

menshikh-iv:
Thanks for report @appierys, that's really suspicious, we'll investigate it, related issue #151.

mpenkov:
I'm looking into this now.

mpenkov:
I've reproduced the issue, but the performance gap between 1.5.3 and .4 wasn't as large as originally reported. @appierys, could you tell us the extension of the S3 file you were trying to open? The extension influences how smart_open actually opens the file.

Here are my efforts to reproduce the bug:

```
(so1.5.3)sergeyich:issue152 misha$ time python reproduce.py s3://commoncrawl/cc-index/collections/CC-MAIN-2013-20/indexes/cluster.idx
705it [00:01, 383.09it/s]

real    0m3.395s
user    0m0.233s
sys     0m0.044s
(so1.5.3)sergeyich:issue152 misha$ deactivate
sergeyich:issue152 misha$ source so1.5.4/bin/activate
(so1.5.4)sergeyich:issue152 misha$ time python reproduce.py s3://commoncrawl/cc-index/collections/CC-MAIN-2013-20/indexes/cluster.idx
977it [00:15, 64.74it/s]
real    0m18.065s
user    0m0.813s
sys     0m0.164s
(so1.5.4)sergeyich:issue152 misha$
```

1.5.4 appears to be around 6 times slower than 1.5.3.

reproduce.py:

```python
import sys

from tqdm import tqdm
from smart_open import smart_open
[...]

PR #155, opened 2017-12-02
Fix issues #152, #153 and #154

### LINKED ISSUE #153: Unable to iterate over gzipped object on S3
Encountered while reproducing issue #152.

Given reproduce.py:

```python
import sys

from tqdm import tqdm
from smart_open import smart_open
for i, _ in enumerate(tqdm(smart_open(sys.argv[1], 'rb'))):
    if i > 1e5:
        break
```

This reproduces the bug:

```
(smartopen)sergeyich:issue152 misha$ git rev-parse --short HEAD
d10166c
(smartopen)sergeyich:issue152 misha$ time python reproduce.py s3://commoncrawl/crawl-002/2010/09/25/0/1285411480200_0.arc.gz
0it [00:00, ?it/s]Traceback (most recent call last):
  File "reproduce.py", line 5, in <module>
    for i, _ in enumerate(tqdm(smart_open(sys.argv[1], 'rb'))):
  File "/Users/misha/envs/smartopen/lib/python3.6/site-packages/tqdm/_tqdm.py", line 953, in __iter__
    for obj in iterable:
TypeError: 'closing' object is not iterable


real    0m3.131s
user    0m0.906s
sys     0m0.151s
```

### LINKED ISSUE #154: Don't raise an exception on failure to CreateBucket
```
  File "/opt/crash_uploader", line 165, in <module>
    upload_user_coredump(pid, executable)
  File "/opt/crash_uploader", line 123, in upload_user_coredump
    with smart_open(dumpfile, 'wb') as s3out:
  File "/usr/local/lib/python2.7/dist-packages/smart_open/smart_open_lib.py", line 175, in smart_open
    return s3_open_uri(parsed_uri, mode, **kw)
  File "/usr/local/lib/python2.7/dist-packages/smart_open/smart_open_lib.py", line 255, in s3_open_uri
    fobj = smart_open_s3.open(parsed_uri.bucket_id, parsed_uri.key_id, s3_mode, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/smart_open/s3.py", line 60, in open
    fileobj = BufferedOutputBase(bucket_id, key_id, min_part_size=s3_min_part_size, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/smart_open/s3.py", line 255, in __init__
    s3.create_bucket(Bucket=bucket)
  File "/usr/local/lib/python2.7/dist-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the CreateBucket operation: Access Denied
```

In general implicit bucket creation seems like a bad call, in case of a misconfig. But if it is going to happen, it definitely shouldn't block writing to that bucket if it already exists. (In our case this is running from an IAM box that does not have permission to create buckets, but can write to an existing bucket.)

Testing indicates this broke in 1.5.4

mpenkov:
We didn't have a readline in our S3 reader, and that was slowing us down.

Here is a breakdown of the contents:

- 448192e, fa7c183, f7a8c56: Address #152 by added a .readline method. Does not solve the problem entirely - the regression still exists, but it's significantly smaller than before.
- 6f8f7b8: responding to feedback
- d453af2 and 084f43e: avoid creating buckets if they do not exist (resolves #154)
- 8017985, a3f9fe1, 2188b2b: added handling for the errors keyword argument
- 6c22ed8, 3e93fb1, 14588a5: performed a refactoring for disabling seek functionality.  This improves speed under Py3, but breaks under Py2.7 because the gzip reader needs seek under that Python version. The last commit re-enables seek, but keeps the refactoring, in case we want to do more with it later.
- 2c79505 - fix #153

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -29,7 +29,6 @@ MODES = (READ, READ_BINARY, WRITE, WRITE_BINARY)
 """Allowed I/O modes for working with S3."""
 
 BINARY_NEWLINE = b'\n'
-TEXT_NEWLINE = b'\n'
 DEFAULT_BUFFER_SIZE = 256 * 1024
 
 
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -78,6 +78,21 @@ class RawReader(object):
     def __init__(self, s3_object):
         self.position = 0
         self._object = s3_object
+        self._body = s3_object.get()['Body']
+
+    def read(self, size=-1):
+        if size == -1:
+            return self._body.read()
+        return self._body.read(size)
+
+
+class SeekableRawReader(object):
+    """Read an S3 object.
+
+    Support seeking around, but is slower than RawReader."""
+    def __init__(self, s3_object):
+        self.position = 0
+        self._object = s3_object
         self._content_length = self._object.content_length
 
     def read(self, size=-1):
@@ -95,10 +110,6 @@ class RawReader(object):
 
 
 class BufferedInputBase(io.BufferedIOBase):
-    """Reads bytes from S3.
-
-    Implements the io.BufferedIOBase interface of the standard library."""
-
     def __init__(self, bucket, key, buffer_size=DEFAULT_BUFFER_SIZE,
                  line_terminator=BINARY_NEWLINE, **kwargs):
         session = boto3.Session(profile_name=kwargs.pop('profile_name', None))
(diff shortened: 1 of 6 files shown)


smart_open/s3.py
```diff
@@ -28,6 +28,10 @@
 MODES = (READ, READ_BINARY, WRITE, WRITE_BINARY)
 """Allowed I/O modes for working with S3."""
 
+BINARY_NEWLINE = b'\n'
+TEXT_NEWLINE = b'\n'
```
piskvorky: Both are binary?

smart_open/tests/test_s3.py
```diff
@@ -9,7 +9,7 @@
 else:
     import unittest
 
-import boto
+import boto3
```
piskvorky: This seems to be a big change; does it belong here?

smart_open/s3.py
mpenkov: You're right. We're not using TEXT_NEWLINE right now, so I removed it.

smart_open/tests/test_s3.py
mpenkov: @piskvorky It's not really _that_ big a change. This is test code, and it writes a mock object to a mock S3 bucket. The code for doing this with boto and boto3 is slightly different, but the end result is the same (the tests still pass without changing the code). If you look at the remainder of the changes in test_s3.py, you'll see the tests aren't strongly coupled to either boto or boto3.

The real benefit to using boto3 in tests is that it matches the implementation: our S3 implementation uses boto3 under the covers. Also, using boto3 is future-proof: newer versions of moto mock boto and boto3 separately. That is, objects mocked via boto are not visible to boto3 and vice versa. This means that if we upgrade to the most recent moto version (unlike the year-old version currently used in the test environment), our boto-based tests will break.

This change solves that problem before it happens.

menshikh-iv:
Great fixes, thanks a lot @mpenkov :1st_place_medal:

Issue #156, opened 2017-12-03
Get rid of code used only for Py2.6 compatibility


mpenkov:
Some parts of the code, like the closings around GzipFile and Bz2File, were only necessary while we supported Py2.6. Since we no longer support Py2.6, that code is not necessary. In the case of GzipFile and Bz2File, they are context managers in their own right since Py2.7, and the closing is redundant.

PR #158, opened 2017-12-13
[WIP] Add integration testing with real S3. Partial fix #151, #156

### LINKED ISSUE #151: Add integration and regression tests
Right now, for testing `smart_open` (local file part) - it is not enough to run only tests from `smart_open`, needs to run gensim tests with smart_open from `master` (so I caught a bug today - hotfix #150).

Need to look at `gensim` usage of `smart_open` and add tests according to this behavior (better - something more "general" for all readers/writers, because the tests are rather monotonous).

CC: @mpenkov

menshikh-iv:
Partial fix of #151 (for S3) and small cleanup according to #156.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ matrix:
     - python: '2.7'
       env:
         - SO_DISABLE_MOCKS: "1"
-        - SO_BUCKET_NAME: "test-smartopen-travis-ci-27"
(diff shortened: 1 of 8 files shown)


menshikh-iv:
What else needed

- [ ] Fix bench for py27 https://travis-ci.org/RaRe-Technologies/smart_open/jobs/316367528#L806
- [ ] Create uniq bucket for benchmark run + cleanup after

mpenkov:
4bf0e11 intentionally breaks the build to test that travis actually exits on error. 800b787 unbreaks it.

2dec2b4 is a mistake. a43541f corrects the mistake.

PR #164, opened 2017-12-22
Migrate to `boto3`. Fix #43

### LINKED ISSUE #43: Migrate to boto3

mpenkov:
Creating a new pull request to get around travis' secure data issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
 
 install:
   - pip install .[test]
-  - pip uninstall botocore boto3
+  - pip uninstall --yes botocore boto3
   - pip install boto3
   - pip freeze
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ matrix:
 
 install:
   - pip install .[test]
+  - pip uninstall botocore boto3
+  - pip install boto3
   - pip freeze
 
 
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -1,17 +1,29 @@
 # -*- coding: utf-8 -*-
 """Implements file-like objects for reading and writing from/to S3."""
-import boto3
-import botocore.client
 
 import io
+import contextlib
+import functools
+import itertools
 import logging
 
+import boto3
+import botocore.client
 import six
 
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+# Multiprocessing is unavailable in App Engine (and possibly other sandboxes).
+# The only method currently relying on it is s3_iter_bucket, which is instructed
+# whether to use it by the MULTIPROCESSING flag.
+_MULTIPROCESSING = False
+try:
+    import multiprocessing.pool
+    _MULTIPROCESSING = True
+except ImportError:
+    logger.warning("multiprocessing could not be imported and won't be used")
 
 START = 0
 CURRENT = 1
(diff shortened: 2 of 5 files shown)


smart_open/tests/test_smart_open.py
```diff
@@ -194,8 +195,9 @@ def test_open_with_keywords_explicit_r(self):
     @mock_s3
     def test_read_never_returns_none(self):
         """read should never return None."""
-        conn = boto.connect_s3()
-        conn.create_bucket("mybucket")
+        s3 = boto3.resource('s3')
+        s3.create_bucket(Bucket='mybucket')
```
menshikh-iv: Maybe, better to use "global" bucket /key (as in https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/tests/test_s3.py#L16) + same `maybe_mock_s3`?

smart_open/tests/test_smart_open.py
```diff
@@ -968,6 +975,9 @@ class S3OpenTest(unittest.TestCase):
     @mock_s3
     def test_r(self):
         """Reading a UTF string should work."""
+        #
```
menshikh-iv: What's your plan to do with this type of test (with `FIXME`)?

mpenkov:
@piskvorky @menshikh-iv Moving all tests across to boto3 requires moving some of our core functionality from boto to boto3. This affects two features:

1. smart_open accepts boto keys. If we switch to boto3, we could either drop this feature, or silently substitute incoming boto objects with boto3 ones.
2. s3_iter_bucket accepts a boto Bucket, and yields boto Keys. If we switch to boto3, we will have to do something about this feature, e.g. always accept and return boto3. We can also write two versions of the function, one for boto2 and one for boto3, but this is labor-intensive.

I can see three ways going forward. In my opinion, in order of decreasing subjective goodness:

1. Stop using boto/boto3 in the interface, and use native datatypes only, i.e. strings, to identify resources. This decouples our API from boto and boto3 - which particular library we use to do the work becomes purely an implementation detail.
2. Use boto3 only.
3. Use boto3, maintain boto functionality as separate.
4. Keep things as is.

I think 1) is a bit extreme, because it will surprise (anger) some people. 3) is poor, because it makes us do additional work to support a deprecated library, but marginally better than 4). I think 2) is a good compromise, because it isn't difficult to switch between boto and boto3 in application code.

What do you guys think?

menshikh-iv:
Thanks for detailed description @mpenkov, I'm +1 for (2) because we want to migrate to boto3 a long time ago, see #43, I see no reasons to support boto2 anymore (more than that, this project doesn't develop actively now https://github.com/boto/boto (they send us to boto3)).

piskvorky:
I agree with @menshikh-iv .

.travis.yml
```diff
@@ -30,6 +30,8 @@ matrix:
 
 install:
   - pip install .[test]
+  - pip uninstall --yes botocore boto3
```
menshikh-iv: better to edit `setup.py` as we discussed, I think this will work

smart_open/s3.py
```diff
@@ -1,17 +1,29 @@
...
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+# Multiprocessing is unavailable in App Engine (and possibly other sandboxes).
+# The only method currently relying on it is s3_iter_bucket, which is instructed
+# whether to use it by the MULTIPROCESSING flag.
+_MULTIPROCESSING = False
+try:
+    import multiprocessing.pool
+    _MULTIPROCESSING = True
```
menshikh-iv: why can `multiprocessing` be unavailable?

smart_open/tests/test_smart_open.py
```diff
@@ -672,168 +615,6 @@ def write_callback(request):
         assert responses.calls[3].request.url == "http://127.0.0.1:8440/file"
 
 
-class S3IterBucketTest(unittest.TestCase):
```
menshikh-iv: These tests have no more sense, am I right?

mpenkov: Yes. Most of them moved to the s3 module.

smart_open/s3.py
mpenkov: I trust the above comment. It may be unavailable in certain environments, but I've never been in one.

menshikh-iv:
@mpenkov what's missed here (especially what's about cases when this is impossible to replace old boto)? Is current PR ready to merge?

mpenkov:
@menshikh-iv Yes, I think it's ready.

menshikh-iv:
Great work @mpenkov :fire:

PR #168, opened 2018-01-02
[hdfs] Support host and port in HDFS URIs

### LINKED ISSUE #162: HDFS URI support does not support network location
It seems that "hdfs://host:port/path/file.txt" url scheme is not supported by smart open currently. Therefore, with smart_open it is only possible to access the local HDFS filesystem with the hdfs:// protocol. The URI is parsed correctly by the urlsplit, but the network location is always ignored.

The hdfs dfs command supports the host:port in the URI out of the box, so adding this would be very easy. However, then the url scheme "hdfs://tmp/test.txt" would no longer work as the 'tmp' here would be interpreted as the network location, which it really is if you read the URI literally according to specification.

For me it would be expected result that:
 - ParseUri('hdfs://host:port/path/file.txt', 'wb') ==> ["hdfs", "dfs", "-text", "hdfs://host:port/path/test.txt"]
 - ParseUri ('hdfs:///path/file.txt', 'wb') ==> ["hdfs", "dfs", "-text", "/path/test.txt"]
 - ParseUri ('hdfs://host/path/file.txt', 'wb') ==> ["hdfs", "dfs", "-text", "hdfs://host/path/test.txt"]

Let me know what you think.

vvaten:
Fixes #162.

The original PR by @vvaten correctly identified that smart_open's `ParseUri` mangled HDFS URIs in a way that's against [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986). Since 2018, HDFS parsing moved out of `smart_open_lib.py` into `smart_open/hdfs.py`, but the underlying bug came along for the ride: `parse_uri` still collapses `netloc + path` into a single slash-prefixed string, so `hdfs://host:8020/tmp/test.txt` becomes `/host:8020/tmp/test.txt` before reaching the `hdfs dfs` CLI — and the CLI no longer addresses the right cluster.

This PR ports the original intent against the current code: when `netloc` is non-empty, `parse_uri` returns the **full URI** so the CLI can resolve the cluster itself.

### Before vs after

| Input URI                         | `parse_uri(...)['uri_path']` (before) | `parse_uri(...)['uri_path']` (after) | Effective `hdfs dfs` call            |
|-----------------------------------|----------------------------------------|---------------------------------------|--------------------------------------|
| `hdfs:///tmp/test.txt`            | `/tmp/test.txt`                        | `/tmp/test.txt` *(unchanged)*         | default cluster, `/tmp/test.txt`     |
| `hdfs://host/tmp/test.txt`        | `/host/tmp/test.txt` 🐛                | `hdfs://host/tmp/test.txt`            | `host` cluster, `/tmp/test.txt`      |
| `hdfs://host:8020/tmp/test.txt`   | `/host:8020/tmp/test.txt` 🐛           | `hdfs://host:8020/tmp/test.txt`       | `host:8020` cluster, `/tmp/test.txt` |
| `viewfs://cluster/tmp/test.txt`   | `/cluster/tmp/test.txt` 🐛             | `viewfs://cluster/tmp/test.txt`       | `cluster` mount table resolves       |

`URI_EXAMPLES` is updated to reflect the supported shapes (the misleading `hdfs://path/file` form is removed in favor of `hdfs://host/path/file` and `hdfs://host:port/path/file`).

### Docs

- `help.txt` regenerated via `python update_helptext.py`.
- `README.rst`: replaced `hdfs://path/file` and `hdfs://user/hadoop/my_file.txt` with explicit `hdfs://host:port/...` forms.

### Tests

- New parametrized `test_parse_uri` / `test_parse_uri_invalid` in `tests/test_hdfs.py` covering the three netloc variants for `hdfs`/`viewfs` plus error cases.
- Existing `tests/test_smart_open.py::*::test_hdfs` mocks updated to assert the full-URI passthrough for both read and write paths.

Local run: `pytest tests/test_hdfs.py tests/test_smart_open.py -k hdfs` → 32 passed.


mpenkov:
Thank you for your pull request. It looks good to me.

@menshikh-iv Might be a good idea to merge this _after_ our HDFS integration tests are up. What do you think?

menshikh-iv:
@mpenkov I agree.
@vvaten thanks for PR! Sorry for waiting, but first we need to finish https://github.com/RaRe-Technologies/smart_open/issues/151.

smart_open/smart_open_lib.py
```diff
@@ -355,6 +356,9 @@ class ParseUri(object):
       * file:///home/user/file
       * file:///home/user/file.bz2
 
+    NOTE: hdfs://path/file does no longer work as it is against the URI
```
menshikh-iv: Can you add more information to comment (when this happens, what HDFS version affected, etc)?

vvaten: This is a generic change that affects all HDFS versions. They do support hdfs://host/path/file URI format. Using hdfs://path/file in smart_open violates this and also violates the URI specification (RFC3986) where the hostname part is always after the first two / characters. The correct way to refer to local content is hdfs:///path/file instead of hdfs://path/file.

govindmurthi21:
Hello is there any way this can be merged ? As part of my work I was trying to read and write to hdfs and traced the bug to this 
Hdfs commands require a fully qualified uri such as ‘’hdfs dfs ls hdfs://path’’

ddelange:
Hi @vvaten :wave:

I know it's been a while. It looks like this PR is still valid, only outdated. Do you want to pull `develop` and get this merged?

PR #175, opened 2018-03-14
Add support for `pathlib.Path`. Fix #170

### LINKED ISSUE #170: Support for opening `pathlib.Path`
Would you be willing include support for the standard library `Path` object?

```python
>>> from pathlib import Path
>>> from smart_open import smart_open

>>> path = Path('~/test.txt').expand_user().resolve()
>>> smart_open(path)

TypeError: don't know how to handle uri PosixPath('/home/user/test.txt')
```

I would be willing to create the PR! Thanks!

clintval:
This closes #170. First time contributing so please let me know if this needs fixing! Thanks!

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -102,6 +102,7 @@ def smart_open(uri, mode="rb", **kw):
     3. a URI for Amazon's S3 (can also supply credentials inside the URI):
        `s3://my_bucket/lines.txt`, `s3://my_aws_key_id:key_secret@my_bucket/lines.txt`
     4. an instance of the boto.s3.key.Key class.
+    5. an instance of the pathlib.Path class.
 
     Examples::
 
@@ -162,6 +163,14 @@ def smart_open(uri, mode="rb", **kw):
     # validate mode parameter
     if not isinstance(mode, six.string_types):
         raise TypeError('mode should be a string')
+
+    # Support opening ``pathlib.Path`` objects by casting them to strings.
+    try:
+        from pathlib import Path
+        if isinstance(uri, Path):
+            uri = str(uri)
+    except ImportError:
+        pass
 
     if isinstance(uri, six.string_types):
         # this method just routes the request to classes handling the specific storage
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -27,8 +27,19 @@ import os
 import subprocess
 import sys
 import requests
+import importlib
 import io
 import warnings
+
+# Import ``pathlib`` if the builtin ``pathlib`` or the backport ``pathlib2`` are
+# available. The builtin ``pathlib`` will be imported with higher precedence.
+for pathlib_module in ('pathlib', 'pathlib2'):
+    try:
+        pathlib = importlib.import_module(pathlib_module)
+        PATHLIB_SUPPORT = True
+        break
+    except ImportError:
+        PATHLIB_SUPPORT = False
 
 from boto.compat import BytesIO, urlsplit, six
 import boto.s3.connection
@@ -165,12 +176,8 @@ def smart_open(uri, mode="rb", **kw):
         raise TypeError('mode should be a string')
 
     # Support opening ``pathlib.Path`` objects by casting them to strings.
-    try:
-        from pathlib import Path
-        if isinstance(uri, Path):
-            uri = str(uri)
-    except ImportError:
-        pass
+    if PATHLIB_SUPPORT and isinstance(uri, pathlib.Path):
+        uri = str(uri)
 
     if isinstance(uri, six.string_types):
         # this method just routes the request to classes handling the specific storage
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -191,6 +191,16 @@ class SmartOpenReadTest(unittest.TestCase):
             actual = fin.read()
         self.assertEqual(expected, actual)
 
+    def test_open_pathlib_path(self):
+        """If ``pathlib.Path`` is available we should open it."""
+        try:
+            from pathlib import Path
+            fpath = Path(os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt'))
+            with smart_open.smart_open(fpath, encoding='cp852') as fin:
+                fin.read()
+        except ImportError:
+            pass
+
     @mock_s3
     def test_read_never_returns_none(self):
         """read should never return None."""
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -10,6 +10,7 @@ import unittest
 import logging
 import tempfile
 import os
+import pkgutil
 import sys
 import hashlib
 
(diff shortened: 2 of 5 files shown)


clintval:
Doesn't appear as if test failures are due to this patch FYI.

smart_open/smart_open_lib.py
```diff
@@ -163,6 +164,14 @@ def smart_open(uri, mode="rb", **kw):
     if not isinstance(mode, six.string_types):
         raise TypeError('mode should be a string')
 
+    # Support opening ``pathlib.Path`` objects by casting them to strings.
+    try:
+        from pathlib import Path
```
menshikh-iv: Oh, `pathlib` available only for `python>=3.4`, that's sad

smart_open/tests/test_smart_open.py
```diff
@@ -191,6 +191,16 @@ def test_open_with_keywords_explicit_r(self):
...
         self.assertEqual(expected, actual)
 
+    def test_open_pathlib_path(self):
+        """If ``pathlib.Path`` is available we should open it."""
+        try:
+            from pathlib import Path
+            fpath = Path(os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt'))
+            with smart_open.smart_open(fpath, encoding='cp852') as fin:
+                fin.read()
+        except ImportError:
```
menshikh-iv: That's a bad practice to silently pass the test, better to add a condition about python version for test https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

clintval:
Hi @menshikh-iv ready for another review.

Interestingly `pathlib.Path` was introduced in Python 3.4 but support for opening `Path` objects using the builtin `open()` was not introduced until Python 3.6. The test now supports these cases.

smart_open/smart_open_lib.py
clintval: Instead of checking for `python>=3.4` and assuming `pathlib` support I opted for the try-except pattern since there is a backport of `pathlib` that users may wish to use for `python<3.4`:

https://pypi.python.org/pypi/pathlib2/

smart_open/tests/test_smart_open.py
```diff
@@ -191,6 +191,16 @@ def test_open_with_keywords_explicit_r(self):
             actual = fin.read()
         self.assertEqual(expected, actual)
 
+    def test_open_pathlib_path(self):
+        """If ``pathlib.Path`` is available we should open it."""
+        try:
+            from pathlib import Path
+            fpath = Path(os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt'))
+            with smart_open.smart_open(fpath, encoding='cp852') as fin:
+                fin.read()
```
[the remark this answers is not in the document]
clintval: Hi @mpenkov this commit is outdated. There is a more recent commit addressing review from @menshikh-iv.

smart_open/smart_open_lib.py
menshikh-iv: Maybe add it to `setup.py`, wdyt @mpenkov?  https://github.com/RaRe-Technologies/smart_open/blob/ecb24c23f4ad58904d371d88eeb1f83389cae16e/setup.py#L19-L23

smart_open/tests/test_smart_open.py
```diff
@@ -191,6 +191,24 @@ def test_open_with_keywords_explicit_r(self):
             actual = fin.read()
         self.assertEqual(expected, actual)
 
+    @unittest.skipIf(
+        sys.version_info < (3, 4),
```
menshikh-iv: better to check that `pathlib` can be imported or not (not python version, because I can have installed `pathlib` from backport.

menshikh-iv:
@mpenkov suspicious dependency conflict for `3.3` in CI https://travis-ci.org/RaRe-Technologies/smart_open/jobs/353985032, any ideas?

@clintval please merge the latest master into PR.

clintval:
Ok @mpenkov, I now skip the unit test if the module `pathlib` cannot be found or imported.

I rebased onto `master` too.

smart_open/tests/test_smart_open.py
```diff
@@ -191,6 +192,24 @@ def test_open_with_keywords_explicit_r(self):
             actual = fin.read()
         self.assertEqual(expected, actual)
 
+    @unittest.skipIf(
+        pkgutil.find_loader('pathlib') is None,
```
menshikh-iv: wow, I don't know about `pkgutil`, @clintval  :+1:

menshikh-iv:
@clintval yes, add it (and change skipping condition, because of `pathlib` and `pathlib2` (2 different names, we need only one (any))

menshikh-iv:
btw, really suspicious fail on Travis for `3.3`, I'll investigate it later.

clintval:
Opening `pathlib.Path` and `pathlib2.Path` instances is now supported!

The tests reflect this support.

See these test logs for how this PR acts without `pathlib2` installed:
https://travis-ci.org/RaRe-Technologies/smart_open/builds/354849770

And these more recent logs for how this PR acts with `pathlib2` installed:
https://travis-ci.org/RaRe-Technologies/smart_open/builds/354850402

The only unsupported condition is if you have both `pathlib` and `pathlib2` backport installed, `smart_open` will only open paths from the builtin library and not from the backport. If you want concurrent `pathlib` and `pathlib2` backport support let me know and I can modify the conditions.

menshikh-iv:
@clintval I'm +1 for using built-in `pathlib` even `pathlib2` installed.

menshikh-iv:
@clintval I resolved CI problem, now all fine
LGTM for me, wdyt @mpenkov?

smart_open/tests/test_smart_open.py
```diff
@@ -193,16 +194,28 @@ def test_open_with_keywords_explicit_r(self):
...
+        for pathlib_module in ('pathlib', 'pathlib2'):
+            try:
+                pathlib = importlib.import_module(pathlib_module)
+                break
+            # Unit test will skip if either module is unavailable so it's safe
+            # to assume we can import _at least_ one working ``pathlib``.
+            except ImportError:
+                pass
 
         # builtin open() supports pathlib.Path in python>=3.6 only
```
mpenkov: I don't think we need to bother with this is in the tests. From the point of view of the test, it's an unnecessary detail. In this particular case, we just want to read the expected data in the quickest and simplest possible way. So in this case, I think it's better to just do:

```python
with open(fpath) as fin:
    expected = fin.read().decode('cp852')
```

smart_open/tests/test_smart_open.py
```diff
@@ -193,16 +194,28 @@ def test_open_with_keywords_explicit_r(self):
         self.assertEqual(expected, actual)
 
     @unittest.skipIf(
```
mpenkov: I think a simpler way of writing this is:

```python
@unittest.skipIf(smart_open_lib.PATHLIB_SUPPORT is False, "your reason here")
```

smart_open/tests/test_smart_open.py
```diff
@@ -193,16 +194,28 @@ def test_open_with_keywords_explicit_r(self):
...
+        (pkgutil.find_loader('pathlib') is None and
+         pkgutil.find_loader('pathlib2') is None),
+        "do not test pathlib support if pathlib or backport are not available")
     def test_open_and_read_pathlib_path(self):
-        """If pathlib.Path is available we should be able to open and read."""
-        from pathlib import Path
+        """If ``pathlib.Path`` is available we should be able to open and read."""
+        fpath = os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt')
+
+        # Import ``pathlib`` if the builtin ``pathlib`` or the backport
```
mpenkov: I think we're duplicating unnecessary work here. We've already gone through the pain of importing the necessary pathlib module in smart_open_lib. So, we can just get rid of all this import stuff, and use:

```python
from smart_open.smart_open_lib import pathlib
path = pathlib.Path('/foo/bar')
```

whenever you need access to the actual pathlib module.

clintval:
Thanks @mpenkov! Great review.

I completely agree and pushed another commit slimming the test down.

Here are the `Path` instances we support opening:

| Python Version | `pathlib2.Path` | `pathlib.Path`   |
| ---                     | :---:                  | :---:                  |
| <3.4                  |  ✔                  |                         |
| ≥3.4                  |                         | ✔                   |

smart_open/tests/test_smart_open.py
```diff
@@ -193,35 +191,16 @@ def test_open_with_keywords_explicit_r(self):
...
-        path_smart_open = pathlib.Path(fpath)
+        from smart_open.smart_open_lib import pathlib
 
-        with open(path_open, 'rb') as fin:
-            expected = fin.read().decode('cp852')
-        with smart_open.smart_open(path_smart_open, encoding='cp852') as fin:
+        fpath = pathlib.Path(os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt'))
+        with smart_open.smart_open(fpath, mode='r', encoding='cp852') as fin:
             actual = fin.read()
-        self.assertEqual(expected, actual)
```
mpenkov: We still want to check that the contents are equal. We just don't want to read the expected file via open/pathlib, because that won't work for all Python versions, as you've correctly pointed out.

So the correct thing to do here is (untested pseudo-code):

```python
expected = open(fpath).read().decode('cp852')
self.assertEqual(expected, actual)
```

You can use a with statement if you like, I guess that's more Pythonic.

clintval: I'll add that back in, I misunderstood your previous comment.

clintval: @mpenkov thanks for your patience. I amended the commit.

mpenkov:
@menshikh-iv LGTM, please merge when you're ready.

menshikh-iv:
Thanks @clintval :+1:, good work!

PR #177, opened 2018-03-18
Issue-176: Make author name ascii

baldwindc:
Unicode author names are causing `python setup.py install` to break. I didn't spend enough time trying to determine why, but this does resolve the issue. 

Tested on `Python 2.7.10`.


menshikh-iv:
@baldwindc thanks! This done in slightly different way in #178

baldwindc:
Heh no worries. That was my original implementation, until someone recommended the utf-8 encode :)

PR #180, opened 2018-03-20
Ordered mode for `s3_iter_results`

samkaufman:
For a local project, I needed to iterate over a bucket's contents _in order_ (they represent time series data), so I added a flag to switch between use of `imap` and `map` in `s3_iter_results`. I thought my changes might be more generally useful.

Because requiring ordered consumption makes it more likely that one or more producer queues will get backed up and cause a memory overflow, an additional `maxresults` option is added to limit the number of outstanding fetch tasks.


smart_open/smart_open_lib.py
```diff
@@ -902,7 +903,7 @@ def s3_iter_bucket_process_key(key, retries=3):
             pass
 
 
-def s3_iter_bucket(bucket, prefix='', accept_key=lambda key: True, key_limit=None, workers=16, retries=3):
+def s3_iter_bucket(bucket, prefix='', accept_key=lambda key: True, key_limit=None, workers=16, retries=3, ordered=False, maxresults=None):
```
menshikh-iv: please add this arguments to docstrings with description, what is it

mpenkov: I think that maxresults is a poor solution to the problem of the producer queues filling up, because it:

1. Requires the user to know about the implementation details
2. Requires the user to specify this parameter in order to avoid running out of memory while iterating (something that should never really happen, because that is a key benefit of iteration).
3. Terminates the iteration early, without yielding all possible results

I think a better solution would be to process the keys in batches. 

- Make a constant `BATCH_SIZE = 1000` and read up to BATCH_SIZE keys from the iterable at a time.
- Process the batch using imap
- Yield the keys one by one
- Move on with the next batch

I think this approach addresses the three problems I described above.

mpenkov: Also, please explain how maxresults is different from key_limit. They seem very strongly related (if not the same) and it seems redundant keeping them both.

menshikh-iv:
ping @samkaufman, when you plan to finish PR?

samkaufman:
@menshikh-iv I can't commit to a timeline at the moment, but I'll get to this as soon as I can. And thank you, @mpenkov, for the feedback.

menshikh-iv:
ping @samkaufman, are you planning to finish PR?

samkaufman:
@menshikh-iv Yes. Thanks for the ping. I'll include documentation and tests.

@mpenkov Replying to previous feedback:

- `key_limit` controls the total number of keys yielded before `s3_iter_bucket` returns, whereas `maxresults` controls the total number of keys in memory at any given time.
- I don't believe `maxresults` does terminate early, whereas `key_limit` will. If it does, that's a mistake on my part.
- Using a `BATCH_SIZE` constant won't suffice for workloads where, say, 1000 items is too large to fit in memory. The amount of memory consumed is not dependent on just the number of items, but also the size of the items. For example, I am using `smart_open` with a deep learning workload, where each item can be several hundred megabytes. I don't think this is a case of the user needing to understand implementation details of the library, and I don't know that there's a good way around requiring the user to understand something about memory in this case.
- The proposal to yield in batches as opposed to limiting the number of produced keys has the downside of non-overlapping I/O at the end of a batch. For instance, at batch item 999 of 1000, only one request can be outstanding. My proposed PR allows, for example, items 999, 1000, and 1001 to all be in-flight even if `maxresults` is set to 1000.

menshikh-iv:
@samkaufman great :+1: don't forget to resolve merge conflict first, please ping me when you are done with PR.

mpenkov:
@samkaufman Please add tests for your new functionality.

samkaufman:
@mpenkov Yep. I haven't forgotten. I'll ping @menshikh-iv when done.

samkaufman:
Quick update: Been beating my head against Python [issue #33997](https://bugs.python.org/issue33997). Not sure what a good solution is yet.

mpenkov:
How does that issue affect smart_open, e.g. does it hang our tests or do something else nasty?

samkaufman:
Exactly. `multiprocessing.Pool.terminate` hangs during tests. I don't think I'm doing anything that shouldn't be allowed by `multiprocessing`, but this is somehow a side effect of my changes, so I expect there's a workaround. I'll have to revisit when I have a second (and some rest).

mpenkov:
Hi @samkaufman, sorry for keeping this PR idle so long.

After thinking about it some more, I think the feature that this PR adds would belong more to the level above the smart_open library (application level) as opposed to the library itself. I think so for two reasons:

1. This new feature satisfies a narrow and very specific use case.
2. This feature could be implemented at the application level, above smart_open. The task can be generalized as: sort the output of a generator, keeping at most `max_in_memory` items in memory. In this specific case, the generator is the s3_iter_bucket function from smart_open, but it could be any generator in theory.

For the two reasons above, I'm closing this PR. Sorry for the long delay in getting back to you. I hope this doesn't discourage you from future contributions.

samkaufman:
Hi @mpenkov,

Thank you for getting back to me—I appreciate it—and no trouble about the delay.

This cannot be implemented as you described because memory will still be allocated and held internally by smart_open, no matter what is done with the results of the generator. Put another way, smart_open will continue to make requests and hold their results, even if the generator is never consumed by the application.

PR #182, opened 2018-04-03
Fix bug with changing `f._current_pos` when call `f.readline()`

inksink:
Use `_read_from_buffer` to set self._current_pos

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -232,12 +232,10 @@ class BufferedInputBase(io.BufferedIOBase):
             #
             if self._line_terminator in self._buffer:
                 next_newline = self._buffer.index(self._line_terminator)
-                the_line.write(self._buffer[:next_newline + 1])
-                self._buffer = self._buffer[next_newline + 1:]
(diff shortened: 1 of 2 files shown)


smart_open/s3.py
```diff
@@ -232,12 +232,10 @@ def readline(self, limit=-1):
             #
             if self._line_terminator in self._buffer:
                 next_newline = self._buffer.index(self._line_terminator)
```
piskvorky: Unrelated to this PR, but this seems inefficient: why search for `_line_terminator` twice? Here and in the line above? This is a linear, potentially costly operation, so we should only do it once.

mpenkov: You're right about it being costly, but when I wrote this code, I measured the alternative (try-catch), and that was slower (see the comment above this function). This behavior depends on the data you're processing, and what proportion of reads end up having the line terminator inside the buffer.

inksink:
OK. I will add tests for it.

inksink:
I have changed `test_readline()` to test `tell()` after `readline()`

PR #188, opened 2018-04-10
Сlose the old body explicitly after `seek` for S3. Fix #187

### LINKED ISSUE #187: S3 old object body was not closed after seek()
In class SeekableRawReader:
If we seek(), self._body will be replaced with a new object['Body']:
```
self._body = self._object.get(Range=range_string)['Body']
```
But old body should been closed. Why not : 
```
self._body.close()
self._body = self._object.get(Range=range_string)['Body']
```

inksink:
Solve issue #187

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -114,6 +114,16 @@ class SeekableRawReader(object):
         self._position = position
         range_string = _range_string(self._position)
         logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
+
+        #
+        # Close old body explicitly.
+        # When first seek(), self._body is not exist. Catch the exception and do nothing.
+        #
+        try:
+            self._body.close()
+        except AttributeError as e:
+            pass
+
         if position == self._content_length == 0 or position == self._content_length:
             #
             # When reading, we can't seek to the first byte of an empty file.


smart_open/s3.py
```diff
@@ -114,6 +114,16 @@ def seek(self, position):
...
         range_string = _range_string(self._position)
         logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
+
+        #
+        # Close old body explicitly.
+        # When first seek(), self._body is not exist. Catch the exception and do nothing.
+        #
+        try:
+            self._body.close()
+        except Exception as e:
```
mpenkov: Why not be explicit and catch AttributeError?

menshikh-iv: I'm +1 too for catching more "restricted", after this fix I think we can merge.

piskvorky: Is really only `AttributeError`? Can `close()` raise something else?

mpenkov: @piskvorky I don't think we expect close itself to raise anything.

I think the intention here is to mimic the behavior of:

```python
if body is not None:
    body.close()
```

without paying the price of the conditional every time.  I think the author is assuming that most the time, the body will be non-null. I think this assumption is correct.

PR #191, opened 2018-04-22
Replace `ParsedUri` class with functions, cleanup internal argument parsing

mpenkov:
This is better than what we previously had for two reasons.

First, we now know what options we parse from a URI *up front*. Previously, this was only possible by reading the relatively long ParsedUri constructor. Now, we just look at the fields of the namedtuple. This makes the code easier to understand, debug and improve. For example, it now becomes obvious that we don't actually use the ordinary_calling_format option anywhere: we just parse it.

Second, we make the code easier to understand by replacing a class with a data structure (namedtuple) and a parsing function. The previous ParsedUri class was really a [function masquerading as a function](https://youtu.be/jD0VJk9sOeY?t=21m15s) - this is considered un-Pythonic. If a class is being used as a function, it should really just be a function.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -95,7 +95,14 @@ Uri = collections.namedtuple(
 """Represents all the options that we parse from user input.
 
 Some of the above options only make sense for certain protocols, e.g.
-bucket_id is only for S3."""
+bucket_id is only for S3.
+"""
+#
+# Set the default values for all Uri fields to be None.  This allows us to only
+# specify the relevant fields when constructing a Uri.
+#
+# https://stackoverflow.com/questions/11351032/namedtuple-and-default-values-for-optional-keyword-arguments
+#
 Uri.__new__.__defaults__ = (None,) * len(Uri._fields)
 
 
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -22,6 +22,7 @@ The main methods are:
 """
 
 import codecs
+import collections
 import logging
 import os
 import os.path as P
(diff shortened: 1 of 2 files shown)


smart_open/smart_open_lib.py
```diff
@@ -372,82 +394,105 @@ class ParseUri(object):
...
+    elif parsed_uri.scheme.startswith('http'):
+        return Uri(scheme=parsed_uri.scheme, uri_path=uri_as_string)
+    else:
+        raise NotImplementedError(
+            "unknown URI scheme %r in %r" % (parsed_uri.scheme, uri_as_string)
+        )
+
+
+def _parse_uri_hdfs(parsed_uri):
+    assert parsed_uri.scheme == 'hdfs'
```
menshikh-iv: Why asserts needed (you check exact same conditions before in `if`/`elif`), here and everywhere? Sanity-check only?

smart_open/smart_open_lib.py
```diff
@@ -77,6 +78,27 @@
...
+        'ordinary_calling_format',
+        'access_id',
+        'access_secret',
+    )
+)
+"""Represents all the options that we parse from user input.
+
+Some of the above options only make sense for certain protocols, e.g.
+bucket_id is only for S3."""
+Uri.__new__.__defaults__ = (None,) * len(Uri._fields)
```
menshikh-iv: definitely need a comment here, too hacky

smart_open/smart_open_lib.py
mpenkov: Yes, it's a sanity check only.

PR #193, opened 2018-05-05
Handle edge case (read 0 bytes) in read function. Fix #171

### LINKED ISSUE #171: SeekableBufferedInputBase.read(0) returns data
Steps to reproduce:

```python
from smart_open.s3 import SeekableBufferedInputBase
input = SeekableBufferedInputBase(...)
assert input.read(0) == b''
```

I quickly investigated the problem. It seems the bug is in the `read()` method in the parent `BufferedInputBase` class:

```python
    def read(self, size=-1):
        """Read up to size bytes from the object and return them."""
        if size <= 0:
            if len(self._buffer):
                from_buf = self._read_from_buffer(len(self._buffer))
            else:
                from_buf = b''
            self._current_pos = self._content_length
            return from_buf + self._raw_reader.read()

       ...
```

`if size < 0` maybe?

mpenkov:


menshikh-iv:
Thanks @mpenkov :+1:

Issue #134, opened 2017-10-04
Streaming zipfile content


jcaisip:
I need to read or stream a zipfile content from the s3 bucket. My usual workflow was download the zipfile to my ec2 instance and then run a python script with the zipfile module  to read its content without extracting the zipfile. However, I would like to start streaming directly from s3 to save disk space. Do you think `smart_open` library can help me with my use case? Thanks.

piskvorky:
Zip is an archive format (one `.zip` can contain multiple files), so I'm not sure what "streaming" means in this context.

`smart_open` supports streaming of compressed files from S3, yes -- such as `.gz` or `.bz2`.

jcaisip:
I was trying to stream archive file contents in my use case. But since it  isn't supported yet, then I guess I really have to unzip the archive so I can file the file contents. Thanks @piskvorky !

PR #205, opened 2018-06-29
Adding support for streaming compressed archive files such as .zip, .tar.gz, .tar.bz2

handsomezebra:
This PR includes the following changes:
1. Compression wrapper for archive files such as .zip, .tar.gz, .tgz, and .tar.bz2. Also adding optional argument `member` to specify which member within the archive to open. Note that tar.gz/bz2 files are not really seekable so the reading of a member need to go through all the members before it.
2. A seekable http stream implementation for reading those archive from http. 
3. Related unit test and integration test changes.

I have done the unit test and integration test for python2 and python3 environment.

Please help to review. Thanks!

### DIFF
--- a/integration-tests/test_s3.py
+++ b/integration-tests/test_s3.py
@@ -89,7 +89,7 @@ def test_s3_encrypted_file(benchmark):
 
     key = _S3_URL + '/sanity.txt'
     text = 'с гранатою в кармане, с чекою в руке'
-    actual = benchmark(write_read, key, text, 'w', 'r', 'utf-8', multipart_upload={
+    actual = benchmark(write_read, key, text, 'w', 'r', 'utf-8', s3_upload={
         'ServerSideEncryption': 'AES256'
     })
     assert actual == text


handsomezebra:
Pinging...
Any comments about this PR?

piskvorky:
@handsomezebra what are the semantics of this feature? What does it mean to "stream an archive of files"?

Can you describe your use-case – what motivated this feature, how do you (or your colleagues) use it?

handsomezebra:
Sorry I don't understand your question. The whole purpose of smart_open is "Utils for streaming large files (S3, HDFS, gzip, bz2...)", right? This pack just add support for tar.gz, tar.bz2 and zip.

handsomezebra:
Ok, found an old issue discussing about this: #134. The semantic is very straightforward by looking at the code:

```smart_open.smart_open('myfile.zip', member='mydir/myfile.txt')```

So basically streaming from a member of the archive file.

**Comment by piskvorky** [comment]
Yes, the purpose of `smart_open` is to stream *files* – not directories / collections of files.

Can you describe your use case? What's your motivation for adding/using this feature?

**Comment by piskvorky** [comment]
I see what you mean. `smart_open` has no concept of an "inner file", so in this sense, the current PR  requires some thought, because it essentially proposes a new top API concept.

I can see how streaming a single file from an archive like ZIP that supports random access (not TAR though) could be useful. @mpenkov @menshikh-iv WDYT? What should the API look like?

I'm thinking: does this concept of sub-paths within input files apply to other situations, too? Such as referring to a subpart of a document: a column in a CSV, a specific section in a Markdown file? It's essentially a second identifier, specified through a new parameter (named `member` here).

Also, it raises the question of third, fourth, … identifiers (nested sub-paths). How do you refer to a file inside a ZIP file inside a ZIP file?

**Comment by mpenkov** [comment]
To me it sounds like this zip functionality doesn't _really_ need to be _inside_ the smart_open API. I'm not completely familiar with the use case @handsomezebra described, but I think you can satisfy it using something like:

```python
with smart_open.smart_open('s3://bucket/file.zip', 'rb') as fin:
    with zipfile.ZipFile(fin):
        # extract whatever member you want here
```

This doesn't require modifying the smart_open API and maintaining new functionality. My suggestion is to deal with this use case through documentation: add a section listing recipes for using smart_open to achieve tasks like this.

My thoughts are the same on applying the concept of sub-paths to other situations. smart_open does a good job of streaming files to/from remote locations. What you do to those files (split a column from CSV, go to a specific section of markdown file) is application-specific - smart_open shouldn't have to know anything about it.

In summary, I think that handling tar/zip members should happen _outside_ the smart_open library.

handsomezebra:
This doesn't work if the underlying stream is not seekable. If you look at my code you can see I added a seekable HTTP stream.

Conceptually this is not a change to the API, since it still streams a single file. The only difference is it is within an archive file instead of a local folder, http path, s3 path. If the API change is really an issue, we can define some kind of schema to specify the single file within the archive, for example "s3://bucket/file.zip#my_folder/my_file.txt".

mpenkov:
Seekable streams are great. We should definitely merge _that_ particular functionality. Once we do that, the sample I provided will satisfy your use case, right?

menshikh-iv:
I agree with @mpenkov 
- seekable HTTP streams **must** be merged (nice addition)
- if we have `seek`, @mpenkov example should work correctly (no API changes needed) -> any file from archive can be readed in streamming manner (or I missed something @handsomezebra?)

handsomezebra:
Thanks for accepting the seekable HTTP function.

But I don't want to skip the discussion about streaming compressed archive files. As I said a file name pattern "s3://bucket/file.zip#my_folder/my_file.txt" is an alternative way to specify a single file within an archive file. If we go by this there is no any API change. Any concerns about supporting that?

If we could support:
```python
for line in smart_open.smart_open('s3://bucket/file.zip#my_folder/my_file.txt', encoding='utf-8'):
    # do something for each line
```

Compare it to the following:
```python
with smart_open.smart_open('s3://bucket/file.zip', 'rb') as fin:
    with zipfile.ZipFile(fin, 'rb') as fin_zip:
        with fin_zip.open("my_folder/my_text.txt", 'rb') as fin_inner:
            for line in io.TextIOWrapper(fin_inner, 'utf-8'):
                # do something for each line
```
which one is easier for the users?

mpenkov:
> As I said a file name pattern "s3://bucket/file.zip#my_folder/my_file.txt" is an alternative way to specify a single file within an archive file. If we go by this there is no any API change. Any concerns about supporting that?

I'm not opposing this change because the API is sacred and set in stone (BTW, it isn't). I'm opposing this change because I think this functionality **does not belong in the library**. So, whether the proposed change requires adding a new keyword argument or not does not influence my stance.

Your sample code:

> for line in smart_open.smart_open('s3://bucket/file.zip#my_folder/my_file.txt', 'rb', encoding='utf-8'):
    # do something for each line

How would you handle the different [keyword arguments](https://docs.python.org/3/library/zipfile.html#zipfile-objects) that the zipfile library supports? If you don't couple the libraries, it's simple:

```python
with smart_open.smart_open('s3://bucket/file.zip', 'rb') as fin:
    with zipfile.ZipFile(fin, 'rb', compression=zipfile.ZIP_LZMA, allowZip64=False, compresslevel=9) as fin_zip:
       # do work here
```

If you insist on coupling smart_open to zipfile, then in order to support it fully, you need to pass the above arguments through smart_open. I think it's obvious that none of those keyword arguments affect the functionality of smart_open itself. We already have a ton on keyword arguments to deal with - I don't think we need any more.

> which one is easier for the users?

I'd argue that they are both trivial to the user. Writing four lines of code to integrate a library to satisfy 
a use case isn't unusual. In any case, it isn't significantly harder than writing the one line of code you propose.

I guess my opposition to this is [philosophical](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well). smart_open handles streaming. The zipfile library handles zip files. The user is responsible for integrating the two to solve this particular use case.

menshikh-iv:
@handsomezebra can you push HTTP-seekable + testing code as another PR, please? This part is a great addition and we want to merge it.

handsomezebra:
Sure let me do it sometime this week.

Issue #207, opened 2018-07-05
Different behaviour with `open`


menshikh-iv:
**Problem:** `smart_open==1.6.0` doesn't work correctly with `numpy.frombuffer` function (`numpy < 1.13.0`, preferable to test with `numpy==1.11.3`, `python2.7`).

**Example:**
**env**:
- `numpy==1.11.3`
- `gensim==3.4.0`
- `smart_open==1.6.0`
```python
from gensim.models import FastText
from gensim.test.utils import datapath

m = FastText.load_fasttext_format(datapath("lee_fasttext"))
```
Trace
```python
IOError                                   Traceback (most recent call last)
<ipython-input-1-afdc56d6c4b5> in <module>()
      2 from gensim.test.utils import datapath
      3 
----> 4 m = FastText.load_fasttext_format(datapath("lee_fasttext"))

/home/ivan/release/gensim/gensim/models/fasttext.py in load_fasttext_format(cls, model_file, encoding)
    697             model_file += '.bin'
    698         model.file_name = model_file
--> 699         model.load_binary_data(encoding=encoding)
    700         return model
    701 

/home/ivan/release/gensim/gensim/models/fasttext.py in load_binary_data(self, encoding)
    712             self._load_model_params(f)
    713             self._load_dict(f, encoding=encoding)
--> 714             self._load_vectors(f)
    715 
    716     def _load_model_params(self, file_handle):

/home/ivan/release/gensim/gensim/models/fasttext.py in _load_vectors(self, file_handle)
    819 
    820         self.num_original_vectors = num_vectors
--> 821         self.wv.vectors_ngrams = np.fromfile(file_handle, dtype=dtype, count=num_vectors * dim)
    822         self.wv.vectors_ngrams = self.wv.vectors_ngrams.reshape((num_vectors, dim))
    823         assert self.wv.vectors_ngrams.shape == (

IOError: first argument must be an open file
```

If I replace 
```python
with utils.smart_open(self.file_name, 'rb') as f:
```
from https://github.com/RaRe-Technologies/gensim/blob/875b028ccd009fe2fa7665e177b8ab0b5e2dc40d/gensim/models/fasttext.py#L711
to
```python
with open(self.file_name, 'rb') as f:
```
this will works correctly (also, it works correctly with `smart_open==1.5.6`, i.e. bug from `1.6.0`).

piskvorky:
[...]

PR #208, opened 2018-07-07
Use built-in open whenever possible

mpenkov:
Resolve issue #207


integration-tests/test_207.py
```diff
@@ -0,0 +1,28 @@
+import os
+import sys
+
+try:
+    import numpy as np
```
menshikh-iv: I think better to add numpy to test dependencies https://github.com/RaRe-Technologies/smart_open/blob/master/setup.py#L19

integration-tests/test_207.py
```diff
@@ -0,0 +1,28 @@
...
+    sys.exit(1)
+
+import smart_open
+
+
+def tofile():
+    dt = np.dtype([('time', [('min', int), ('sec', int)]), ('temp', float)])
+    x = np.zeros((1,), dtype=dt)
+
+    fname = "test.dat"
```
menshikh-iv: better to use really tempfiles, instead of hardcoded names

integration-tests/test_207.py
```diff
@@ -0,0 +1,28 @@
+import os
```
menshikh-iv: can you run this file in CI too (and other integration tests that missed) please?

smart_open/smart_open_lib.py
```diff
@@ -284,6 +284,16 @@ def _shortcut_open(uri, mode, **kw):
         open_kwargs['encoding'] = encoding
         mode = mode.replace('b', '')
 
+    #
+    # Under Py3, the built-in open accepts kwargs, and it's OK to use that.
+    # Under Py2, the built-in open _doesn't_ accept kwargs, but we still use it
```
menshikh-iv: what's about `buffering` https://docs.python.org/2/library/functions.html#open ?

integration-tests/test_207.py
menshikh-iv: Or even add it like standart unittest (not integration), wdyt?

piskvorky:
@mpenkov please use descriptive titles & put fixes #XYZ only in the description. It makes for easier issue/PR scanning, and also allows Github to hyperlink the pages. Thanks.

integration-tests/test_207.py
mpenkov: TBH, I don't think numpy belongs in our dependencies, because we only need it for one thing: to run this test.

What else would be the benefit of including it?

integration-tests/test_207.py
mpenkov: I added this tests. There are no other tests that we can run automatically right now - the remaining tests require user input. They'll need a bit of work before they can be fully automatic.

mpenkov:
@menshikh-iv I've responded to your comments. Please have a look.

integration-tests/test_207.py
menshikh-iv: > because we only need it for one thing: to run this test.

yes, and I'm talking about test dependencies (not requirements of `smart_open` package)

> What else would be the benefit of including it?

No, only for correct run test, that's all

integration-tests/test_207.py
menshikh-iv: Aha, ok, we can do it later (out of scope of current PR)

menshikh-iv:
@mpenkov done

piskvorky:
LGTM.

@menshikh-iv @mpenkov How about another release? This PR is a welcome, important change.

Btw, `smart_open` just reached 1,000 starts on Github 🌟 Thanks a lot @mpenkov for your continued support, it's really great!

mpenkov:
@piskvorky Thank you!!

@menshikh-iv Is there anything left to do for this? If not, let's merge.

menshikh-iv:
@mpenkov LGTM for me too, thank you!
@piskvorky almost nothing to release IMO, I would like to do this later (after several more PRs will be merged).

PR #211, opened 2018-08-04
Update README

piskvorky:
Reorganize README + update the README examples to be py3k-compatible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,9 +1,8 @@
-=============================================
-smart_open -- utils for streaming large files
-=============================================
+======================================================
+smart_open — utils for streaming large files in Python
+======================================================
 
-|License|_ |Travis|_ 
-
+|License|_ |Travis|_
 
 .. |License| image:: https://img.shields.io/pypi/l/smart_open.svg
 .. |Travis| image:: https://travis-ci.org/RaRe-Technologies/smart_open.svg?branch=master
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -13,146 +13,143 @@ smart_open -- utils for streaming large files
...
+  >>> for line in smart_open('s3u://user:secret@host:port@mybucket/mykey.txt'):
   ...    print line
 
-  >>> with smart_open.smart_open('/home/radim/foo.txt.bz2', 'wb') as fout:
-  ...    fout.write("some content\n")
+  >>> # you can also use a boto.s3.key.Key instance directly:
+  >>> key = boto.connect_s3().get_bucket("my_bucket").get_key("my_key")
+  >>> with smart_open(key) as fin:
+  ...     for line in fin:
+  ...         print line
```
mpenkov: Should we use `print(line)` instead of `print line`? The former works under both Py2 and Py3, the latter doesn't.

piskvorky: Yes! Let me fix that.

README.rst
```diff
@@ -13,146 +13,143 @@ smart_open -- utils for streaming large files
...
 
-.. code-block:: python
+Why?
+----
 
-  >>> # get all JSON files under "mybucket/foo/"
-  >>> bucket = boto.connect_s3().get_bucket('mybucket')
-  >>> for key, content in s3_iter_bucket(bucket, prefix='foo/', accept_key=lambda key: key.endswith('.json')):
-  ...     print key, len(content)
+Working with large S3 files using Amazon's default Python library, `boto <http://docs.pythonboto.org/en/latest/>`_, is a pain. Its ``key.set_contents_from_string()`` and ``key.get_contents_as_string()`` methods only work for small files (loaded in RAM, no streaming).
```
mpenkov: We're using [boto3](https://boto3.readthedocs.io/en/latest/) these days, not boto.

boto3 is a bit easier to use than boto, and the API is completely different, but the sentiment of the above statement is still true.

piskvorky:
@mpenkov @menshikh-iv we also have a project logos (both positive / negative background), but I'm not sure where to put it:

smart_open_logo_positive_short_tb

smart_open_logo_positive_complete_tb

README.rst
```diff
@@ -13,150 +12,145 @@ smart_open -- utils for streaming large files
...
-There are nasty hidden gotchas when using ``boto``'s multipart upload functionality, and a lot of boilerplate.
-
-``smart_open`` shields you from that. It builds on boto but offers a cleaner API. The result is less code for you to write and fewer bugs to make.
-
-Installation
-------------
-::
-
-    pip install smart_open
+  >>> smart_open('s3://', s3_upload={ 'ServerSideEncryption': 'AES256' })
```
menshikh-iv: PEP8 (whitespaces before/after `{` and `}`)

README.rst
```diff
@@ -13,150 +12,145 @@ smart_open -- utils for streaming large files
...
 
-.. code-block:: python
+Why?
+----
 
-  >>> # get all JSON files under "mybucket/foo/"
-  >>> bucket = boto.connect_s3().get_bucket('mybucket')
-  >>> for key, content in s3_iter_bucket(bucket, prefix='foo/', accept_key=lambda key: key.endswith('.json')):
-  ...     print key, len(content)
+Working with large S3 files using Amazon's default Python library, `boto <http://docs.pythonboto.org/en/latest/>`_ and `boto3 <https://boto3.readthedocs.io/en/latest/>`_, is a pain. Its ``key.set_contents_from_string()`` and ``key.get_contents_as_string()`` methods only work for small files (loaded in RAM, no streaming).
```
menshikh-iv: Need to check, how this rendered on PyPI (as I remember, PyPI have really "strict" parser and readme render can be easily broken)

menshikh-iv:
@piskvorky about logo - maybe instead of the header (but I'm not sure that this is a good idea, by PyPI uploading reason)

README.rst
piskvorky: I don't think I made any changes here (just moved the sections around).

menshikh-iv: Checked, rendering works https://test.pypi.org/project/smart-open/

piskvorky: Great. I didn't know about test.pypi.org, nice find!

PR #212, opened 2018-08-05
Make close method idempotent + add dummy flush method for `S3`

mpenkov:
According to the [python docs](https://docs.python.org/3/library/io.html#io.IOBase.close), the close method should have no effect if the file is already closed.

Also, we need to expose a flush method in order to quack like a built-in file object.

There are many commits in this branch, most of them are just back-and-forth between me and the reporting user. It took us a long time to understand what the problem actually was.


smart_open/tests/test_s3.py
```diff
@@ -333,6 +333,24 @@ def test_nonexisting_bucket(self):
...
+    def test_double_close(self):
+        s3 = boto3.resource('s3')
+        s3.create_bucket(Bucket='test_double_close')
+        text = u'там за туманами, вечными, пьяными'.encode('utf-8')
+        fout = smart_open.s3.open('test_double_close', 'key', 'wb')
+        fout.write(text)
+        fout.close()
+        fout.close()
+
+    def test_flush_close(self):
```
menshikh-iv: what's an idea of this test?

smart_open/s3.py
```diff
@@ -472,6 +473,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.close()
 
 
+def _insert_spaces(the_string):
```
menshikh-iv: what's a function `_insert_spaces` (here and below)? I see no usages.

integration-tests/s3_cp.py
```diff
@@ -0,0 +1,12 @@
+import io
+import sys
+
+import smart_open
+
+s3_url, local_path = sys.argv[1:]
+with smart_open.smart_open(s3_url, 'rb') as fin, open(local_path, 'wb') as fout:
```
menshikh-iv: description needed: I see copy s3 -> local disk, what you test here?

smart_open/tests/test_s3.py
mpenkov: test_flush_close checks if we have a flush method.

test_double_close checks that calling close twice doesn't cause a crash.

Both of these tests were written as part of TDD. They capture the problematic behavior of smart open before the corresponding changes to s3 were made.

integration-tests/s3_cp.py
mpenkov: This is an analog of the "aws s3 cp" command. I used it for debugging only, so I deleted it.

smart_open/s3.py
mpenkov: Used for debugging only. Deleted.

menshikh-iv:
Thank you @mpenkov :+1:

PR #225, opened 2018-09-01
Drop support for python `3.3 and `3.4` + workaround for testing

menshikh-iv:
**Done:**

- Drop support for `3.3` and `3.4`
- Remove workaround for boto3 installation (no more needed)
- Pin boto3/botocore versions (by moto issue reason, see https://github.com/spulec/moto/issues/1793)

CC: @mpenkov @piskvorky

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,8 +68,6 @@ setup(
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
         'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Topic :: System :: Distributed Computing',
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ def read(fname):
 
 tests_require = [
     'mock',
-    'moto==0.4.31',
+    'moto',
     'pathlib2',
     'responses',
 ]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -49,7 +49,10 @@ def read(fname):
         'boto >= 2.32',
         'bz2file',
         'requests',
-        'boto3'
+        # Temporary pin boto3 & botocore, because moto doesn't work with new version
+        # See https://github.com/spulec/moto/issues/1793
+        'boto3 < 1.8.0',
```
piskvorky: Isn't moto needed only for testing? `install_requires` affects all users… should we restrict them from using newer boto3/botocore, just because of our testing dependency?

menshikh-iv: Moto needed only for testing, that's true, I understand that this influences to all users, but this is "temporary" only (no big deal if we slightly limit boto version for some time).

As alternative solution, I can revert this boto* restriction and instead of it, install pinned versions manually (for travis only), this can be better solution, wdyt @piskvorky @mpenkov?

piskvorky: Up to you; of course not limiting users is preferable, but it's probably no big deal, like you say.

mpenkov:
If we can’t run our unit tests with the newest boto3, then can we be sure
things really work?

If the answer is no, then we should restrict our to boto3 versions that we
know work.

setup.py
mpenkov: If we can’t run our unit tests with the newest boto3, then can we be sure
things really work with that boto3 version?

If the answer is no, then we should restrict our users to boto3 versions that we know work.

petergaultney: This is breaking things downstream from you, 17 days later. 

Please revert this, or move to a development-only dependency, or else it will be necessary for us to stop using this library.

menshikh-iv: @petergaultney hi, can you please describe, why does it break something? What's a problem with it?

petergaultney: yeah, #227 is my exact issue!  thanks!

Issue #210, opened 2018-07-15
Support for 's3a://' URIs


macks22:
**`smart-open` version**: 1.6.0

Currently, attempting to use a URI scheme start with `'s3a://'` results in `NotImplementedError`:
```python
    def _parse_uri(uri_as_string):
        """
        Parse the given URI from a string.
    
        Supported URI schemes are "file", "s3", "s3n", "s3u" and "hdfs".
    
          * s3 and s3n are treated the same way.
          * s3u is s3 but without SSL.
    
        Valid URI examples::
    
          * s3://my_bucket/my_key
          * s3://my_key:my_secret@my_bucket/my_key
          * s3://my_key:my_secret@my_server:my_port@my_bucket/my_key
          * hdfs:///path/file
          * hdfs://path/file
          * webhdfs://host:port/path/file
          * ./local/path/file
          * ~/local/path/file
          * local/path/file
          * ./local/path/file.gz
          * file:///home/user/file
          * file:///home/user/file.bz2
        """
        if os.name == 'nt':
            # urlsplit doesn't work on Windows -- it parses the drive as the scheme...
            if '://' not in uri_as_string:
                # no protocol given => assume a local file
                uri_as_string = 'file://' + uri_as_string
        parsed_uri = urlsplit(uri_as_string, allow_fragments=False)
    
        if parsed_uri.scheme == "hdfs":
            return _parse_uri_hdfs(parsed_uri)
        elif parsed_uri.scheme == "webhdfs":
            return _parse_uri_webhdfs(parsed_uri)
        elif parsed_uri.scheme in ("s3", "s3n", "s3u"):
            return _parse_uri_s3x(parsed_uri)
        elif parsed_uri.scheme in ('file', '', None):
            return _parse_uri_file(parsed_uri)
        elif parsed_uri.scheme.startswith('http'):
            return Uri(scheme=parsed_uri.scheme, uri_path=uri_as_string)
        else:
            raise NotImplementedError(
>               "unknown URI scheme %r in %r" % (parsed_uri.scheme, uri_as_string)
            )
E           NotImplementedError: unknown URI scheme 's3a' in 's3a://reference_dirpath/reference_filename.csv'
```

As far as I can tell, there is not currently a workaround for this. While the API does support using `boto.s3.key.Key`, it does not support the `boto3` Object metadata dict returned from `boto3.client('s3').get_object(Bucket=bucket, Key=key)`; gives:
```python
[...]

PR #229, opened 2018-09-15
Added support for s3a URLs

mpenkov:
This resolves https://github.com/RaRe-Technologies/smart_open/issues/210


menshikh-iv:
Thanks for PR @mpenkov, is it really works? I'm confused a bit because you just allow other prefixes and don't change other code

PR #230, opened 2018-09-15
Clean up kwargs

mpenkov:
We used to liberally pass arbitrary keyword arguments (e.g. **kwargs). This is a bad practice because it's difficult to keep track of function parameters and ensure they're being passed down to the right function.

This PR starts to clean up our use of arbitrary keyword arguments.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -374,15 +374,28 @@ def _open_binary_stream(uri, mode, kwargs):
             return fobj, filename
         elif parsed_uri.scheme in ("s3", "s3n", 's3u'):
             endpoint_url = None if kwargs.host is None else 'http://' + kwargs.host
+
+            #
+            # TODO: this shouldn't be smart_open's responsibility.
+            #
+            if kwargs.s3_session is None:
+                import boto3
+                session = boto3.Session(
+                    profile_name=kwargs.profile_name,
+                    aws_access_key_id=parsed_uri.access_id,
+                    aws_secret_access_key=parsed_uri.access_secret,
+                )
+                resource = session.resource('s3', endpoint_url=endpoint_url)
+            else:
+                assert parsed_uri.access_id is None, 'aws_access_key_id conflicts with session'
+                assert parsed_uri.access_secret is None, 'aws_secret_access_key conflicts with session'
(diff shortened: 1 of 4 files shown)


smart_open/smart_open_lib.py
```diff
@@ -322,8 +322,17 @@ def _shortcut_open(uri, mode, kwargs):
...
     open_kwargs = {}
-    if kwargs.errors is not None:
+    if kwargs.errors is not None and 'b' not in mode:
+        #
+        # Ignore errors parameter if we're reading binary.  This logic is
+        # necessary because:
+        #
+        #   - open complains if you pass non-null errors in binary mode
+        #   - our default mode is rb (incompatible with default Python open)
+        #   - our default errors is non-null (to match default Python open)
```
piskvorky: What does non-null mean? Is null == None?

smart_open/smart_open_lib.py
```diff
@@ -322,8 +322,17 @@ def _shortcut_open(uri, mode, kwargs):
...
 
     open_kwargs = {}
-    if kwargs.errors is not None:
+    if kwargs.errors is not None and 'b' not in mode:
+        #
+        # Ignore errors parameter if we're reading binary.  This logic is
+        # necessary because:
+        #
+        #   - open complains if you pass non-null errors in binary mode
+        #   - our default mode is rb (incompatible with default Python open)
```
piskvorky: How about we change this to be compatible?

Or at least make the mode mandatory, so people have to make a conscious decision between binary/text.

smart_open/smart_open_lib.py
mpenkov: Yes.

smart_open/smart_open_lib.py
mpenkov: Yes, we will do this in the next major release.

mpenkov: BTW, we cannot make the mode mandatory, as that breaks compatibility with the built-in open.

We will switch to reading text mode as default in the next release.

menshikh-iv:
Hi @mpenkov, how is going here? Where you plan to finish this (internal changes + warnings)?

mpenkov:
I'll have a look on the weekend. Thanks for reminding me.

mpenkov:
I figured out a better way to do this.

PR #236, opened 2018-09-24
Adding streaming http support.

handsomezebra:
This pack adds the streaming support for http.
The responses lib doesn't support mock of streaming http so I didn't have a good way to unit test it. Any suggestions?


menshikh-iv:
Hi @handsomezebra, I think you can mock it manually with standart python [mock](https://docs.python.org/3/library/unittest.mock.html).

CC: @mpenkov

smart_open/http.py
```diff
@@ -75,32 +84,42 @@ def detach(self):
...
+    def read(self, size=-1):
         """
         Mimics the read call to a filehandle object.
         """
-        logger.debug('read: %r', locals())
-        if size is None:
-            return self.response.raw.read()
+        logger.debug("reading with size: %d", size)
+        if self.response is None:
+            return ''
```
menshikh-iv: shouldn't this be an `b''` instead of string (here and below)?

smart_open/http.py
```diff
@@ -114,3 +133,118 @@ def readinto(self, b):
             return 0
         b[:len(data)] = data
         return len(data)
+
+
+class SeekableBufferedInputBase(BufferedInputBase):
```
menshikh-iv: Maybe better use a more specific name (in case if this applied only for HTTP)?

mpenkov: IMHO, this is already part of the http module, so everything in there is specific to http a priori, so the name is fine. We use a similar naming scheme in the S3 module as well.

smart_open/http.py
```diff
@@ -75,32 +84,42 @@ def detach(self):
...
-        logger.debug('read: %r', locals())
-        if size is None:
-            return self.response.raw.read()
+        logger.debug("reading with size: %d", size)
+        if self.response is None:
+            return ''
+
+        if size == 0:
+            return ''
+        elif size < 0:
```
mpenkov: Unless there's a good reason to nest conditionals here, I think rewriting them as un-nested is better.

```python
elif size < 0 and not self._read_buffer:
    retval = self.response.raw.read()
elif size < 0:
    retval = self._read_buffer + self.response.raw.read()
    self._read_buffer = b''
```

smart_open/http.py
```diff
@@ -114,3 +133,118 @@ def readinto(self, b):
...
+        if whence == START:
+            new_pos = offset
+        elif whence == CURRENT:
+            new_pos = self._current_pos + offset
+        elif whence == 2:
+            new_pos = self.content_length + offset
+
+        new_pos = _clamp(new_pos, 0, self.content_length)
+
+        if self._current_pos != new_pos:
```
mpenkov: I'd use an early return here. Saves the reader from remembering they're inside a conditional, and saves the writer an extra level of indentation. Also dodges the need to nest conditionals.

```python
if self._current_pos == new_pos:
    return self._current_pos
```

menshikh-iv:
ping @handsomezebra, are you planning to finish PR?
CC @mpenkov

mpenkov:
Still need to add tests.

mpenkov:
@handsomezebra I cleaned up your commit for you, added some tests, and merged this in. Congrats on your first PR to smart_open :hooray:

PR #237, opened 2018-10-01
Clean up _parse_uri_s3x function, resolve edge cases

### LINKED ISSUE #195: Example with custom S3 isnt working
The examples list

```python
>>> # using a completely custom s3 server, like s3proxy:
>>> for line in smart_open.smart_open('s3u://user:secret@host:port@mybucket/mykey.txt'):
...    print line

```

But this doesn't work. `smart_open` does not parse the host/port bits.

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L383

mpenkov:
Fix #195

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,7 +47,7 @@ What?
   ...     print(line)
 
   >>> # stream from WebHDFS
-  >>> for line in smart_open('webhdfs://host:port/user/hadoop/my_file.txt'):
+  >>> for line in smart_open('webhdfs://host:1234/user/hadoop/my_file.txt'):
   ...     print(line)
 
   >>> # stream content *into* S3 (write mode):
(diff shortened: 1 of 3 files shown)


menshikh-iv:
Looks good @mpenkov! 

Can you also fix `README.md` please (maybe put concrete values or better highlight these pieces, like `<HOST>:<PORT>`, or something different?). Goal - avoid missunderstanding.

menshikh-iv:
@mpenkov :fire:

piskvorky:
-1 on the README changes. `1234` tells me nothing, `port` better. In fact, it's misleading because it's inconsistent with the other placeholders: `host`, `user`, `secret` etc -- why is one concrete and the other general? Should we replace all?

piskvorky:
Nah. In my experience, a person who thinks `host:port` should be taken literally won't really be reading or understanding such additional texts anyway. And they'll definitely copy-paste `1234` too, so that doesn't help.

PR #241, opened 2018-10-10
added support for opening a specific s3 object version 

mbertoldi:
Added version_id kwarg to open a specific s3 object version:

```python
with smart_open('s3://mybucket/mykey', mode='r', version_id='version') as s3_object:
    print(s3_object.read())
```

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -80,10 +80,13 @@ def open(bucket_id, key_id, mode, **kwargs):
 
 class RawReader(object):
     """Read an S3 object."""
-    def __init__(self, s3_object):
+    def __init__(self, s3_object, version_id=None):
         self.position = 0
         self._object = s3_object
-        self._body = s3_object.get()['Body']
+        if version_id:
+            self._body = s3_object.get(VersionId=version_id)['Body']
+        else:
+            self._body = s3_object.get()['Body']
 
     def read(self, size=-1):
         if size == -1:
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -94,9 +94,13 @@ class RawReader(object):
 class SeekableRawReader(object):
     """Read an S3 object."""
 
-    def __init__(self, s3_object):
+    def __init__(self, s3_object, version_id=None):
         self._object = s3_object
-        self._content_length = self._object.content_length
+        if version_id:
+            self._content_length = self._object.get(VersionId=version_id)['ContentLength']
+        else:
+            self._content_length = self._object.content_length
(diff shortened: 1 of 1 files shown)


smart_open/s3.py
```diff
@@ -80,10 +80,13 @@ def open(bucket_id, key_id, mode, **kwargs):
 
 class RawReader(object):
     """Read an S3 object."""
-    def __init__(self, s3_object):
+    def __init__(self, s3_object, version_id=None):
         self.position = 0
         self._object = s3_object
-        self._body = s3_object.get()['Body']
+        if version_id:
```
mpenkov: Do you really need to check if version_id is None? I you explicitly do `s3_object.get(VersionId=None)`, what is the result?

menshikh-iv:
related to #230

smart_open/s3.py
mbertoldi: Hi, boto validate the parameters using the keys in kwargs, so i i pass VersionId=None it will raise a ParamValidationError:
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter VersionId, value: None, type: <type 'NoneType'>, valid types: <type 'basestring'>

mpenkov: Interesting. It's a keyword argument, so it must have a default value. Can you figure out what that default value is?

mbertoldi: All these parameters are in the kwargs dict, so they don't have a default value.
Boto3 are created automatically using a json definition (see boto3 service.py and json definitions in botocore/data)
Before making a network request to AWS api all parameters are validated and serialized.

mpenkov: OK, I see. Can you please make a function:

```python
def _get(s3_object, version=None, **kwargs):
  if version is not None:
      kwargs['Version'] = version
  return s3_object.get(**kwargs)
```

and then use it in your code instead of the conditionals?

Right now, we've checking for "version is not None" in too many places, it's repetitive, error prone and asking for trouble.

mpenkov:
@mbertoldi Hi, just checking in. Do you think you could finish this PR?

mbertoldi:
> Il giorno 15 mar 2019, alle ore 16:22, Michael Penkov <notifications@github.com> ha scritto:
> 
> @mbertoldi <https://github.com/mbertoldi> Hi, just checking in. Do you think you could finish this PR?
> 

Hi, sure, now i'm quite busy, but as soon as possible I'll do the unit tests :)

Ciao,
Michele

mpenkov:
Closing in favor of #325

PR #245, opened 2018-11-05
Raise `ValueError` if s3 key does not exist

### LINKED ISSUE #174: Don't swallow ClientError when bucket doesn't exist
In `smart_open.s3.BufferedOutputBase.__init__()` (lines 339-342) why does this library swallow the botocore ClientError and return a ValueError? It makes it tricky to handle the error case properly.

adrpar:
As mentioned on ISSUE https://github.com/RaRe-Technologies/smart_open/issues/174 `botocore.client.ClientError` is raised instead of `ValueError` on S3 keys that do not exists.

This PR fixes this inconsistency.

Fix #174

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -96,7 +96,10 @@ class SeekableRawReader(object):
 
     def __init__(self, s3_object):
         self._object = s3_object
-        self._content_length = self._object.content_length
+        try:
+            self._content_length = self._object.content_length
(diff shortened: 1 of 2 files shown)


adrpar:
Thank you for the review. I added the test (thought I could get away with being lazy :) )

menshikh-iv:
Thank you @adrpar, congratz with the first contribution!

PR #246, opened 2018-11-08
Ensure _list_bucket uses continuation token for subsequent pages

tcsavage:
When trying to `s3_iter_bucket` a large set of objects (more than 1000), we found we were given the same set of 1000 over and over again, _forever_.

The issue was that the continuation token to fetch the next page, although it was being captured, was not being given to subsequent invocations of `list_objects_v2`. This pull request makes sure that if we have a continuation token, that we use it.

Unfortunately, I have been unable to run the test suite but this change should fix the test "[test_list_bucket_long](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/tests/test_s3.py#L394-L402)" which is currently being skipped because it was taking "too long". Probably because it would run forever.

### DIFF
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -391,7 +391,6 @@ class IterBucketTest(unittest.TestCase):
         expected = ['key_%d' % x for x in range(num_keys)]
         self.assertEqual(sorted(keys), sorted(expected))
 
-    @unittest.skip('this test takes too long for some unknown reason')
     def test_list_bucket_long(self):
         num_keys = 1010
         populate_bucket(num_keys=num_keys)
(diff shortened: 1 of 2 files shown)


tcsavage:
@mpenkov As I mentioned, I was unable to run the tests on my local machine, but I will remove the skip decorator and see if it passes in Travis.

smart_open/tests/test_s3.py
```diff
@@ -391,7 +391,6 @@ def test_list_bucket(self):
         expected = ['key_%d' % x for x in range(num_keys)]
         self.assertEqual(sorted(keys), sorted(expected))
 
-    @unittest.skip('this test takes too long for some unknown reason')
```
piskvorky: LOL! So we caught the error, but commented out the test instead of fixing :)

tcsavage:
The build failed, but it looks like a problem with `test_old` taking too long to run. `test_list_bucket_long` itself passed just fine. Is `test_old` known to be a bit flakey?

tcsavage:
@mpenkov @piskvorky The tests pass, could we merge and release before this gets stale?

menshikh-iv:
Thank you @tcsavage, congratz with the first contribution :1st_place_medal:

PR #256, opened 2019-01-17
Drop deprecated `sudo` from travis config

cclauss:
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).  Also, don’t hardcode __dist: trusty__ because it is currently the Travis default but that might change soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: python
-dist: trusty
-sudo: false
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Thank you @cclauss :+1:

PR #260, opened 2019-02-20
Add support for .xz / lzma

vmarkovtsev:
Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ What?
 
 ``smart_open`` is a Python 2 & Python 3 library for **efficient streaming of very large files** from/to S3, HDFS, WebHDFS, HTTP, or local (compressed) files. It's a drop-in replacement for Python's built-in ``open()``: it can do anything ``open`` can (100% compatible, falls back to native ``open`` wherever possible), plus lots of nifty extra stuff on top.
 
-``smart_open`` is well-tested, well-documented and sports a simple, Pythonic API:
+``smart_open`` is well-tested, well-documented, and has a simple, Pythonic API:
 
 .. code-block:: python
 
@@ -99,6 +99,11 @@ Or, if you prefer to install from the `source tar.gz <http://pypi.python.org/pyp
     python setup.py install
 
 To run the unit tests (optional), you'll also need to install `mock <https://pypi.python.org/pypi/mock>`_ , `moto <https://github.com/spulec/moto>`_ and `responses <https://github.com/getsentry/responses>`_ (``pip install mock moto responses``). The tests are also run automatically with `Travis CI <https://travis-ci.org/RaRe-Technologies/smart_open>`_ on every commit push & pull request.
+
+Supported archive types
+-----------------------
+``smart_open`` allows reading and writing gzip, bzip2 and xz files. They are transparently handled
+over HTTP, too.
 
 S3-Specific Options
 -------------------
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -59,12 +59,14 @@ else:
     from bz2 import BZ2File
 
 import gzip
-
-COMPRESSED_EXT = ('.gz', '.bz2')  # supported compressed file extensions
-
-if not IS_PY2:
+try:
     import lzma
-    COMPRESSED_EXT += '.xz',
+except ImportError:
+    # py<3.3
+    from backports import lzma
+
+
+COMPRESSED_EXT = ('.gz', '.bz2', '.xz')  # supported compressed file extensions
 
 #
 # This module defines a function called smart_open so we cannot use
@@ -586,7 +588,7 @@ def _compression_wrapper(file_obj, filename, mode):
         return BZ2File(file_obj, mode)
     elif ext == '.gz':
         return gzip.GzipFile(fileobj=file_obj, mode=mode)
-    elif ext == '.xz' and not IS_PY2:
+    elif ext == '.xz':
         return lzma.LZMAFile(filename=file_obj, mode=mode, format=lzma.FORMAT_XZ)
     else:
         return file_obj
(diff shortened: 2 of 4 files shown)


smart_open/smart_open_lib.py
```diff
@@ -577,6 +580,8 @@ def _compression_wrapper(file_obj, filename, mode):
         return BZ2File(file_obj, mode)
     elif ext == '.gz':
         return gzip.GzipFile(fileobj=file_obj, mode=mode)
+    elif ext == '.xz' and not IS_PY2:
```
piskvorky: Do we want smart_open to behave differently on the same file type, across py2 / py3? (decompress in py3, leave compressed in py2)

That could generate nasty surprises for users, which I don't like.

vmarkovtsev: We can try to import pyLZMA on PY2, but if we fail, we don't have any other option besides writing a RuntimeWarning. WDYT?

vmarkovtsev:
@piskvorky CI passed, README updated, py2.7 handled. PTAL

piskvorky:
Thanks for the swift resolution!

@mpenkov WDYT?

smart_open/tests/test_smart_open.py
```diff
@@ -259,7 +259,34 @@ def test_http_bz2(self):
...
 
-        # decompress the gzip and get the same md5 hash
+        # decompress the bzip2 and get the same md5 hash
+        self.assertEqual(smart_open_object.read(), test_string)
+
+    @responses.activate
+    def test_http_xz(self):
+        """Can open xz via http?"""
+        test_string = b'Hello World Compressed.'
+        #
```
mpenkov: Yes, I agree with the text in this TODO. We should avoid unnecessary file IO (create temporary file, write to file, read from file, delete file) in unit tests, especially when the alternative is cleaner and simpler (encode to a io.BytesIO directly).

smart_open/tests/test_smart_open.py
```diff
@@ -259,7 +259,34 @@ def test_http_bz2(self):
...
+        with smart_open.smart_open(test_file, 'wb') as outfile:
+            outfile.write(test_string)
+
+        with open(test_file, 'rb') as infile:
+            compressed_data = infile.read()
+
+        if os.path.isfile(test_file):
+            os.unlink(test_file)
+
+        responses.add(responses.GET, "http://127.0.0.1/data.xz",
```
mpenkov: No need to line break here. We have a line limit of 120 characters.

smart_open/tests/test_smart_open.py
```diff
@@ -882,17 +913,22 @@ def test_open_gz(self):
...
+        self._test_write_read('.gz')
+
     def test_write_read_bz2(self):
         """Can write and read bz2?"""
-        with tempfile.NamedTemporaryFile('wb', suffix='.bz2', delete=False) as infile:
-            test_file_name = infile.name
-        self.write_read_assertion(test_file_name)
+        self._test_write_read('.bz2')
+
+    def test_write_read_xz(self):
```
mpenkov: You also need a test that successfully reads a known .xz file. Currently, you test that your .xz reader and writer are compatible with each other, but you're not testing whether they are actually working with the .xz format (e.g. they could be using gzip by mistake and this test would still pass).

smart_open/tests/test_smart_open.py
```diff
@@ -882,17 +913,22 @@ def test_open_gz(self):
         assert m.hexdigest() == '18473e60f8c7c98d29d65bf805736a0d', \
             'Failed to read gzip'
 
-    def test_write_read_gz(self):
-        """Can write and read gzip?"""
-        with tempfile.NamedTemporaryFile('wb', suffix='.gz', delete=False) as infile:
+    def _test_write_read(self, suffix):
```
mpenkov: This function feels unnecessary. All it does is declare test_file_name and pass it into the write_read_assertion function.

Why not declare test_file_name in the write_read_assertion function directly? That would make the tests easier to read (need to step into only one extra function call instead of two).

README.rst
```diff
@@ -100,6 +100,11 @@ Or, if you prefer to install from the `source tar.gz <http://pypi.python.org/pyp
 
 To run the unit tests (optional), you'll also need to install `mock <https://pypi.python.org/pypi/mock>`_ , `moto <https://github.com/spulec/moto>`_ and `responses <https://github.com/getsentry/responses>`_ (``pip install mock moto responses``). The tests are also run automatically with `Travis CI <https://travis-ci.org/RaRe-Technologies/smart_open>`_ on every commit push & pull request.
 
+Supported archive types
+-----------------------
+``smart_open`` allows reading and writing gzip, bzip2 and xz files. They are transparently handled
+over HTTP, too.
```
mpenkov:
```suggestion
over HTTP, S3, and other protocols, too.
```

smart_open/tests/test_smart_open.py
vmarkovtsev: This has been copied from `test_http_bz`. I understand that you wish to refactor the tests, but the goal of this PR is different. I will submit a refactoring PR separately, ok?

smart_open/tests/test_smart_open.py
vmarkovtsev: Again, copy-pasted from `test_http_bz`. Shall I edit both functions at once or submit a separate PR?

smart_open/tests/test_smart_open.py
vmarkovtsev: We don't test it for bzip2 as well (just noticing). OK, I will add this test.

smart_open/tests/test_smart_open.py
vmarkovtsev: It used to be much worse before - two archive test function copy-pasted. NP, I will change it.

PR #261, opened 2019-02-20
Remove unused SSLError import

vmarkovtsev:
Checked by `grep` - indeed, not used.


mpenkov:
OK, looks good. Thank you.

PR #263, opened 2019-02-23
Handle "+", fix file objects and refactor SmartOpenHttpTest

vmarkovtsev:
Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>

It all started with the refactoring suggestion in https://github.com/RaRe-Technologies/smart_open/pull/260#discussion_r258734796 and went a little bit too far:

- I found out that the file objects were handled loosely. We used to check only for the existence of `read()` disregarding the actual open mode. I added `_is_stream()` function and fixed that.
- Added `name` attribute check to the file objects handler. Thus it becomes possible to open a regular stream, specify `name` and read/write compressed data. This was a prerequisite for the "file-less" `SmartOpenHttpTest`-s. 

Then I started writing the tests and found a few minor bugs:

- rw modes (r+, w+, a+) were not checked in `_compression_wrapper()`. All three compression codecs do not support reading and writing at the same time.
- rw modes were not correctly handled in `_encoding_wrapper()`: only the reader's encoder was applied. I changed that to chaining reader's and writer's encoders in rw modes.

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -390,7 +390,7 @@ def _open_binary_stream(uri, mode, **kw):
         filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
-        raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
+        raise TypeError("don't know how to handle uri %r" % uri)
 
 
 def _s3_open_uri(parsed_uri, mode, **kwargs):
@@ -587,7 +587,7 @@ def _compression_wrapper(file_obj, filename, mode):
         warnings.warn('streaming gzip support unavailable, see %s' % _ISSUE_189_URL)
         file_obj = io.BytesIO(file_obj.read())
     if ext in COMPRESSED_EXT and mode.endswith('+'):
-        raise ValueError('impossible to open in read+write+compressed mode')
+        raise ValueError('transparent (de)compression unsupported for mode %r' % mode)
 
     if ext == '.bz2':
         return BZ2File(file_obj, mode)
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -381,36 +381,16 @@ def _open_binary_stream(uri, mode, **kw):
         if host is not None:
             kw['endpoint_url'] = _add_scheme_to_host(host)
         return smart_open_s3.open(uri.bucket.name, uri.name, mode, **kw), uri.name
-    elif _is_stream(uri, mode):
+    elif hasattr(uri, 'read'):
         # simply pass-through if already a file-like
         # we need to return something as the file name, but we don't know what
-        # so we probe for uri.name (e.g., this works with tempfile.NamedTemporaryFile)
+        # so we probe for uri.name (e.g., this works with open() or tempfile.NamedTemporaryFile)
         # if the value ends with COMPRESSED_EXT, we will note it in _compression_wrapper()
-        # if there is no such an attribute, we return "unknown" - this effectively disables
-        # any compression (the actual string does not matter ofc)
+        # if there is no such an attribute, we return "unknown" - this effectively disables any compression
         filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
         raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
-
-
-def _is_stream(fileobj, mode):
-    """
-    Detect whether the specified object is a file object with required capabilities
-    implied by `mode`.
-    """
-    has_read = hasattr(fileobj, 'read')
-    has_write = hasattr(fileobj, 'write')
-    if not has_read and not has_write:
-        return False
-    if mode.endswith('+'):
-        return has_read and has_write
-    if mode[0] == 'r':
-        return has_read
-    if mode[0] in ('w', 'a'):
-        return has_write
-    # we should never get here
-    assert False
 
 
 def _s3_open_uri(parsed_uri, mode, **kwargs):
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -381,12 +381,31 @@ def _open_binary_stream(uri, mode, **kw):
         if host is not None:
             kw['endpoint_url'] = _add_scheme_to_host(host)
         return smart_open_s3.open(uri.bucket.name, uri.name, mode, **kw), uri.name
-    elif hasattr(uri, 'read'):
+    elif _is_stream(uri, mode):
         # simply pass-through if already a file-like
-        filename = '/tmp/unknown'
+        filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
         raise TypeError('don\'t know how to handle uri %s' % repr(uri))
+
+
+def _is_stream(fileobj, mode):
+    """
+    Detect whether the specified object is a file object with required capabilities
+    implied by `mode`.
+    """
+    has_read = hasattr(fileobj, 'read')
+    has_write = hasattr(fileobj, 'write')
+    if not has_read and not has_write:
+        return False
+    if mode.endswith('+'):
+        return has_read and has_write
+    if mode[0] == 'r':
+        return has_read
+    if mode[0] in ('w', 'a'):
+        return has_write
+    # we should never get here
+    return False
 
 
 def _s3_open_uri(parsed_uri, mode, **kwargs):
@@ -582,6 +601,8 @@ def _compression_wrapper(file_obj, filename, mode):
     if _need_to_buffer(file_obj, mode, ext):
         warnings.warn('streaming gzip support unavailable, see %s' % _ISSUE_189_URL)
         file_obj = io.BytesIO(file_obj.read())
+    if ext in COMPRESSED_EXT and mode.endswith('+'):
+        raise ValueError('impossible to open in read+write+compressed mode')
 
     if ext == '.bz2':
         return BZ2File(file_obj, mode)
@@ -622,11 +643,11 @@ def _encoding_wrapper(fileobj, mode, encoding=None, errors=DEFAULT_ERRORS):
     if encoding is None:
         encoding = SYSTEM_ENCODING
 
-    if mode[0] == 'r':
-        decoder = codecs.getreader(encoding)
-    else:
-        decoder = codecs.getwriter(encoding)
-    return decoder(fileobj, errors=errors)
+    if mode[0] == 'r' or mode.endswith('+'):
+        fileobj = codecs.getreader(encoding)(fileobj, errors=errors)
+    if mode[0] in ('w', 'a') or mode.endswith('+'):
+        fileobj = codecs.getwriter(encoding)(fileobj, errors=errors)
+    return fileobj
 
 def _add_scheme_to_host(host):
     if host.startswith('http://') or host.startswith('https://'):
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -383,10 +383,15 @@ def _open_binary_stream(uri, mode, **kw):
         return smart_open_s3.open(uri.bucket.name, uri.name, mode, **kw), uri.name
     elif _is_stream(uri, mode):
         # simply pass-through if already a file-like
+        # we need to return something as the file name, but we don't know what
+        # so we probe for uri.name (e.g., this works with tempfile.NamedTemporaryFile)
+        # if the value ends with COMPRESSED_EXT, we will note it in _compression_wrapper()
+        # if there is no such an attribute, we return "unknown" - this effectively disables
+        # any compression (the actual string does not matter ofc)
         filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
-        raise TypeError('don\'t know how to handle uri %s' % repr(uri))
+        raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
 
 
 def _is_stream(fileobj, mode):
@@ -405,7 +410,7 @@ def _is_stream(fileobj, mode):
     if mode[0] in ('w', 'a'):
         return has_write
     # we should never get here
-    return False
+    assert False
 
 
 def _s3_open_uri(parsed_uri, mode, **kwargs):
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -8,7 +8,6 @@
 
 import bz2
 import io
-import functools
 import unittest
 import logging
 import tempfile
@@ -29,7 +28,6 @@ from smart_open import smart_open_lib
 
 logger = logging.getLogger(__name__)
 
-PY2 = sys.version_info[0] == 2
 CURR_DIR = os.path.abspath(os.path.dirname(__file__))
 SAMPLE_TEXT = 'Hello, world!'
 SAMPLE_BYTES = SAMPLE_TEXT.encode('utf-8')
@@ -207,8 +205,7 @@ class SmartOpenHttpTest(unittest.TestCase):
     def _test_compressed_http(self, suffix, query):
         """Can open <suffix> via http?"""
         raw_data = b'Hello World Compressed.' * 10000
-        buffer = six.BytesIO()
-        buffer.name = 'data' + suffix
+        buffer = make_buffer(name='data' + suffix)
         with smart_open.smart_open(buffer, 'wb') as outfile:
             outfile.write(raw_data)
         compressed_data = buffer.getvalue()
@@ -261,7 +258,7 @@ def make_buffer(cls=six.BytesIO, initial_value=None, name=None):
     buf = cls(initial_value) if initial_value else cls()
     if name is not None:
         buf.name = name
-    if PY2:
+    if six.PY2:
         buf.__enter__ = lambda: buf
         buf.__exit__ = lambda exc_type, exc_val, exc_tb: None
     return buf
@@ -286,14 +283,14 @@ class SmartOpenFileObjTest(unittest.TestCase):
             sf.write(SAMPLE_BYTES)
             self.assertEqual(buffer.getvalue(), SAMPLE_BYTES)
 
-    @unittest.skipIf(PY2, "Python 2 does not differentiate between str and bytes")
+    @unittest.skipIf(six.PY2, "Python 2 does not differentiate between str and bytes")
     def test_read_text_stream_fails(self):
         """Attempts to read directly from a text stream should fail."""
-        buffer = make_buffer(six.StringIO, SAMPLE_TEXT)
+        buffer = make_buffer(six.StringIO, initial_value=SAMPLE_TEXT)
         with smart_open.smart_open(buffer, 'r') as sf:
             self.assertRaises(TypeError, sf.read)  # we expect binary mode
 
-    @unittest.skipIf(PY2, "Python 2 does not differentiate between str and bytes")
+    @unittest.skipIf(six.PY2, "Python 2 does not differentiate between str and bytes")
     def test_write_text_stream_fails(self):
         """Attempts to write directly to a text stream should fail."""
         buffer = make_buffer(six.StringIO)
@@ -948,20 +945,14 @@ class CompressionFormatTest(unittest.TestCase):
     """
 
     def write_read_assertion(self, suffix):
-        with tempfile.NamedTemporaryFile('wb', suffix=suffix, delete=False) as infile:
-            test_file = infile.name
-
-        with smart_open.smart_open(test_file, 'wb') as fout:  # 'b' for binary, needed on Windows
+        test_file = make_buffer(name='file' + suffix)
+        with smart_open.smart_open(test_file, 'wb') as fout:
             fout.write(SAMPLE_BYTES)
-
-        with open(test_file, 'rb') as fin:
-            self.assertNotEqual(SAMPLE_BYTES, fin.read())
-
+        self.assertNotEqual(SAMPLE_BYTES, test_file.getvalue())
+        # we have to recreate the buffer because it is closed
+        test_file = make_buffer(initial_value=test_file.getvalue(), name=test_file.name)
         with smart_open.smart_open(test_file, 'rb') as fin:
-            self.assertEqual(fin.read().decode('utf8'), SAMPLE_TEXT)
-
-        if os.path.isfile(test_file):
-            os.unlink(test_file)
+            self.assertEqual(fin.read(), SAMPLE_BYTES)
 
     def test_open_gz(self):
         """Can open gzip?"""
@@ -1055,7 +1046,7 @@ class MultistreamsBZ2Test(unittest.TestCase):
             os.unlink(test_file)
 
     def test_can_read_multistream_bz2(self):
-        if PY2:
+        if six.PY2:
             # this is a backport from Python 3
             from bz2file import BZ2File
         else:
@@ -1068,7 +1059,7 @@ class MultistreamsBZ2Test(unittest.TestCase):
 
     def test_python2_stdlib_bz2_cannot_read_multistream(self):
         # Multistream bzip is included in Python 3
-        if not PY2:
+        if not six.PY2:
             return
         import bz2
 
--- a/smart_open/tests/test_smart_open.py
+++ b/smart_open/tests/test_smart_open.py
@@ -6,6 +6,7 @@
 # This code is distributed under the terms and conditions
 # from the MIT License (MIT).
 
+from bz2 import compress as bzip2_compress, decompress as bzip2_decompress
 import io
 import unittest
 import logging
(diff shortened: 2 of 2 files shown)


smart_open/smart_open_lib.py
```diff
@@ -381,14 +381,33 @@ def _open_binary_stream(uri, mode, **kw):
...
+    if not has_read and not has_write:
+        return False
+    if mode.endswith('+'):
+        return has_read and has_write
+    if mode[0] == 'r':
+        return has_read
+    if mode[0] in ('w', 'a'):
+        return has_write
+    # we should never get here
+    return False
```
piskvorky: If we should never get here, `assert False` better?

smart_open/smart_open_lib.py
```diff
@@ -381,14 +381,33 @@ def _open_binary_stream(uri, mode, **kw):
...
             kw['endpoint_url'] = _add_scheme_to_host(host)
         return smart_open_s3.open(uri.bucket.name, uri.name, mode, **kw), uri.name
-    elif hasattr(uri, 'read'):
+    elif _is_stream(uri, mode):
         # simply pass-through if already a file-like
-        filename = '/tmp/unknown'
+        filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
         raise TypeError('don\'t know how to handle uri %s' % repr(uri))
```
piskvorky:
```suggestion
        raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
```

smart_open/smart_open_lib.py
```diff
@@ -381,14 +381,33 @@ def _open_binary_stream(uri, mode, **kw):
         if host is not None:
             kw['endpoint_url'] = _add_scheme_to_host(host)
         return smart_open_s3.open(uri.bucket.name, uri.name, mode, **kw), uri.name
-    elif hasattr(uri, 'read'):
+    elif _is_stream(uri, mode):
         # simply pass-through if already a file-like
-        filename = '/tmp/unknown'
+        filename = getattr(uri, 'name', 'unknown')
```
piskvorky: What is this `name`? Code deserves a comment.

vmarkovtsev:
@piskvorky CI passed, review suggestions addressed.

Regarding the `name`, I was initially guided by `tempfile.NamedTemporaryFile` convention. Besides,
```python
print(open("smart_open/tests/test_smart_open.py").name)
```
does print the opened file name. I added comments at that line.

vmarkovtsev:
As a nice side effect, this will work:

```
with open("file.gz", "rb") as f:
    decompressed_data = smart_open(f).read()
```

piskvorky:
Thanks a lot! That "side effect" is awesome, love it :)

smart_open/tests/test_smart_open.py
```diff
@@ -200,92 +201,173 @@ def test_http_pass(self):
...
+        file = self.IOWriteWrapper()
+        text = 'Hello, world!'
+        with smart_open.smart_open(file, 'w') as sf:
+            sf.write(text)
+            self.assertEqual(file.fobj.getvalue(), text.encode())
+        file = self.IOReadWrapper()  # no `write()`
+        self.assertRaises(TypeError, lambda: smart_open.smart_open(file, 'w'))
+
+    def test_read_write(self):
+        file = six.BytesIO()
```
mpenkov: You can avoid the instantiate-write-seek pattern with:

```python
buf = six.BytesIO(text.encode())
```

That will initialize the buffer, fill it with the required contents, and seek to the beginning of the buffer in one call. Please apply this here and elsewhere.

smart_open/tests/test_smart_open.py
```diff
@@ -200,92 +201,173 @@ def test_http_pass(self):
...
+        self.assertRaises(TypeError, lambda: smart_open.smart_open(file, 'r+'))
+        self.assertRaises(TypeError, lambda: smart_open.smart_open(file, 'w+'))
+        self.assertRaises(TypeError, lambda: smart_open.smart_open(file, 'a+'))
+
+    def test_append(self):
+        file = self.IOWriteWrapper()
+        text = 'Hello, world!'
+        with smart_open.smart_open(file, 'a') as sf:
+            sf.write(text)
+            self.assertEqual(file.fobj.getvalue(), text.encode())
```
mpenkov: Please move the assertion to outside of the context manager (here and elsewhere).

vmarkovtsev: That's intentional: the file object is closed in the context manager and `getvalue` raises an exception

smart_open/smart_open_lib.py
```diff
@@ -381,12 +381,36 @@ def _open_binary_stream(uri, mode, **kw):
...
+        # if there is no such an attribute, we return "unknown" - this effectively disables
+        # any compression (the actual string does not matter ofc)
+        filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
-        raise TypeError('don\'t know how to handle uri %s' % repr(uri))
+        raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
+
+
+def _is_stream(fileobj, mode):
```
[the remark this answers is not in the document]
vmarkovtsev: Good observation! I will change the function to check `readable` and `writeable` attribute values instead, if they are present.

There is also an issue with `seek()`, which may be required by some compression codecs. I have not had time to dig it deeply yet, but from my experience, this may be a real compatibility problem.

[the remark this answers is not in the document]
vmarkovtsev: Besides, regarding `read` and `write`, we can pass any custom class without some methods implemented. Pretty common practice with some streams derived from networking. After a second thought, I tend to leave my checks, but also improve the compatibility with the standard streams.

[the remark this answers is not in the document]
mpenkov: Personally, I think it's better off to **not** check those things at all. This is both more Pythonic (duck typing, asking for forgiveness instead of permission, and a couple of other things) and provides more clarity to the user.

With this function, if a required method is missing or unsupported, the user will get a general TypeError:

    "don't know how to handle uri %r in mode %r"

Now, if we were a user application, that would be a decent error message. We're not an application, though, we're a library, so the above message is less helpful than the alternative.

The alternative is to let smart_open try to read/write from the stream, and if something is missing or unimplemented, allow it to raise the appropriate exception. That will tell the user of our library exactly **why** we can't handle the stream they gave us. A good example is the seek scenario - if a user gives us an unseekable stream and we absolutely **must** seek it, they'll get the precise reason why we can't process it, not a generic error message.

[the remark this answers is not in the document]
vmarkovtsev: Duck typing does not hinder user-friendly assertions before a shipwreck happens. In my experience, getting an AttributeError somewhere on a 30th place in the call stack does not really help users to figure out what went wrong. It is often the opposite - they have no idea what is going on in the implementation details and they are going to open an issue on GitHub to ask if there is a bug in the library (true story, sadly).

Thus if we are sure that some particular methods or modes will be used in the future - we should check them *before* bumming an unfriendly stack trace. At the same time, as long as we don't mess with `isinstance`, our duck typing quacks.

You've got a valid concern regarding the usefulness of the generic `TypeError`. Indeed, I am going to change the code to raise proper errors which point to the exact problems.

smart_open/tests/test_smart_open.py
```diff
@@ -200,92 +201,173 @@ def test_http_pass(self):
...
-        with tempfile.NamedTemporaryFile('wb', suffix='.bz2', delete=False) as infile:
-            test_file = infile.name
+class SmartOpenFileObjTest(unittest.TestCase):
+    """
+    Test passing raw file objects.
+    """
 
-        with smart_open.smart_open(test_file, 'wb') as outfile:
-            outfile.write(test_string)
+    class IOReadWrapper(object):
```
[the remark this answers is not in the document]
vmarkovtsev: The actual goal of this feature is to allow users to pass custom file object-like instances. This is why I say that https://github.com/RaRe-Technologies/smart_open/pull/263#issuecomment-466686942 is a side effect: we don't really expect people to use `smart_open` this way, do we.

smart_open/tests/test_smart_open.py
```diff
@@ -200,92 +201,173 @@ def test_http_pass(self):
...
 
-        if os.path.isfile(test_file):
-            os.unlink(test_file)
+        def close(self):
+            self.fobj.close()
 
-        responses.add(responses.GET, "http://127.0.0.1/data.xz", body=compressed_data, stream=True)
-        smart_open_object = smart_open.smart_open("http://127.0.0.1/data.xz")
+    def test_read(self):
+        file = self.IOReadWrapper()
```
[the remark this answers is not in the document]
vmarkovtsev: @mpenkov FWIW `file` is *not* a keyword. Proof:

```python
>>> import keyword
>>> keyword.iskeyword("file")
```

It is a *built-in type* and only in Python2:

```python
>>> import builtins
>>> "file" in dir(builtins)
```

But yeah, I am just being pedantic - ofc I will change the name :smile: It's been a long time since I last worked with py2.

smart_open/smart_open_lib.py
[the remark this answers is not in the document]
mpenkov: While I think the PR is good in general, I'm **strongly** opposed to _this_ particular change for the following reasons:

1. Little practical value
2. Lack of correctness
3. Introduction of unnecessary coupling
4. Disproportionate maintenance overhead

I explain each of these points in more detail below.

---

First, the new code adds little practical value. The existing code (checking for the read attribute) already handles 99% of the use case correctly: as I've already mentioned, all streams from the standard library have a read method, regardless of whether they're actually readable. The existing code gives informative and correct information via exception handling. For example, here's an attempt to write to a read-only stream:

```
$ python -c 'import smart_open as s;s.smart_open(open("README.rst")).write("no")'
  File "<string>", line 1, in <module>
io.UnsupportedOperation: not writable
```

> In my experience, getting an AttributeError somewhere on a 30th place in the call stack does not really help users to figure out what went wrong.

This is not a valid argument. Both the new and old code raise exceptions at similar levels in the call stack, so the experience for the user will be similar.

Attempts to improve on the existing 99% are subject to diminishing returns. More specifically, the new code is spending an unnecessary amount of effort preparing for edge cases that do not occur in practice. This is why the new unit tests need custom buffer classes: no I/O classes in the Python standard library behave in the way the new code and tests expect streams to behave. It's like boiling the ocean to make a cup of tea.

---

Second, the new code still does not achieve its intended goal: successfully checking whether an input parameter is a stream that we can work with. You've checked for presence of the read/write function, but that isn't enough:

1. Is the stream actually [readable/writable](https://docs.python.org/3.5/library/io.html#io.IOBase.readable)?
1. Does the stream support seeking? Does it support seeking for particular values of the `whence` parameter (e.g. seek from end of file)?
2. Is the stream actually open?
3. Is it a binary or unicode stream? Is that compatible with the mode we're working with?
4. ... and much more

I'm not proposing you actually implement checks for the above. I'm telling you that the only way to 100% effectively and correctly determine whether you can peform an I/O operation on a stream (e.g. read or write) is to actually try it and see.

Also, the exception raised by the new code is misleading. TypeError suggests there is a problem with the _type_, which isn't really the case here.

---

Third, the new code introduces unnecessary coupling between input validation and the actual implementation.

> Thus if we are sure that some particular methods or modes will be used in the future - we should check them **before** bumming an unfriendly stack trace.

This is a bad idea. The new validation code essentially needs to know details of the implementation. You've introduced coupling: if the implementation changes in the future, we'll need to remember to update the validation code. What's worse, this coupling is _hidden inside_ the new input validation function. So now, whoever wants to update the implementation needs to magically know to update the input validation as well. It's an unnecessary burden.

---

Fourth, this change adds unnecessary maintenance overhead for little practical value (see my first point above). As a maintainer, which option would you choose?

1. Old implementation, 99% correct, one line of code, covered by unit tests
2. New implementation, 99% correct, 50 new lines of code, 100 new lines of tests

---

I hope I've conveyed my position clearly.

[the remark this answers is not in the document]
vmarkovtsev: Thank you @mpenkov  for such an interesting discussion! First of all, you still did not convince me and my opinion is not changed, **but** I will revert the code - it is not worth writing a paper here, and you are the maintainer :)

Second, you are commenting on the old version. Here is the current one:
<details>
<pre>
def _check_stream(fileobj, mode):
    """
    Detect whether the specified object is a file object with required capabilities
    implied by `mode`. Raise a TypeError if the file object capabilities are insufficient.

    Return True if `fileobj` is a file object; otherwise, False.
    """
    has_read = hasattr(fileobj, 'read')
    has_write = hasattr(fileobj, 'write')
    if not has_read and not has_write:
        return False
    # otherwise, we consider that we have a file object
    # do we have the required methods?
    # we always do for open()-like file objects, however, it may not be the case with custom types
    if mode.endswith('+') and (not has_read or not has_write):
        raise TypeError(
            'File object-like %r does not have the "read" or "write" methods required by mode %s' % (fileobj, mode))
    if mode[0] == 'r' and not has_read:
        raise TypeError('File object-like %r does not have the "read" method required by mode %s' % (fileobj, mode))
    if mode[0] in ('w', 'a') and not has_write:
        raise TypeError('File object-like %r does not have the "write" method required by mode %s' % (fileobj, mode))
    # ok, methods are here, but we need to further check the mode of open()-like file objects
    try:
        open_mode = getattr(fileobj, 'mode')
    except AttributeError:
        # this is not a file object derived from open(), let's hope for the best
        return True
    if (mode.endswith('+') and not open_mode.endswith('+')) or \
            (not open_mode.endswith('+') and open_mode[0] != mode[0]):
        raise TypeError("%r was opened in an incompatible mode: need %s have %s" % (fileobj, mode, open_mode))
    return True
</pre>
</details>

Below is my view on your points.

### Practical valueThere is a big difference between getting a raw exception and a targetted exception. Compare:

```
io.UnsupportedOperation: not writable
```

and

```python
<_io.TextIOWrapper name='setup.py' mode='r' encoding='UTF-8'> was opened in an incompatible mode: need wb have r
```

We see at once that:

* We opened "setup.py"
* We've got a conflict between the modes

I do understand that you may be used to a different way of developing software, and `smart_open` may have well-educated, capable for debugging users. This is my (somewhat long) experience.

> More specifically, the new code is spending an unnecessary amount of effort preparing for edge cases that do not occur in practice.

Again: please, comprehend me, I already wrote it: the standard streams **is not** the target goal. Custom classes are.

### Lack of correctness

100% of the cases are not checked, of course. So are the assertions in any unit tests, for example.  One targets covering the most important points first, or those which are easily to cover. Rome was not build in a day, and iteration by iteration, this code may be improved in the future.

I also refer to programming by contract. The old way of specifying the contract was implicit: it breaks where it breaks. The new way is definitely explicit. Now let's read Zen of Python, no. 2.
You can still argue whether it is Pythonic or not though - it is quite subjective. And after all, "practicality beats purity".

### Introduction of unnecessary coupling

I don't know what to say about your understanding of coupling. I will ask! What's your attitude to PEP 484 (type hints) and mypy in particular?

### Disproportionate maintenance overhead

This is hard to oppose because an absense of code is the best code, of course. However, I don't agree with the "correctness" thesis: the lack of features is also correct by definition, and I am adding a new feature.

So happy that I am doing open source! Thanks once again for a great discussion. Maybe I will convert it to a blog post - I know some folks who may be interseted.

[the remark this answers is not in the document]
mpenkov: > Again: please, comprehend me, I already wrote it: the standard streams **is not** the target goal. Custom classes are.

_Properly implemented_ custom I/O classes will behave _identically_ to standard streams, because they will inherit from the [io.IOBase](https://docs.python.org/2/library/io.html#i-o-base-classes) abstract class (or something similar from the io module). That's literally what the entire io module was made for. Now,

> There should be one-- and preferably only one --obvious way to do it.

and the io stdlib is that way, in this case.

That narrows your target goal even further: now you're targeting custom classes that don't inherit from stdlib IO classes (for whatever reason), and don't even bother to implement the standard interface. It's nice that you're being considerate and trying to handle such implementations, but you're preparing for a problem that doesn't exist yet, and your effort essentially duplicates the IOBase abstract class.

> I also refer to programming by contract. The old way of specifying the contract was implicit: it breaks where it breaks. The new way is definitely explicit. 
> Now let's read Zen of Python, no. 2.

Don't stop reading at point no. 2, no. 3 and 4 are no less important ;)

> What's your attitude to PEP 484 (type hints) and mypy in particular?

I'm indifferent to type hints and mypy. I don't suffer from the problems that they solve, so I'm not in a rush to adopt them.

> So happy that I am doing open source! Thanks once again for a great discussion. Maybe I will convert it to a blog post - I know some folks who may be interseted.

Glad you're enjoying the experience. Again, thank you for your work and the discussion. Be sure to let me know how the blog post goes.

vmarkovtsev:
@mpenkov Review applied, the controversial function removed, ready for the next round.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
-        responses.add(responses.GET, "http://127.0.0.1/data.xz", body=compressed_data, stream=True)
-        smart_open_object = smart_open.smart_open("http://127.0.0.1/data.xz")
 
-        # decompress the xz and get the same md5 hash
-        self.assertEqual(smart_open_object.read(), test_string)
+class SmartOpenFileObjTest(unittest.TestCase):
+    """
+    Test passing raw file objects.
+    """
+    text = 'Hello, world!'
```
mpenkov: Please move to module scope and mark as a constant (TEXT = '...').

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
 
+    def patch_string_io(func):
```
mpenkov: First, this does not need to be a static method. It's simpler as a pure function.

Also, I think this patch_string_io + decorator stuff is more complicated than it needs to be. AFAIK, it's there because you want the buffer to be with-able (a context manager). Wouldn't this be simpler?

```python
def make_named_buffer(klass=six.BytesIO, value=None, name=None):
    buf = klass(value) if value else klass()
    if name:
        buf.name = name
    if six.PY2:
        buf.__enter__ = lambda self: self
        buf.__exit__ = lambda self, exc_type, exc_val, exc_tb: None
    return buf

binbuf = make_named_buffer(name='binary')
strbuf = make_named_buffer(six.StringIO, name='text')
```

That hits several birds with one stone:

- Shortcuts creating a named buffer (either binary or text) - see [previous review](https://github.com/RaRe-Technologies/smart_open/pull/263#discussion_r259609156)
- Simplifies the existing code (no wrappers, functools, decorators, static methods or other voodoo)

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+            finally:
+                del six.StringIO.__enter__
+                del six.StringIO.__exit__
+
+        return wrapped_patch_string_io
+
+    @patch_string_io
+    def test_read_bytes(self):
+        """Can we read bytes from a byte stream?"""
+        buffer = six.BytesIO(self.text.encode())
```
mpenkov: This looks strange, you're patching six.StringIO, but using BytesIO as a context manager here.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+            sf.write(data)
+        self.assertEqual(bz2.decompress(buffer.getvalue()), data)
+
+    def test_open_side_effect(self):
+        """Does our trick with "name" attribute work with open()?"""
+        data = b'Hello, world!' * 1000
+        with tempfile.NamedTemporaryFile(prefix='smart_open_tests_', delete=False) as tmpf:
+            tmpf.write(data)
+        try:
+            with open(tmpf.name, 'rb') as openf:
```
mpenkov: What are you actually testing here?

1. Open a file given its path as a string
2. Read its contents
3. Compare the read contents against known values

How is this different to other tests that do the same thing? Why do you need to create and delete a temporary file as part of the test? It looks unnecessary. Please let me know if I'm missing something here.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+        """Can we use the "name" attribute to compress on the fly?"""
+        data = b'Hello, world!' * 1000
+        buffer = six.BytesIO()
+        buffer.name = 'data.bz2'
+        with smart_open.smart_open(buffer, 'wb') as sf:
+            sf.write(data)
+        self.assertEqual(bz2.decompress(buffer.getvalue()), data)
+
+    def test_open_side_effect(self):
+        """Does our trick with "name" attribute work with open()?"""
```
mpenkov: What trick are you referring to? If it's not obvious right now, it definitely won't be obvious to anyone after a few weeks.

smart_open/smart_open_lib.py
```diff
@@ -383,10 +383,14 @@ def _open_binary_stream(uri, mode, **kw):
...
-        filename = '/tmp/unknown'
+        # we need to return something as the file name, but we don't know what
+        # so we probe for uri.name (e.g., this works with open() or tempfile.NamedTemporaryFile)
+        # if the value ends with COMPRESSED_EXT, we will note it in _compression_wrapper()
+        # if there is no such an attribute, we return "unknown" - this effectively disables any compression
+        filename = getattr(uri, 'name', 'unknown')
         return uri, filename
     else:
-        raise TypeError('don\'t know how to handle uri %s' % repr(uri))
+        raise TypeError("don't know how to handle uri %r in mode %r" % (uri, mode))
```
mpenkov: I don't think the mode is relevant here. This if block doesn't check the mode, it only checks the uri.

smart_open/smart_open_lib.py
```diff
@@ -582,6 +586,8 @@ def _compression_wrapper(file_obj, filename, mode):
     if _need_to_buffer(file_obj, mode, ext):
         warnings.warn('streaming gzip support unavailable, see %s' % _ISSUE_189_URL)
         file_obj = io.BytesIO(file_obj.read())
+    if ext in COMPRESSED_EXT and mode.endswith('+'):
+        raise ValueError('impossible to open in read+write+compressed mode')
```
mpenkov: Nothing is impossible :) Let's use a more positive exception message.

```suggestion
        raise ValueError('transparent (de)compression currently unsupported for mode %r' % mode)
```

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+        with smart_open.smart_open(buffer, 'rb') as sf:
+            data = sf.read()
+        self.assertEqual(data, self.text.encode())
+
+    @patch_string_io
+    def test_write_bytes(self):
+        """Can we write bytes to a byte stream?"""
+        buffer = six.BytesIO()
+        with smart_open.smart_open(buffer, 'wb') as sf:
+            sf.write(self.text.encode())
```
mpenkov: We're using TEXT.encode() a lot. It's making our tests more difficult to read than they really need to be.

If it's worth declaring TEXT as a constant, then it's worth declaring its corresponding bytes as a constant too.

```
TEXT = 'hello world'
BYTES = TEXT.encode('utf-8')
```

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+            self.assertEqual(buffer.getvalue(), self.text.encode())
+
+    @unittest.skipIf(PY2, "Python 2 does not make a difference between str and bytes")
```
mpenkov: Fixed grammar.

```suggestion
    @unittest.skipIf(PY2, "Python 2 does not differentiate between str and bytes")
```

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+    @patch_string_io
+    def test_read_str(self):
+        """Can we read strings from a text stream?"""
+        buffer = six.StringIO(self.text)
+        with smart_open.smart_open(buffer, 'r') as sf:
+            self.assertRaises(TypeError, sf.read)  # we expect binary mode
+
+    @unittest.skipIf(PY2, "Python 2 does not make a difference between str and bytes")
```
mpenkov: Grammar

```suggestion
    @unittest.skipIf(PY2, "Python 2 does not make differentiate between str and bytes")
```

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+    @patch_string_io
+    def test_write_str(self):
+        """Can we write strings to a text stream?"""
+        buffer = six.StringIO()
+        with smart_open.smart_open(buffer, 'w') as sf:
+            self.assertRaises(TypeError, sf.write, self.text)  # we expect binary mode
```
mpenkov: What's the actual problem here?

The buffer is text. You've opened it in write text mode. You're writing text to it. Why should the test fail?

The name of the test and the docstring may be wrong, too.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +202,158 @@ def test_http_pass(self):
...
+            sf.write(self.text.encode())
+            self.assertEqual(buffer.getvalue(), self.text.encode())
+
+    @unittest.skipIf(PY2, "Python 2 does not make a difference between str and bytes")
+    @patch_string_io
+    def test_read_str(self):
+        """Can we read strings from a text stream?"""
+        buffer = six.StringIO(self.text)
+        with smart_open.smart_open(buffer, 'r') as sf:
+            self.assertRaises(TypeError, sf.read)  # we expect binary mode
```
mpenkov: Sorry, I don't understand: what's the problem here?

- There's a text buffer
- You've opened it in text mode for reading
- You're trying to read from it

Why do "we expect binary mode"? Why is this test supposed to fail?

smart_open/tests/test_smart_open.py
vmarkovtsev: Because of Python 2, where both are the same.

smart_open/smart_open_lib.py
vmarkovtsev: You need to agree with @piskvorky here https://github.com/RaRe-Technologies/smart_open/pull/263#discussion_r259590067

smart_open/smart_open_lib.py
vmarkovtsev: I would remove "currently" to remove the false sense of hope. The way codecs are implemented is incompatible with rw, and nobody is going to rewrite them in Python core.

smart_open/tests/test_smart_open.py
vmarkovtsev: The problem is that `smart_open` always assumes that the wrapped stream is a binary stream. We do not support writing to text streams, and this test checks that.

smart_open/tests/test_smart_open.py
vmarkovtsev: https://github.com/RaRe-Technologies/smart_open/pull/263#discussion_r260144512 binary and text buffers behave differently on Python 3

smart_open/smart_open_lib.py
mpenkov: @piskvorky, thoughts?

smart_open/tests/test_smart_open.py
vmarkovtsev: I am missing `.bz2` suffix here, right.

smart_open/tests/test_smart_open.py
mpenkov: I see. Let's use a more informative test name and docstring here.

```python
def test_write_text_stream_fails(self):
    """smart_open does not support writing to a text stream.""" 
```

smart_open/tests/test_smart_open.py
mpenkov: OK, as above then, let's have a more informative test name docstring.

smart_open/tests/test_smart_open.py
mpenkov: OK, then we have another reason to avoid the wrapper-decorator approach :)

smart_open/tests/test_smart_open.py
mpenkov: So you're testing that you can do I/O with transparent decompression to file.bz2?

Let's make that obvious from the test name and docstring. test_side_effect tells nothing to the reader.

smart_open/tests/test_smart_open.py
piskvorky: > The problem is that `smart_open` always assumes that the wrapped stream is a binary stream.

Is this by necessity? Or why do we assume that?

mpenkov: If it’s a text stream, there’s no new functionality that smart_open can add to it. Reading and writing is possible, but it’d be the same as doing stream.read() and stream.write() directly.

vmarkovtsev: Besides, there is no way to check if the wrapped stream is binary or text.

smart_open/smart_open_lib.py
piskvorky: The previous version used complex URI checking, which depended on the `mode`, hence the change in logging.

If the current version doesn't rely on `mode` any more (reverted), I guess we don't need to show it (but probably no harm either).

mpenkov: My motivation for removing the mode is that keeping it misleads the user into thinking mode is part of the problem. In this particular case, it isn’t.

smart_open/tests/test_smart_open.py
vmarkovtsev: We already test in-memory with `test_name_read`. This is the test to ensure that the regular `open()` works, and it requires a real file.

Adding more info to the docstring.

smart_open/tests/test_smart_open.py
vmarkovtsev: Funnily enough, this was the previous code before I decided to go decorated. Changing.

vmarkovtsev: Python 2 fails with this code

```
======================================================================
ERROR: test_read_bytes (__main__.SmartOpenFileObjTest)
Can we read bytes from a byte stream?
----------------------------------------------------------------------
  File "test_smart_open.py", line 278, in test_read_bytes
    with smart_open.smart_open(buffer, 'rb') as sf:
TypeError: <lambda>() takes exactly 1 argument (0 given)

======================================================================
ERROR: test_write_bytes (__main__.SmartOpenFileObjTest)
Can we write bytes to a byte stream?
----------------------------------------------------------------------
  File "test_smart_open.py", line 285, in test_write_bytes
    with smart_open.smart_open(buffer, 'wb') as sf:
TypeError: <lambda>() takes exactly 1 argument (0 given)

----------------------------------------------------------------------
```

vmarkovtsev: I see, because we set the attribute on the instance, not on the type

vmarkovtsev:
Addressed

smart_open/tests/test_smart_open.py
```diff
@@ -6,7 +6,9 @@
 # This code is distributed under the terms and conditions
 # from the MIT License (MIT).
 
+import bz2
 import io
+import functools
```
mpenkov: Unused import

smart_open/tests/test_smart_open.py
```diff
@@ -29,6 +31,8 @@
 
 PY2 = sys.version_info[0] == 2
```
mpenkov: We don't need this, we have six.PY2 already.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +204,150 @@ def test_http_pass(self):
...
-        fpath = os.path.join(CURR_DIR, 'test_data/crlf_at_1k_boundary.warc.gz')
-        with open(fpath, 'rb') as infile:
-            data = infile.read()
-
-        with gzip.GzipFile(fpath) as fin:
-            expected_hash = hashlib.md5(fin.read()).hexdigest()
+    def _test_compressed_http(self, suffix, query):
+        """Can open <suffix> via http?"""
+        raw_data = b'Hello World Compressed.' * 10000
+        buffer = six.BytesIO()
```
mpenkov: Please use the `make_buffer` function to create the buffer and name it.

smart_open/tests/test_smart_open.py
```diff
@@ -200,93 +204,150 @@ def test_http_pass(self):
...
+        """Can we write bytes to a byte stream?"""
+        buffer = make_buffer()
+        with smart_open.smart_open(buffer, 'wb') as sf:
+            sf.write(SAMPLE_BYTES)
+            self.assertEqual(buffer.getvalue(), SAMPLE_BYTES)
+
+    @unittest.skipIf(PY2, "Python 2 does not differentiate between str and bytes")
+    def test_read_text_stream_fails(self):
+        """Attempts to read directly from a text stream should fail."""
+        buffer = make_buffer(six.StringIO, SAMPLE_TEXT)
```
mpenkov: For symmetry and consistency:

```suggestion
        buffer = make_buffer(six.StringIO, initial_value=SAMPLE_TEXT)
```

smart_open/tests/test_smart_open.py
```diff
@@ -886,21 +947,18 @@ class CompressionFormatTest(unittest.TestCase):
...
-    TEXT = 'Hello'
-
     def write_read_assertion(self, suffix):
         with tempfile.NamedTemporaryFile('wb', suffix=suffix, delete=False) as infile:
             test_file = infile.name
 
-        text = self.TEXT.encode('utf8')
         with smart_open.smart_open(test_file, 'wb') as fout:  # 'b' for binary, needed on Windows
-            fout.write(text)
+            fout.write(SAMPLE_BYTES)
```
mpenkov: Why is it necessary for this test to write to a temporary file? Is using an in-memory buffer not sufficient for this test?

I'm not keen on unit tests that have side effects like writing to local storage. Sometimes its a necessary evil, but it smells more like an integration test than a unit test.

smart_open/tests/test_smart_open.py
```diff
@@ -886,21 +947,18 @@ class CompressionFormatTest(unittest.TestCase):
...
 
         with open(test_file, 'rb') as fin:
-            self.assertNotEqual(text, fin.read())
+            self.assertNotEqual(SAMPLE_BYTES, fin.read())
 
         with smart_open.smart_open(test_file, 'rb') as fin:
-            self.assertEqual(fin.read().decode('utf8'), self.TEXT)
+            self.assertEqual(fin.read().decode('utf8'), SAMPLE_TEXT)
```
mpenkov: Isn't this the same as:

```suggestion
            self.assertEqual(fin.read(), SAMPLE_BYTES)
```

vmarkovtsev:
Addressed

mpenkov:
@piskvorky Looks good enough to merge for me, WDYT?

mpenkov:
This PR has been heavily reviewed and is fairly low risk. In the interest of keeping the ball rolling, I'm going to merge it. @piskvorky Please let me know if there are any issues.

mpenkov:
@vmarkovtsev Thanks again for your contribution, keep them coming :+1:

PR #58, opened 2016-01-20
SSH/SCP/SFTP protocol support

val314159:
Uses paramiko library (nice and low level)

Solves: Issue https://github.com/piskvorky/smart_open/issues/36

PR #267, opened 2019-02-26
SSH/SCP/SFTP protocol support

### LINKED ISSUE #36: Add support for scp(ssh) protocol

mpenkov:
Continued from https://github.com/RaRe-Technologies/smart_open/pull/58 cause I couldn't write to the original author's repo.

Fix #36


smart_open/ssh.py
```diff
@@ -0,0 +1,81 @@
...
+
+try:
+    import paramiko
+except ImportError:
+    logger.warning('paramiko missing, SSH will be disabled.  `pip install paramiko` to suppress')
+
+#
+# Global storage for SSH connections.
+#
+_SSH = {}
```
piskvorky: How does such global cache play with multiprocessing / forking?

mpenkov: For multiprocessing: no problem, there will be one cache for subprocess.

For forking: I think it will be a problem. We don't want different threads to access the same connection simultaneously. Perhaps we can include the current thread_id in the cache key?

piskvorky: I don't think we have any solid contract in place in `smart_open` (yet). My naive expectation (as a user) would be "a connection open in one process/thread is always valid for that process/tread, but not necessarily across fork boundaries".

How do the other connection types (S3, HTTP, HDFS…) manage in the face of multiple processes or threads?

mpenkov: OTOH they "should be fine". We haven't had any concurrency-related bug reports.

On the other hand, we don't have any explicit testing for concurrency problems, either.

PR #268, opened 2019-02-26
Clean up keyword arguments

mpenkov:
My second crack at https://github.com/RaRe-Technologies/smart_open/pull/230

We used to liberally pass arbitrary keyword arguments (e.g. **kwargs). This is a bad practice because it's difficult to keep track of function parameters and ensure they're being passed down to the right function.

This PR starts to clean up our use of arbitrary keyword arguments.

Main changes:

- renamed smart_open.smart_open to smart_open.open
- renamed ignore_extension parameter to ignore_ext
- default open mode is now "r", to match standard library
- keyword arguments for that function now _completely_ match that of built-in open, with two exceptions (ignore_ext and tkwa)
- tkwa contains keyword arguments for whatever transport layer smart_open ends up using (e.g. HTTPS, S3, etc)
- Unsupported keyword arguments don't crash the library (same as old behavior) but cause a log warning (new behavior). Supported keyword arguments are identified via introspection.
- The transport keyword args are documented both in the top-level open function _and_ each individual module (S3, HTTPS, etc). There is no code duplication because of some docstring magic, hidden from the user.

Advantages:

- Simpler interface for the user, less arguments to open function
- Adding additional keyword arguments will no longer require updating the top-level interface
- Better documentation for keyword parameters (previously, they were documented via examples only)

### DIFF
--- a/smart_open/doctools.py
+++ b/smart_open/doctools.py
@@ -1,0 +1,114 @@
+"""Common functions for working with docstrings.
+
+For internal use only.
+"""
+import inspect
+import io
+
+
+def extract_kwargs(docstring):
+    """Extract keyword argument documentation from a function's docstring.
+
+    Parameters
+    ----------
+    docstring: str
+        The docstring to extract keyword arguments from.
+
+    Returns
+    -------
+    list of (str, str, list str)
+
+    str
+        The name of the keyword argument.
+    str
+        Its type.
+    str
+        Its documentation as a list of lines.
+
+    Notes
+    -----
+    The implementation is rather fragile.  It expects the following:
+
+    1. The parameters are under an underlined Parameters section
+    2. Keyword parameters have the literal ", optional" after the type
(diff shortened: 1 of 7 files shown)


smart_open/doctools.py
```diff
@@ -0,0 +1,114 @@
+"""Common functions for working with docstrings.
```
piskvorky: Missing file header.

smart_open/doctools.py
```diff
@@ -0,0 +1,121 @@
...
+#
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
+#
+# This code is distributed under the terms and conditions from the MIT License (MIT).
+#
+"""Common functions for working with docstrings.
+
+For internal use only.
+"""
+import inspect
```
piskvorky: Code style: blank line after (and before) a docstring.

PR #272, opened 2019-03-12
Accept custom HTTP headers when reading over HTTP(S)

canada11:
Http calls woudn't take headers before. I added two changes that allows http calls to now accept user defined headers for http calls.


smart_open/http.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, url, mode='r', buffer_size=DEFAULT_BUFFER_SIZE,
         self.buffer_size = buffer_size
         self.mode = mode
 
+        _HEADERS.update(headers)
```
piskvorky: Mutating global vars is a no-no (make a copy first).

canada11: Ok, I'll change it.

canada11: Better?

piskvorky: I think the other way makes more sense: overwriting (a copy of) our defaults with the user-supplied headers. Two reasons:

1. So that user-supplied headers takes precedence over defaults.
2. So that we don't mutate the user-supplied `headers` dict. It could lead to nasty surprises, even now -- since the default `headers` param is mutable, changing it will affect *all* future invocation of `__init__`, not just this one call!

The standard Python pattern is to avoid mutable params, using `headers=None` instead to denote the default, which is then set in the function body.

canada11: Makes sense.

canada11:
Does this version look good?

smart_open/http.py
```diff
@@ -30,7 +30,7 @@ class BufferedInputBase(io.BufferedIOBase):
     """
 
     def __init__(self, url, mode='r', buffer_size=DEFAULT_BUFFER_SIZE,
-                 kerberos=False, user=None, password=None):
+                 kerberos=False, user=None, password=None, headers=None):
```
mpenkov: The default headers are simple, so why not do?

```suggestion
                 kerberos=False, user=None, password=None, headers=_HEADERS):
```

piskvorky: -1 on that (for the same reason mentioned above -- mutable parameter defaults lead to subtle bugs).

smart_open/http.py
```diff
@@ -48,7 +48,12 @@ def __init__(self, url, mode='r', buffer_size=DEFAULT_BUFFER_SIZE,
         self.buffer_size = buffer_size
         self.mode = mode
 
-        self.response = requests.get(url, auth=auth, stream=True, headers=_HEADERS)
+        if headers:
```
piskvorky: `if headers is None:` safer.

Like this, if a user explicitly passes `headers={}`, we'd ignore his choice and use `_HEADERS` anyway.

smart_open/http.py
```diff
@@ -48,7 +48,12 @@ def __init__(self, url, mode='r', buffer_size=DEFAULT_BUFFER_SIZE,
...
+            self.headers = headers
+        else:
+            self.headers = _HEADERS
```
piskvorky: Similar objection: we don't want instances (individual objects) to share globals in any way.

To see what I mean, consider this scenario:

```python
x = BufferedInputBase()  # use default _HEADERS
x.headers['Accept-Encoding'] = 'compress, gzip'  # or whatever
x.headers['Other-Header'] = 'value'  # or whatever

y = BufferedInputBase()  # use default again, but…
print(y.headers)  # surprise!
```

mpenkov: Nice catch, thanks for the example @piskvorky

mpenkov: Perhaps we could do this:

```python
        if headers:
            self.headers = headers
        else:
            self.headers = dict(_HEADERS)
```

A shallow copy should be sufficient to ensure the contents of _HEADERS remain unchanged.

piskvorky: Yes, except `if headers is None` is the standard pattern (a user passing `{}` is not the same as default `None`).

canada11:
I added a unit test, but I wasn't able to understand why it didn't run (after reading the travis ci output). It seems like I did the exact same thing as the other unit tests

canada11:
@mpenkov @piskvorky Any help? The output looks a little different than I remember from a couple months ago, but any points can still help = )

mpenkov:
@canada11 Have you tried running the test locally? Does it pass?

smart_open/http.py
```diff
@@ -39,20 +39,27 @@ def open(uri, mode, kerberos=False, user=None, password=None):
         The username for authenticating over HTTP
     password: str, optional
         The password for authenticating over HTTP
+    headers: dict, optional
+        Any headers to send in the request. If none, default headers sent are:
```
piskvorky:
```suggestion
        Any headers to send in the request. If `None`, default headers sent are:
```

smart_open/http.py
```diff
@@ -39,20 +39,27 @@ def open(uri, mode, kerberos=False, user=None, password=None):
...
     Note
     ----
-    If neither kerberos or (user, password) are set, will connect unauthenticated.
+    If neither kerberos or (user, password) are set, will connect
+    unauthenticated, unless set separately in headers.
 
     """
     if mode == 'rb':
-        return BufferedInputBase(uri, mode, kerberos=kerberos, user=user, password=password)
+        return BufferedInputBase(uri, mode, kerberos=kerberos,
```
piskvorky: Hanging indent please (not vertical indent).

smart_open/http.py
piskvorky:
```suggestion
        Any headers to send in the request. If `None`, default headers sent are:
```

smart_open/tests/test_http.py
```diff
@@ -83,3 +83,27 @@ def test_seek_from_end(self):
...
+        # use default again, global shoudn't overwritten from x
+        y = smart_open.http.BufferedInputBase(URL)
+        # should be default headers
+        self.assertEqual(y.headers, {'Accept-Encoding': 'identity'})
+        # should be assigned headers
+        self.assertEqual(x.headers, {'Accept-Encoding': 'compress, gzip', 'Other-Header': 'value'})
+
+    @responses.activate
+    def test_headers(self):
+        """Does the top-level http.open function correctly handle headers?"""
```
piskvorky:
```suggestion
        """Does the top-level http.open function handle headers correctly?"""
```

mpenkov:
OK, this looks good to merge.

@canada11 Thank you for your efforts, and congrats on your first contribution to smart_open :1st_place_medal: Keep them coming!

@interpolatio Thanks for helping us work out the unit test problems.

Issue #271, opened 2019-03-07
smart_open() printing lots of output during transfer when not desired


canada11:
When I use smart_open() (connecting to s3) it prints a ton of stuff like this: 

```
...
133
142
170
154
144
137
147
...
```

Can I turn that off? What is it? There's a *lot* of it, I don't want it to print this.

Also some docs around the project would be greatly appreciated. 
For example another question (because there's no docs):

How do I know if the s3 connection is using multipart upload or uploading serially? Maybe you guys always have it do multipart? I wouldn't know though, all I see is it seems like you have to get some parameters correct for it to use multipart upload

piskvorky:
`smart_open` never prints anything. If you include a reproducible example, we'll have look.

canada11:
(python 3.6.7, boto3 1.9.103, botocore 1.12.103)
```
from smart_open import smart_open
with smart_open('s3://<bucket>/test.csv', 'wb', profile_name='<your-profile>') as s3_landing_file:
    for line in smart_open('https://raw.githubusercontent.com/kavanagh/zipcodes/master/public/data/zcta_county_rel_10.csv'):
        s3_landing_file.write(line)
```

canada11:
While hitting ctrl+c while it's outputting to the screen, I had this output:
```
  File "<stdin>", line 2, in <module>
  File "/<local_path>/lib/python3.6/site-packages/smart_open/http.py", line 82, in read
    logger.debug('read: %r', locals())
  File "/usr/lib/python3.6/logging/__init__.py", line 1286, in debug
    def debug(self, msg, *args, **kwargs):
```

Maybe debug was turned on somewhere in smart_open?

mpenkov:
The output you are seeing is a stack trace. This always gets printed to standard error when you interrupt your Python program by pressing ctrl+c.

What do you see if you don't hit ctrl+c? Please post the full output.

canada11:
I tried to get the output in a file by running a python script and I noticed that there was no output when run as a script. But the output happens when I run the code by hand in a python shell interpreter. I copied the output from the interpreter and it is 44413 lines that *all* are this:
```
>>>s3_landing_file.write(line)
186
159
144
150
150
...
147
148
142
154
>>>
```
I gave the first and last 5 lines of the output and everything in between for 44413 lines all look like that.

So yeah, new update: it seems to be in the python interpreter only?

mpenkov:
That’s really strange.

[...]

PR #274, opened 2019-03-13
updated readme with real-world (doctest) examples

### LINKED ISSUE #144: Document the peculiarity of using rb as the default open mode
By default, smart_open specifies the open mode to be `rb`. This is in contrast to the default `r` in both Py2 and Py3.

smart_open needs to quack like the built-in open. While the above discrepancy is not fatal, it complicates our implementation. For example, in Py3 the following raises an Exception:

```
>>> q = open("/dev/null", 'rb', encoding='utf-8')
ValueError: binary mode doesn't take an encoding argument
```

We cannot do the same, because we cannot tell the difference between:

1. User specified `rb` mode explicitly (as above)
2. User didn't specify a mode, and the default `rb` was set by smart_open

We should document the above somewhere central (like a project wiki) to make it obvious to future developers.

### LINKED ISSUE #258: Document passing boto3 credentials
Need a recipe page or something to include this:

https://github.com/RaRe-Technologies/smart_open/issues/195

mpenkov:
PR contents:

- Document new `open` function
- Instrument examples so that they are runnable with doctest
- Fix #144 
- Fix #258 

We can now test this documentation using:

    python -m doctest README.rst

This takes around a minute to complete because of the long s3_iter_buckets test that downloads several keys. This means we can keep the documentation current and correct.

Unfortunately, we cannot actually run some of the tests because they require elaborate setup (like HDFS, WebHDFS). These tests are disabled, so they do not influence the doctest result.


README.rst
```diff
@@ -12,77 +12,90 @@ smart_open — utils for streaming large files in Python
...
+``smart_open`` is a Python 2 & Python 3 library for **efficient streaming of very large files** from/to S3, HDFS, WebHDFS, HTTP, or local (compressed) files.
+It's a drop-in replacement for Python's built-in ``open()``: it can do anything ``open`` can (100% compatible, falls back to native ``open`` wherever possible), plus lots of nifty extra stuff on top.
 
 ``smart_open`` is well-tested, well-documented, and has a simple, Pythonic API:
 
 .. code-block:: python
 
-  >>> from smart_open import smart_open
-
+  >>> from smart_open import open
```
piskvorky: -1: what's the reason for this renaming? Both confusing (shadows a built-in) and backward incompatible.

piskvorky:
@mpenkov can you add motivation to the PR description?

Testing is good, is this the right place for it? I'm not keen on making our README less readable (`# doctest: +SKIP` etc).

The README is there to convey the mission and usage examples of `smart_open`, not be bogged down by some internal developer minutia (irrelevant to users).

README.rst
mpenkov: > -1: what's the reason for this renaming?

It's not really a renaming. It's a new function. The old function still works exactly the same as before, we just don't want people to use it anymore (for reasons, see below).

> Both confusing (shadows a built-in)

I don't think it's any more confusing than, for example, io.open or any of the other similarly named functions in the [standard library](https://docs.python.org/3/search.html?q=open&check_keywords=yes&area=default). There are more than a dozen such cases, so I think the precedent for using the name `open` is strong. 

Do you recommend to do this instead:

```python
import smart_open
fin = smart_open.open(...)
```

>>> and backward incompatible.

Yes, it's backwards incompatible with the old smart_open.smart_open, but that function several big disadvantages:

- in was incompatible the built-in open function, contradicting the main mission of the project. For example, it opened things as binary by default.
- It accepted keyword arguments without checking them (via **kwargs). This made the parameters difficult to document.

In summary, the new function solves the above two problems, so it's **incompatible by design**. The old function still works the same way, so why would the incompatibility be a problem?

In summary, this new function is a **true** replacement for the built-in `open` function. It accepts the same parameters that the built-in `open` does. It also mirrors alternative entry points like `io.open` and over a dozen of other `*.open`, so I think calling it `open` it natural.

vmarkovtsev: I really hope that the users will not do `%pylab`-style star-importing everything, as it will be yet another gotcha with porting Jupyter notebooks to projects, and debugging in general.

mpenkov: @vmarkovtsev I don't see the merit of your argument, for three reasons.

First, think about this: what could go wrong (TM)? The smart_open.open function behaves **identically** to the built-in open - it even falls back to the built-in open whenever possible. In the vast majority of cases, it will give **exactly the same result**.

Second, there really is no practical benefit to `from smart_open import *`, given there are only two public functions in there (the other one being s3_iter_buckets).  

Finally, star-importing everything is [generally discouraged](https://stackoverflow.com/questions/2386714/why-is-import-bad), anyway. Yeah, people still do it, but the consequences are on them.

mpenkov:
@piskvorky I got rid of the instrumentation, so the README is now both doctest-able and readable.

README.rst
vmarkovtsev: This is the thing, many people still do it. I am not for or against `open` btw, just writing that this will definitely cause problems in some workflows. Ours, for example.

Question: if one wants to read a gzip file verbatim, and `open` is already overwritten, they will have to use `io.open`, correct?

mpenkov: Opening a gzip file in binary mode with smart_open.open will give the same results as io.open. This is one of the cases where smart_open falls back to io.open. I expect this is the use case you're referring to.

Opening a gzip file in text mode *will* give different results. smart_open will decode the file transparently. io.open will give you a UnicodeError trying to decode the bytes into meaningful text. It's unlikely your workflow does this, as the UnicodeError is not recoverable at that stage.

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
+  >>> for line in open('http://example.com/index.html'):
+  ...     print(line, end='')
+  ...     break
+  <!doctype html>
 
-  >>> # stream from WebHDFS
-  >>> for line in smart_open('webhdfs://host:port/user/hadoop/my_file.txt'):
-  ...     print(line)
+  >>> # stream from HDFS
+  >> for line in open('hdfs://user/hadoop/my_file.txt', encoding='utf8'):
```
piskvorky: Is it normal to do `>>` (two symbols instead of three)? I never saw that before. It mismatches indent with other `>>>` lines (looks un-Pythonic).

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
+  >>> tkwa = dict(session=boto3.Session(profile_name='smart_open'))
+  >>> url = 's3://smart-open-py37-benchmark-results/test.txt'
+  >>> lines = [b'first line\n', b'second line\n', b'third line\n']
+  >>> [len(l) for l in lines]
+  [11, 12, 11]
+  >>>
+  >>> with open(url, 'wb', tkwa=tkwa) as fout:
+  ...    for line in lines:
+  ...         fout.write(line)
+  11
```
piskvorky: Do we really need this output in the README? It looks more confusing than illuminating to me.

smart_open/webhdfs.py
```diff
@@ -1,3 +1,16 @@
...
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
+#
+# This code is distributed under the terms and conditions from the MIT License (MIT).
+#
+
+"""Implements reading and writing to/from WebHDFS.
+
+The main entry point is the :func:`~smart_open.webhdfs.open` function.
+
+"""
```
piskvorky: Blank line after docstring.

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
 
+  >>> gz_path = 'smart_open/tests/test_data/crime-and-punishment.txt.gz'
+  >>>
   >>> # stream from/to compressed files, with transparent (de)compression:
-  >>> for line in smart_open('./foo.txt.gz', encoding='utf8'):
-  ...    print(line)
+  >>> for line in open(gz_path, encoding='utf-8'):
+  ...    print(line[:50])
+  ...    break
+  В начале июля, в чрезвычайно жаркое время, под веч
```
piskvorky: Why does this show Russian? Above says "Crime and Punishment"?

README is not a place for finesse. I'd prefer the examples to be straightforward and clear and unambiguous, erring on the side of bluntness rather than subtlety.

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
+  ...         break
   ...     fin.seek(0)  # seek to the beginning
-  ...     b1000 = fin.read(1000)  # read 1000 bytes
-
-  >>> # stream from HDFS
-  >>> for line in smart_open('hdfs://user/hadoop/my_file.txt', encoding='utf8'):
-  ...     print(line)
+  ...     print(fin.read(4))
+  User-Agent: *
+  0
```
piskvorky: Where does this `0` come from? I see only two `print`s?

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
-  >>> with smart_open('s3://mybucket/mykey.txt', 'wb') as fout:
-  ...     for line in [b'first line\n', b'second line\n', b'third line\n']:
-  ...          fout.write(line)
+  >>> # stream from WebHDFS
+  >> for line in open('webhdfs://host:port/user/hadoop/my_file.txt'):
+  ..     print(line)
+
+  >>> # stream content *into* S3 (write mode), using a custom Session for authentication:
+  >>> import boto3
+  >>> tkwa = dict(session=boto3.Session(profile_name='smart_open'))
```
piskvorky: Why `tkwa`? What does that stand for?

piskvorky: Also, this example seems overly complicated. In the intro, we want the simplest (and most readable) version, with simple `smart_open(`wb`). No `tkwa` (?).

README.rst
mpenkov: It's the return value of the seek method

README.rst
mpenkov: tkwa stands for Transport KeyWord Arguments

This code actually writes to our S3 bucket, so it needs some sort of authentication. We can do authentication in other (hidden) ways for this example, e.g. using environment variables. We can move the example with tkwa further down the list, as it demonstrates how to pass custom parameters for more advanced use cases.

README.rst
mpenkov: OK, I will use a simpler example gzip file with English text.

README.rst
mpenkov: The two symbol thing comes from here: https://stackoverflow.com/questions/1809037/python-doctest-skip-entire-block

It's a hack that forces doctest to ignore that line.

README.rst
mpenkov: Well, we can hide it by doing a variable assignment, e.g. `bytes_written = fout.write(line)`.

The reason that output is there is because it's what the library outputs in real life if a user entered that code in a REPL. That's the beauty of the tests: it's the real deal, not an artist's impression of what the output may look like.

README.rst
```diff
@@ -12,77 +12,105 @@ smart_open — utils for streaming large files in Python
...
+  >>>
+  >>> with open(url, 'wb', tkwa=tkwa) as fout:
+  ...    for line in lines:
+  ...         fout.write(line)
+  11
+  12
+  11
+  >>>
+  >>> with open(url, tkwa=tkwa) as fin:
+  ...    print(fin.read(), end='')
```
piskvorky: Since this is not binary mode, what encoding does this use? Let's be explicit.

README.rst
piskvorky: The value is ignored AFAICS, and should not be printed. What is doing this printing? I do not see any such output when I run the code (as I don't think I should).

README.rst
piskvorky: Yes, move further down please. Let's keep simple things simple (`smart_open`s main allure).

piskvorky: > tkwa stands for Transport KeyWord Arguments

How about `transport_kw`?  `tkwa` sounds a little too magical to me.

README.rst
piskvorky: OK, thanks. Is the code block is copy-pastable though (ipython)?

README.rst
piskvorky: This is not what I see in REPL though. What REPL are you using? It sounds bizarre that a REPL would print (??) unassigned return values.

mpenkov: This is pretty common: we [recently](https://github.com/RaRe-Technologies/smart_open/issues/271) had a user ask about this particular behavior. So, I think it's something we should document in our examples.

To answer your question directly, it depends on your REPL:

```
(smart_open) misha@cabron:~$ ipython
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import random

In [2]: for _ in range(3):
   ...:     random.randint(1, 10)

In [3]:                                                                                                                                                                                                       
Do you really want to exit ([y]/n)? y
(smart_open) misha@cabron:~$ python
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import random
>>> for _ in range(3):
...     random.randint(1, 10)
... 
3
3
1
>>> 
```

So ipython suppresses that output, but the standard python REPL does not.

README.rst
mpenkov: Unfortunately, no, it isn't pasteable.

If you'd like the examples to be pasteable as-is _and_ bypass tests, then we could provide code listings instead of REPL extracts. For example, instead of:

```python
>>> # read a file via WebHDFS
>>> for line in smart_open('webhdfs://host:port/user/hadoop/my_file.txt'):
...     print(line)
```

we would show:

```python
# read a file via WebHDFS
for line in smart_open('webhdfs://host:port/user/hadoop/my_file.txt'):
    print(line)
```

Is that an acceptable compromise?

README.rst
mpenkov: Again, it's the REPL doing the printing.

README.rst
mpenkov: OK, I will use transport_kwargs (for consistency with X_kwargs used elsewhere in the library).

mpenkov: How about t_params? It's less magical and shorter than transport_kw.

README.rst
piskvorky: Yes, looks fine to me.

README.rst
piskvorky: Looks almost like a REPL bug to me! (printing stuff that noone asked it to print… WTH).

Either way, I'm strongly -1 on including such output in our examples or README. Very confusing, even to me.

README.rst
piskvorky: `t_` too magical again :-) I like names to be indicative of their semantic purpose (rather than syntax type, or optimizing for short length).

`transport_kwargs`? `transport_params`?

README.rst
mpenkov: OK, I've suppressed printing via variable assignment.

README.rst
mpenkov: OK, I think transport_params is the way to go.

We may eventually want to add compressor_params later, so params by itself would be too ambiguous.

mpenkov:
@piskvorky I'd like to merge this soon, and do a smart_open release as a practice run for the upcoming gensim release. Please let me know in the next 24 hours if that doesn't sound right, otherwise, I will proceed.

README.rst
piskvorky: As long as normal `from smart_open import smart_open` continues to work, it's probably OK (too much existing code relies on it).

README.rst
```diff
@@ -12,77 +12,83 @@ smart_open — utils for streaming large files in Python
...
+    file:///home/user/file.bz2
+    [ssh|scp|sftp]://username@host//path/file
+    [ssh|scp|sftp]://username@host/path/file
+    file:///home/user/file.xz
+
+For detailed API info, see the online help:
+
+.. code-block:: python
+
+    help('smart_open')
```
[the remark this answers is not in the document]
mpenkov: Are you on the branch of this PR? If not, then the command won't show you anything useful, because the previous way of exposing the package's functionality was broken.

This is part of what you would see if you have this branch checked out:

```
$ python -c 'help("smart_open")' | head -n 40
Help on package smart_open:

NAME
    smart_open

DESCRIPTION
    Utilities for streaming to/from several file-like data storages: S3 / HDFS / local
    filesystem / compressed files, and many more, using a simple, Pythonic API.
    
    The streaming makes heavy use of generators and pipes, to avoid loading
    full file contents into memory, allowing work with arbitrarily large files.
    
    The main functions are:
    
    * `open()`, which opens the given file for reading/writing
    * `s3_iter_bucket()`, which goes over all keys in an S3 bucket in parallel
    * `register_compressor()`, which registers callbacks for transparent compressor handling

PACKAGE CONTENTS
    doctools
    hdfs
    http
    s3
    smart_open_lib
    ssh
    tests (package)
    webhdfs

FUNCTIONS
    open(uri, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None, ignore_ext=False, transport_params={})
        Open the URI object, returning a file-like object.
        
            The URI is usually a string in a variety of formats:
        
            1. a URI for the local filesystem: `./lines.txt`, `/home/joe/lines.txt.gz`, `file:///home/joe/lines.txt.bz2`
            2. a URI for HDFS: `hdfs:///some/path/lines.txt`
            3. a URI for Amazon's S3 (can also supply credentials inside the URI):
               `s3://my_bucket/lines.txt`, `s3://my_aws_key_id:key_secret@my_bucket/lines.txt`
        
            The URI may also be one of:
```

README.rst
```diff
@@ -12,77 +12,83 @@ smart_open — utils for streaming large files in Python
...
 
-  >>> from smart_open import smart_open
-
+  >>> from smart_open import open
+  >>>
   >>> # stream lines from an S3 object
-  >>> for line in smart_open('s3://mybucket/mykey.txt', 'rb'):
-  ...    print(line.decode('utf8'))
+  >>> for line in open('s3://commoncrawl/robots.txt'):
+  ...    print(line, end='')
```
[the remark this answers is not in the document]
mpenkov: `line` already ends with a newline. If you `print(line)`, then you end up with twice as many newlines as you really need.

README.rst
```diff
@@ -12,77 +12,83 @@ smart_open — utils for streaming large files in Python
...
+  ...    break
+  User-Agent: *
 
+  >>> gz_path = 'smart_open/tests/test_data/1984.txt.gz'
+  >>>
   >>> # stream from/to compressed files, with transparent (de)compression:
-  >>> for line in smart_open('./foo.txt.gz', encoding='utf8'):
-  ...    print(line)
+  >>> for line in open(gz_path, encoding='utf-8'):
+  ...    print(line, end='')
```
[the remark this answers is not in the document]
mpenkov: It's a single-line file.

README.rst
```diff
@@ -12,77 +12,83 @@ smart_open — utils for streaming large files in Python
 What?
 =====
 
-``smart_open`` is a Python 2 & Python 3 library for **efficient streaming of very large files** from/to S3, HDFS, WebHDFS, HTTP, or local (compressed) files. It's a drop-in replacement for Python's built-in ``open()``: it can do anything ``open`` can (100% compatible, falls back to native ``open`` wherever possible), plus lots of nifty extra stuff on top.
+``smart_open`` is a Python 2 & Python 3 library for **efficient streaming of very large files** from/to S3, HDFS, WebHDFS, HTTP, or local storage.
+It supports transparent, on-the-fly (de-)compression for a variety of different formats.
```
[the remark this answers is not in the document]
mpenkov: I'm not sure what actually changed here. This suggestion would replace:

```
It supports transparent, on-the-fly (de-)compression for a variety of different formats.
```

with

```
``smart_open`` is a Python 2 & Python 3 library for **efficient streaming of very large files** from/to S3, HDFS, WebHDFS, HTTP, or local storage. It supports transparent, on-the-fly (de-)compression for a variety of different formats.
```

which would duplicate the opening sentence.

What would you like me to change here?

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ This new function replaces `smart_open.smart_open`, which is now deprecated.
 Main differences:
 
 - ignore_extension → ignore_ext
-- new `tkwa` dict parameter to contain keyword parameters for the transport layer (S3, HTTPS, HDFS, etc). 
+- new `t_params` dict parameter to contain keyword parameters for the transport layer (S3, HTTPS, HDFS, etc).
```
mpenkov:
```suggestion
- new `transport_params` dict parameter to contain keyword parameters for the transport layer (S3, HTTPS, HDFS, etc).
```

README.rst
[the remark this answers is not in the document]
piskvorky: No, I was on `master`. Thanks!

README.rst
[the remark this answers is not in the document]
piskvorky: I'd like to use the suggested text instead of the original. Sorry, I don't know how to do multi-line suggestions directly on Github.

README.rst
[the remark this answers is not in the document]
piskvorky: Clever! Although clever is not what we're going for in a README, plus will it work in Python 2?

How about printing `repr` instead so the control characters don't render directly?

README.rst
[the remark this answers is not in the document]
piskvorky: :)

README.rst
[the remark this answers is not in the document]
mpenkov: Yeah, multiline suggestions are tricky (I'm not sure you can actually do them). I've applied the changes I _think_ you wanted to make. If I got it wrong, please feel free to edit the README.rst directly: as a maintainer, you're able to edit stuff on the PR.

README.rst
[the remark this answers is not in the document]
mpenkov: Added more lines

README.rst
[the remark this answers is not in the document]
mpenkov: Added repr

README.rst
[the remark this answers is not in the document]
piskvorky: You're right, nice!

Done; LGTM.

PR #277, opened 2019-03-29
More documentation fixes.

mpenkov:
- Use raw strings to escape newlines in examples
- Read examples from README.rst instead of copy-pasting them
- Add help.txt to let people read the full module docstring without
  downloading and installing smart_open

### DIFF
--- a/smart_open/doctools.py
+++ b/smart_open/doctools.py
@@ -13,7 +13,7 @@ For internal use only.
 
 import inspect
 import io
-import os.path as P
+import os.path
 import re
 
 
@@ -142,8 +142,8 @@ def extract_examples_from_readme_rst(indent='    '):
     -----
     Quite fragile, depends on named labels inside the README.rst file.
     """
-    curr_dir = P.dirname(P.abspath(__file__))
-    readme_path = P.join(curr_dir, '../README.rst')
+    curr_dir = os.path.dirname(os.path.abspath(__file__))
+    readme_path = os.path.join(curr_dir, '..', 'README.rst')
     try:
         with open(readme_path) as fin:
             lines = list(fin)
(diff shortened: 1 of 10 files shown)


smart_open/doctools.py
```diff
@@ -13,6 +13,8 @@
 
 import inspect
 import io
+import os.path as P
```
piskvorky: -1 on import aliases. Please use the full name – especially with built-ins.

smart_open/doctools.py
```diff
@@ -121,3 +123,33 @@ def to_docstring(kwargs, lpad=''):
...
+    -------
+    str
+        The examples.
+
+    Notes
+    -----
+    Quite fragile, depends on named labels inside the README.rst file.
+    """
+    curr_dir = P.dirname(P.abspath(__file__))
+    readme_path = P.join(curr_dir, '../README.rst')
```
piskvorky:
```suggestion
    readme_path = P.join(curr_dir, '..', 'README.rst')
```

smart_open/s3.py
```diff
@@ -77,16 +77,17 @@ def open(
     key_id: str
         The name of the key within the bucket.
     mode: str
-        The mode with which to open the object.  Must be either rb or wb.
+        The mode for opening the object.  Must be either rb or wb.
```
piskvorky:
```suggestion
        The mode for opening the object.  Must be either "rb" or "wb".
```

smart_open/smart_open_lib.py
```diff
@@ -57,6 +57,7 @@
 from smart_open import doctools
 
 logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
```
piskvorky: Better add NullHandler in `__init__.py`, to avoid having to add it to each logger individually?

smart_open/doctools.py
mpenkov: OK, done

smart_open/doctools.py
mpenkov: OK, done

PR #279, opened 2019-04-04
Now allows passing an rsakey to ssh:// open as a transport_param

kevin-thorpe:
We are intending to use this in ephemeral Airflow workers. As such they don't have installed SSH keys but obtain them as part of the job configuration. This patch simply allows the passing of a PEM format private key as a transport parameter when opening files over ssh. 

This change should not break any existing functionality. It is purely optional.


smart_open/ssh.py
```diff
@@ -43,18 +44,22 @@
 DEFAULT_PORT = 22
 
 
-def _connect(hostname, username, port):
+def _connect(hostname, username, port, rsakey = None):
```
mpenkov: Get rid of spaces around keyword parameter, here and everywhere.

```suggestion
def _connect(hostname, username, port, rsakey=None):
```

smart_open/ssh.py
```diff
@@ -43,18 +44,22 @@
...
     key = (hostname, username)
     ssh = _SSH.get(key)
     if ssh is None:
         ssh = _SSH[key] = paramiko.client.SSHClient()
         ssh.load_system_host_keys()
         ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        ssh.connect(hostname, port, username)
+        if rsakey:
```
mpenkov: I think the variable naming here is a bit confusing. How is rsakey different from sshkey and pkey?

Also, is there a benefit to passing rsakey instead of pkey?

Furthermore, I think instead of passing just the key, we should pass connect_kwargs as dictionary and then pass them all to the connect function. That will allow us to support other parameters on top of pkey (there are many: https://docs.paramiko.org/en/2.4/api/client.html#paramiko.client.SSHClient.connect)

smart_open/ssh.py
```diff
@@ -43,18 +44,22 @@
...
+            sshkey = paramiko.RSAKey.from_private_key(StringIO(rsakey))
+            ssh.connect(hostname, port, username, pkey = sshkey)
+        else:
+            ssh.connect(hostname, port, username)
     return ssh
 
 
-def open(path, mode='r', host=None, user=None, port=DEFAULT_PORT):
+def open(path, mode='r', host=None, user=None, port=DEFAULT_PORT, transport_params={}):
```
mpenkov: Please have a look at how other submodules handle transport_params.

mpenkov:
I think we will merge #292 in favor of this PR, it is more complete

smart_open/ssh.py
luckydonald: In the light of https://github.com/RaRe-Technologies/smart_open/issues/304 `transport_params={}` most certainly should be replaced with 
```py
def open(..., transport_params=None):
if transport_params is None:
     transport_params = {}
# end if
```

PR #281, opened 2019-04-09
Makes xz decompression an optional dependency in python<3.3

tdhopper:
xz decompression in Python<3.3 requires liblzma which won't be available on every machine. Move xz decompression option to optional dependency (`pip install smart_open[xz]`) instead of having install fail when liblzma isn't available.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,7 +77,6 @@ Other examples of URLs that ``smart_open`` accepts::
     file:///home/user/file.bz2
     [ssh|scp|sftp]://username@host//path/file
     [ssh|scp|sftp]://username@host/path/file
-    file:///home/user/file.xz
 
 .. _doctools_after_examples:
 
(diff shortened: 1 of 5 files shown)


tdhopper:
Currently travis doesn't test optional dependency. Happy to figure out how to add this if this looks promising.

mpenkov:
@tdhopper Thank you for the pull request. It works as advertised, but I don't like the extra effort we're applying to integrate .xz support into smart_open. In particular, voodoo around import statements and Python versions, as well as explaining things like `pip install smart_open[xz]` seem like a step backwards to me.

Are you able to move the xz support from the library to the README? It's basically a copy-paste from [here](https://github.com/RaRe-Technologies/smart_open/blob/ce3d8acf7810358c93cbe11375dc48485901ccdf/smart_open/smart_open_lib.py#L110) to [here](https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst#supported-compression-formats). You can replace the existing LZMA example in the README with the working cope from smart_open_lib.py.

You're a first-time contributor, so if it's too much, let me know and I'll deal with it on the weekend.

tdhopper:
@mpenkov I'm willing. It seemed odd to me to rip out functionality that's already been released, but that's not my call.

mpenkov:
For the record, here are my reasons for moving .xz support from code to examples:

- Adding support for .xz (or .lzma or any other compression format) is trivial (it's like 5 lines of code).
- Integration-like stuff like this is better handled [outside the library](https://www.brainyquote.com/quotes/antoine_de_saintexupery_103610). Supporting every compression format out there *well* out-of-the-box isn't our goal. We make exceptions for gzip and bz2, because they are so common.
- The existing LZMA support is only partial. The lzma library has [options](https://docs.python.org/3/library/lzma.html#compressing-and-decompressing-data-in-memory) that the current implementation plain ignores.
- Finally, the LZMA stuff was introduced in a recent minor release, it's not too late to quietly find a better place for it

So, please go ahead.

When adding the example to the README, please check that it passes doctests. We already have test .xz data in the repo, so please use that.

tdhopper:
@mpenkov Do you have examples doctests? I don't see one for [this](https://github.com/RaRe-Technologies/smart_open/blob/ce3d8acf7810358c93cbe11375dc48485901ccdf/README.rst#supported-compression-formats).

mpenkov:
Yeah, that snippet is untested, because it's a hand-wavy example (as opposed to a real example, see above in that same file).

The actual test is [here](https://github.com/RaRe-Technologies/smart_open/blob/ce3d8acf7810358c93cbe11375dc48485901ccdf/smart_open/tests/test_smart_open.py#L1021). You should be able to pretty much copy-paste it into the README.

PR #284, opened 2019-04-10
Improve S3 read performance by not copying buffer

aperiodic:
## Add ByteBuffer Class

This class encapsulates the buffer functionality implemented in
smart_open.s3.BufferedInputBase and in
smart_open.http.BufferedInputBase as a standalone class. It uses the
read implementation introduced to the S3 BufferedInputBase in the
previous commit, to minimize copying of the buffer.

Because the use case for these ByteBuffers is to amortize the network
transfer overhead costs of reading data from S3 or via HTTP, it's better
to read a consistent number of bytes when filling up the buffer to
consistently amortize that overhead cost, rather than filling the buffer
up to some a priori capacity. To that end, the ByteBuffer constructor
takes a `chunk_size` that sets the number of bytes to read when filling
the buffer, rather than a fixed buffer capacity (which could easily
lead to reading a different number of bytes from the reader if the
caller ever filled the buffer when it was not empty). However, the
caller can still ask for fewer than `chunk_size` bytes to be read during
filling using the fill method's `size` argument.

Change the BufferedInputBase class (and descendents) in both
smart_open.s3 and smart_open.http to use instances of ByteBuffer for
their read buffers, instead of bytestrings. This simplifies their
implementation, while still avoiding copying on reads to maintain the
94% performance improvement for many small reads on an S3 file
introduced in the first version of this PR.

### Initial, Now Discarded Implementation

Previously, the buffer of an S3 file opened in 'rb' mode was split into
two copies whenever its `read()` method was called: one copy was
returned to the caller, and the other copy became the new buffer. This
lead to bad performance when performing many small reads with the buffer
set to a large enough size to amortize the cost of transferring data
from S3.

In our use case, we set the buffer to 4 MiB when reading a 16.5 MiB file
containing 694k serialized protobuf messages, each prefixed by their
byte length. This required calling the S3 file's `read()` method nearly
1.4 million times, which in turn meant that the up-to-4MiB buffer needed
to be copied and garbage collected the same number of times. All of that
copying and garbage collecting alone took 152 seconds, for a total of
179 seconds to download and parse the file.

Eliminate the copying on each `read()` by adding an instance attribute
`_buffer_pos` to smart_open.s3's BufferedInputBase and its descendents,
in order to track the current position within the buffer. Change
`read()` to use `_buffer_pos` to create the slice of the buffer to
return to the caller, and then move `_buffer_pos` by the size of that
slice.

Add two new internal instance methods to BufferedInputBase to accomodate
the fact that the buffer now contains already-read data. The first,
`_len_remaining_buffer()`, calculates how many unread bytes remain in
the buffer (previously this was simply the length of the buffer); this
is now used everywhere that `len(self._buffer)` was previously used in
BufferedInputBase. The second method, `_remaining_buffer()`, returns
a copy of the unread portion of the buffer; this is used in
BufferedInputBase's `readline()` method to find the next unread newline.

With these changes, that same 16.5 MiB protobuf file with 694k messages
can be downloaded and parsed in 10 seconds, a 94% improvement in
performance!

### DIFF
--- a/smart_open/bytebuffer.py
+++ b/smart_open/bytebuffer.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+"""Implements a byte buffer that allows callers to read data with minimal
+copying, and has a fast __len__ method. The buffer is parametrized by its
+chunk_size, which is the number of bytes that it will read in from the supplied
+reader or iterable when the buffer is being filled. As primary use case for this
+buffer is to amortize the overhead costs of transferring data over the network
+(rather than capping memory consumption), it leads to more predictable
+performance to always read the same amount of bytes each time the buffer is
+filled, hence the chunk_size parameter instead of some fixed capacity.
+
+The bytes are stored in a bytestring, and previously-read bytes are freed when
+the buffer is next filled (by slicing the bytestring into a smaller copy)."""
+
+class ByteBuffer(object):
+    def __init__(self, chunk_size):
+        """The chunk_size indicates the number of bytes that will be read from
+        the supplied reader when filling the buffer. The buffer has no maximum
+        size."""
+        self._chunk_size = chunk_size
+        self.empty()
+
+    def __len__(self):
+        """Return the number of unread bytes in the buffer"""
+        return len(self._bytes) - self._pos
+
+    def read(self, size = -1):
+        """Read up to size bytes from the buffer while advancing the read
+        position, and then return the bytes. If size is negative, return all
+        unread bytes."""
+        part = self.peek(size)
+        self._pos += len(part)
+        return part
+
+    def peek(self, size = -1):
+        """Return up to the next size bytes from the buffer, without advancing
+        the read position. If size is negative, return all unread bytes."""
+        if size < 0 or size > len(self):
+            size = len(self)
+
+        part = self._bytes[self._pos : self._pos + size]
+        return part
+
+    def empty(self):
+        """Remove all bytes from the buffer"""
+        self._bytes = b''
+        self._pos = 0
+
+    def fill(self, reader_or_iterable, size = -1):
+        """Fill the buffer with bytes from `reader_or_iterable` until one of
+        these conditions is met:
+            * size bytes have been read from reader_or_iterable (if size >= 0);
+            * chunk_size bytes have been read from reader_or_iterable;
+            * no more bytes can be read from reader_or_iterable;
+        Return number of bytes added to the buffer.
+        Note: all previously-read bytes in the buffer are removed."""
+        size = size if size >= 0 else self._chunk_size
+        size = min(size, self._chunk_size)
+
+        if self._pos != 0:
+            self._bytes = self._bytes[self._pos:]
+            self._pos = 0
+
+        if hasattr(reader_or_iterable, 'read'):
+            new_bytes = reader_or_iterable.read(size)
+        else:
+            new_bytes = b''
+            while len(new_bytes) < size:
+                try:
+                    new_bytes += next(reader_or_iterable)
+                except StopIteration:
+                    break
+
+        self._bytes += new_bytes
+        return len(new_bytes)
--- a/smart_open/bytebuffer.py
+++ b/smart_open/bytebuffer.py
@@ -17,7 +17,35 @@ class ByteBuffer(object):
 
     The bytes are stored in a bytestring, and previously-read bytes are freed
     when the buffer is next filled (by slicing the bytestring into a smaller
-    copy)."""
+    copy).
+
+    Example
+    -------
+
+    Note that while this example works in both Python 2 and 3, the doctest only
+    passes in Python 3 due to the bytestring literals in the expected values.
+
+    >>> buf = ByteBuffer(chunk_size = 8)
+    >>> message_bytes = iter([b'Hello, W', b'orld!'])
+    >>> buf.fill(message_bytes)
+    8
+    >>> len(buf) # only chunk_size bytes are filled
+    8
+    >>> buf.peek()
+    b'Hello, W'
+    >>> len(buf) # peek() does not change read position
+    8
+    >>> buf.read(6)
+    b'Hello,'
+    >>> len(buf) # read() does change read position
+    2
+    >>> buf.fill(message_bytes)
+    5
+    >>> buf.read()
+    b' World!'
+    >>> len(buf)
+    0
+    """
 
     def __init__(self, chunk_size=io.DEFAULT_BUFFER_SIZE):
         """Create a ByteBuffer instance that reads chunk_size bytes when filled.
--- a/smart_open/bytebuffer.py
+++ b/smart_open/bytebuffer.py
@@ -79,28 +79,29 @@ class ByteBuffer(object):
         self._bytes = b''
         self._pos = 0
 
-    def fill(self, reader_or_iterable, size = -1):
-        """Fill the buffer with bytes from reader_or_iterable until one of these
+    def fill(self, source, size = -1):
+        """Fill the buffer with bytes from source until one of these
         conditions is met:
-            * size bytes have been read from reader_or_iterable (if size >= 0);
-            * chunk_size bytes have been read from reader_or_iterable;
-            * no more bytes can be read from reader_or_iterable;
+            * size bytes have been read from source (if size >= 0);
+            * chunk_size bytes have been read from source;
+            * no more bytes can be read from source;
         Returns the number of new bytes added to the buffer.
         Note: all previously-read bytes in the buffer are removed.
 
         Parameters
         ----------
-        reader_or_iterable: a file-like object, or iterable that contains bytes
+        source: a file-like object, or iterable/list that contains bytes
             The source of bytes to fill the buffer with. If this argument has
             the `read` attribute, it's assumed to be a file-like object and
             `read` is called to get the bytes; otherwise it's assumed to be an
-            iterable that contains bytes, and `next` is used to get them.
+            iterable or list that contains bytes, and a for loop is used to get
+            the bytes.
         size: int, optional
-            The number of bytes to try to read from reader_or_iterable. If not
-            supplied, negative, or larger than the buffer's chunk_size, then
-            chunk_size bytes are read. Note that if reader_or_iterable is an
-            iterable, then it's possible that more bytes will be read if the
-            iterable produces more than one byte on each `next` call.
+            The number of bytes to try to read from source. If not supplied,
+            negative, or larger than the buffer's chunk_size, then chunk_size
+            bytes are read. Note that if source is an iterable or list, then
+            it's possible that more than size bytes will be read if iterating
+            over source produces more than one byte at a time.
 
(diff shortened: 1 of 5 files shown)


smart_open/s3.py
```diff
@@ -190,6 +190,7 @@ def __init__(self, bucket, key, buffer_size=DEFAULT_BUFFER_SIZE,
         self._content_length = self._object.content_length
         self._current_pos = 0
         self._buffer = b''
+        self._buffer_pos = 0
```
mpenkov: Thank you for this contribution. I can't help feeling that using a bytestring for _buffer is sub-optimal here. We have to add functionality to output the current position in the buffer, as well as the remainder of the buffer.

Is there a reason why we can't use io.BytesIO? That datatype would encapsulate the above functionality, instead of adding it to the long list of responsibilities that the class already has. More specifically:

- instead of slicing the buffer, we can do _buffer.read(num_bytes)
- _len_remaining_buffer becomes len(_buffer)
- _remaining_buffer becomes _buffer.getvalue()

What do you think?

aperiodic: I don't see how that would work, given that io.BytesIO instance's don't have a length.
```py
In [1]: import io

In [2]: buf = io.BytesIO()

In [3]: buf.write(b'foo bar baz')
Out[3]: 11

In [4]: len(buf)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-c8926a5dfb43> in <module>
----> 1 len(buf)

TypeError: object of type '_io.BytesIO' has no len()
```
The only way I can think of calculating the length of what's in a BytesIO buffer is to read everything out of the buffer as a bytestring, which would seem to put us right back where we were before this PR.

aperiodic: I do agree with you that it's not ideal for the BufferedInputBase class to be responsible for this. What do you think about creating a ByteStringBuffer class to encapsulate this functionality? That would simplify the implementation of this module's BufferedInputBase, and could also be used in smart_open.http's BufferedInputBase to realize the same performance improvement for files read over HTTP.

aperiodic: I do agree with you that it's not ideal for the BufferedInputBase class to be responsible for this. What do you think about creating a ByteStringBuffer class to encapsulate this functionality? That would simplify the implementation of this module's BufferedInputBase, and could also be used in smart_open.http's BufferedInputBase to realize the same performance improvement for files read over HTTP.

If you do like this idea, do you have any opinion on where such a class should be defined? I could put it in the S3 module, but personally I think it makes a little more sense to create a module  just for this class, so it can be used wherever it's needed with less chance of creating an import loop.

mpenkov: I like the idea of a class to encapsulate the new responsibility. Let's call it ByteBuffer and put it in bytebuffer.py.

Can you take care of it in this PR?

aperiodic: Sounds good. I'll do that, write a few tests for it, and take a shot at making `smart_open.http`s BufferedInputBase class use it for its `_read_buffer`, too.

Should I keep using this PR, or make a new one? If sticking with this PR, do you prefer adding the commits with the new implementation on top of this PR, or force-pushing to replace the current commit? It's all the same to me.

mpenkov: Please continue using this PR and add more commits. It'd be good to avoid force-pushing, but it isn't critical.

smart_open/bytebuffer.py
```diff
@@ -0,0 +1,74 @@
...
+        the supplied reader when filling the buffer. The buffer has no maximum
+        size."""
+        self._chunk_size = chunk_size
+        self.empty()
+
+    def __len__(self):
+        """Return the number of unread bytes in the buffer"""
+        return len(self._bytes) - self._pos
+
+    def read(self, size = -1):
```
mpenkov: [PEP8](https://www.python.org/dev/peps/pep-0008/), here and everywhere.

Please run flake8 on your code and fix any errors that it reports.

```suggestion
    def read(self, size=-1):
```

smart_open/bytebuffer.py
```diff
@@ -0,0 +1,74 @@
...
+buffer is to amortize the overhead costs of transferring data over the network
+(rather than capping memory consumption), it leads to more predictable
+performance to always read the same amount of bytes each time the buffer is
+filled, hence the chunk_size parameter instead of some fixed capacity.
+
+The bytes are stored in a bytestring, and previously-read bytes are freed when
+the buffer is next filled (by slicing the bytestring into a smaller copy)."""
+
+class ByteBuffer(object):
+    def __init__(self, chunk_size):
```
mpenkov: A sensible default value is helpful here:

```suggestion
    def __init__(self, chunk_size=io.DEFAULT_BUFFER_SIZE):
```

smart_open/bytebuffer.py
```diff
@@ -0,0 +1,74 @@
...
+reader or iterable when the buffer is being filled. As primary use case for this
+buffer is to amortize the overhead costs of transferring data over the network
+(rather than capping memory consumption), it leads to more predictable
+performance to always read the same amount of bytes each time the buffer is
+filled, hence the chunk_size parameter instead of some fixed capacity.
+
+The bytes are stored in a bytestring, and previously-read bytes are freed when
+the buffer is next filled (by slicing the bytestring into a smaller copy)."""
+
+class ByteBuffer(object):
```
mpenkov: Please add a docstring for this class. I think most of the above text (module docstring) is really the class' docstring.

Are you familiar with doctest? If you are, please add some code examples of using your new class to the class or module docstring, e.g.

```python
>>> buf = ByteBuffer()
>>> buf.fill('hello world')
>>> len(buf)
11
>>> buf.read()
hello world
>>> # and so on
```

Also, PEP8 requires two blank lines before class declarations.

smart_open/bytebuffer.py
```diff
@@ -0,0 +1,74 @@
...
+
+        part = self._bytes[self._pos : self._pos + size]
+        return part
+
+    def empty(self):
+        """Remove all bytes from the buffer"""
+        self._bytes = b''
+        self._pos = 0
+
+    def fill(self, reader_or_iterable, size = -1):
```
mpenkov: Please document function/method parameters fully, here and everywhere. Here's an [example](https://github.com/RaRe-Technologies/smart_open/blob/ce3d8acf7810358c93cbe11375dc48485901ccdf/smart_open/s3.py#L72).

aperiodic:
Ok, I'll address these comments on Monday UTC, between 16:00 and 23:00 UTC.

smart_open/bytebuffer.py
aperiodic: I've never used doctest before, but it seemed simple enough, so I gave it a shot. However, I ran into problems with making the same doctest work in both Python 2 and 3. If you're interested in the details, I pushed up [a commit with the doctest I tried to use and a commit message describing the problem to another branch on my fork of smart_open](https://github.com/aperiodic/smart_open/commit/9ab06331e880988dca7b33f7f89a1ecce01f53f4).

smart_open/bytebuffer.py
aperiodic: Fixed in https://github.com/RaRe-Technologies/smart_open/pull/284/commits/164339a2a836fb721578180a853d762bf67c3723

smart_open/bytebuffer.py
aperiodic: Fixed in https://github.com/RaRe-Technologies/smart_open/pull/284/commits/2e88391c577a0cbd3c0e22b96765cae653707dab

smart_open/bytebuffer.py
aperiodic: Fixed in https://github.com/RaRe-Technologies/smart_open/pull/284/commits/02ca639b326b9147b4d8d99b6fb8eca6e3acf3b8

aperiodic:
I've addressed all of your latest comments except

aperiodic:
Sorry, hit the wrong button.

I've addressed all of your latest comments except the doctest example in the ByteBuffer class docstring. I wasn't able to get the same example to pass in both Python 2 and 3; you can see [the commit I pushed up to another branch for the details](https://github.com/aperiodic/smart_open/commit/9ab06331e880988dca7b33f7f89a1ecce01f53f4).

smart_open/bytebuffer.py
aperiodic: The class docstring is addressed in https://github.com/RaRe-Technologies/smart_open/pull/284/commits/02ca639b326b9147b4d8d99b6fb8eca6e3acf3b8

mpenkov:
Don't worry about Py2/3 compatibility in the doctests. As long as it passes under Py3 and the example clearly states that, it's OK. More in-depth Py2/3 support details get handled in the unit tests.

So, please add your doctests to this PR.

smart_open/bytebuffer.py
aperiodic: I've added the doctest (with a note that it only passes in Python 3) in https://github.com/RaRe-Technologies/smart_open/pull/284/commits/c604a78b56ea3214dcc8488c68d6fcfdd1410613.

aperiodic:
Sorry, I force-pushed because I accidentally pushed up the earlier version of the doctest commit that was lacking the note about it only passing in Python 3. Hope it's not too much of an issue.

aperiodic:
Is this ready to merge? I believe I've addressed all your comments (I don't know why GitHub still shows changes requested given all those comments have been resolved).

mpenkov:
From your description, the performance gains are huge, but our test benchmarks don't pick them up. This is likely because the benchmarks are lacking. It would be good to add a test case to our integration tests (see integration-tests/test_s3.py) to prevent future regressions. Can you please have a look at this?

aperiodic:
I've got a benchmark, I'm iterating on the size so it demonstrates the improvement without taking an excessively long time (I'm shooting for a minute or so).

aperiodic:
Results from running the benchmark on the current master branch:
```
-------------------------------------------------- benchmark: 1 tests --------------------------------------------------
Name (time in s)                       Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------
test_s3_performance_small_reads     4.8049  5.2253  4.9774  0.1563  4.9470  0.1763       2;0  0.2009       5           1
------------------------------------------------------------------------------------------------------------------------
```
Results from running benchmark on this PR branch:
```
-------------------------------------------------- benchmark: 1 tests --------------------------------------------------
Name (time in s)                       Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------
test_s3_performance_small_reads     1.5846  1.6942  1.6350  0.0407  1.6360  0.0519       2;0  0.6116       5           1
------------------------------------------------------------------------------------------------------------------------
```

mpenkov:
@aperiodic Congrats on your first merged PR to smart_open :1st_place_medal: Awesome work!

aperiodic:
Thank you, and thanks for the review! I enjoyed contributing.

PR #292, opened 2019-04-15
feat: adding password option to ssh, scp and sftp

davidwneary:
*Motivation*: I want to be able to access files on an sftp server that uses username / password for authentication.

This adds the ability to parse a url like `sftp://username:pwd@host:port/path`.

### DIFF
--- a/smart_open/helpers.py
+++ b/smart_open/helpers.py
@@ -1,0 +1,52 @@
+import inspect
+
+
+def inspect_kwargs(kallable):
+    #
+    # inspect.getargspec got deprecated in Py3.4, and calling it spews
+    # deprecation warnings that we'd prefer to avoid.  Unfortunately, older
+    # versions of Python (<3.3) did not have inspect.signature, so we need to
+    # handle them the old-fashioned getargspec way.
+    #
+    try:
+        signature = inspect.signature(kallable)
+    except AttributeError:
+        args, varargs, keywords, defaults = inspect.getargspec(kallable)
+        if not defaults:
+            return {}
+        supported_keywords = args[-len(defaults):]
+        return dict(zip(supported_keywords, defaults))
+    else:
+        return {
+            name: param.default
+            for name, param in signature.parameters.items()
+            if param.default != inspect.Parameter.empty
+        }
+
+
+def check_kwargs(kallable, kwargs, logger):
+    """Check which keyword arguments the callable supports.
+
+    Parameters
+    ----------
+    kallable: callable
+        A function or method to test
+    kwargs: dict
+        The keyword arguments to check.  If the callable doesn't support any
+        of these, a warning message will get printed.
+    logger: Logger
+        The logger to which warnings will be printed
+
+    Returns
+    -------
+    dict
+        A dictionary of argument names and values supported by the callable.
+    """
+    supported_keywords = sorted(inspect_kwargs(kallable))
+    unsupported_keywords = [k for k in sorted(kwargs) if k not in supported_keywords]
+    supported_kwargs = {k: v for (k, v) in kwargs.items() if k in supported_keywords}
+
+    if unsupported_keywords:
+        logger.warn('ignoring unsupported keyword arguments: %r', unsupported_keywords)
+
+    return supported_kwargs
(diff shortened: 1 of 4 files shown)


mpenkov:
Thank you for your contribution.

Can you please have a look at the comments here: https://github.com/RaRe-Technologies/smart_open/pull/279?

Basically, I don't want to bother with handling each individual ssh.connect parameter separately. There are many of them, and from smart_open's point of view, all those parameters are opaque (we just pass them to ssh.connect).

Can you update your PR to handle all connect parameters instead of just password? For example, we want to be able to do this:

```python
tp = {'password': 'secret', 'compress': True}
with smart_open.open('sftp://username@host:port/path') as fin:
    ...
```

There is a potential for conflict, because people can specify a password in two places: 1) URL and 2) transport_params. It doesn't really matter _how_ we resolve this conflict (override), as long as we document it clearly and warn the user when the override occurs.

Please let me know if anything above is unclear.

davidwneary:
Thanks for your comments @mpenkov! Makes total sense, I'll see what I can do

davidwneary:
@mpenkov I've just pushed a commit addressing your previous comment - let me know what you think.

Few comments:
1. I've done this quite quickly just to get it working - I'd like to add some unit tests of `ssh` to test the `transport_params` are passed through properly before merging
2. As mentioned in the commit, I had to move `check_kwargs` into a `helper` module in order to use it in the `ssh` module, let me know if this is undesirable
3. Just re-read the MR you linked to and seen:
    > Furthermore, I think instead of passing just the key, we should pass `connect_kwargs` as dictionary

    My latest changes pass the whole of `transport_params` into the `connect` call so `transport_params` would be of the form:
    ```
    tp = {'password': 'secret', 'compress': True}
    ```
    would 
    ```
    tp = {connect_kwargs: {'password': 'secret', 'compress': True}}
    ```
    be preferable? Do you see there being any transport params in the future that don't belong to the `connect_kwargs`?
4. I've updated the docstring of `ssh.open`, is there any other documentation I should update wrt. the precedence of `transport_params` for `username` and `password`?

Cheers

luckydonald:
Quick note:

> It doesn't really matter how we resolve this conflict (override), as long as we document it clearly and warn the user when the override occurs.

For S3 urls the `transport_params` take precedence over the `url`.

mpenkov:
@davidwneary Thank you for making the changes, and sorry it has taken me so long to review them. Let me answer your questions first.

1. Yes, unit tests would be great. You can simply mock out the connect function from paramiko: as long as the parameters are all making it to that function, then it's OK.
2. I  don't think you need to bother with check_kwargs. It's only necessary for backwards compatibility of the deprecated smart_open.smart_open function. That function accepted **kwargs very liberally and never complained, so in order to maintain that behavior, we had to check_kwargs and log an error. With the transport_params stuff, we don't need to enforce the contract between the user and whatever they're passing keyword arguments to. The interpreter can do it for us.
3. I'll answer this separately below.

We need to choose between:

```python
    tp = {'password': 'secret', 'compress': True}  # option 1
    tp = {connect_kwargs: {'password': 'secret', 'compress': True}}  # option 2
```

I agree that option 1 looks better (flat is better than nested, etc). However, option 2 is more powerful because it allows us to eventually pass keyword arguments to functions other than connect. The [paramiko API](http://docs.paramiko.org/en/2.4/api/client.html) is quite rich, and it's possible that people will want to pass other options as well. Finally, option 2 is also consistent with the rest of the API. So let's go with that.

4. I don't think you need to do any special handling for transport_params, username and password. We can let the interpreter handle that for us. Consider:

```python
def function(a=1, b=2):
   pass

kwargs = {'a': 3, 'b': 4}
function(1, 2, **kwargs)
```

The above code will crash: `TypeError: function() got multiple values for argument 'a'`. I think that is acceptable behavior for us.

mpenkov:
@davidwneary How's it going? Just wanted to check up on the progress of this PR. It looks almost done, if you have the time, let's finish it so we can merge.

interpolatio:
hi @davidwneary
Do you plan to complete PR? or can i continue?

mpenkov:
Closing in favor of #369

PR #298, opened 2019-04-17
Added scripts to automate the release process

mpenkov:
The [existing release howto](https://github.com/RaRe-Technologies/smart_open/wiki/How-to-release-smart_open) is good, but most of those steps can be automated, reducing the load on the maintainer. These scripts are an initial step in that direction.

### DIFF
--- a/release/merge.sh
+++ b/release/merge.sh
@@ -1,0 +1,44 @@
+#
+# Performs the following tasks:
+#
+# - Merges the current release branch into master
+# - Applies a tag to master
+# - Pushes the updated master branch and its tag to upstream
+#
+# Use it like this:
+#
+#     bash release/merge.sh
+#
+# Expects smart_open/VERSION to be correctly incremented for the new release.
+#
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+version="$(head -n 1 smart_open/VERSION)"
+
+read -p "Push version $version to github.com? yes or no: " reply
+if [ "$reply" != "yes" ]
+then
+    echo "aborted by user"
+    exit 1
+fi
+
+#
+# Delete the local master branch in case one is left lying around.
+#
+set +e
+git branch -D master
+set -e
+
+git checkout upstream/master -b master
+git merge --no-ff release-${version}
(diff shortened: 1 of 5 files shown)


release/merge.sh
```diff
@@ -0,0 +1,44 @@
+#
```
piskvorky: Maybe include the hash bang? Does the particular shell matter?

release/summarize_pr.sh
```diff
@@ -0,0 +1,24 @@
...
+# - Removes native xz support (PR [#282](https://github.com/RaRe-Technologies/smart_open/pull/282), [@tdhopper](https://github.com/tdhopper))
+# - backward compatibility fixes (PR [#294](https://github.com/RaRe-Technologies/smart_open/pull/294), [@mpenkov](https://github.com/mpenkov))
+#
+# You can paste this directly into CHANGELOG.md
+#
+for prid in "$@"
+do
+    api_url="https://api.github.com/repos/RaRe-Technologies/smart_open/pulls/${prid})"
+    json="$(wget --quiet -O - ${api_url})"
+    title="$(echo "$json" | jq .title --raw-output)"
```
piskvorky: Is this installed by default? Do the scripts assume any specific environment / OS / dependencies?

piskvorky:
Awesome work @mpenkov ! A step in the right direction indeed.

release/merge.sh
mpenkov: The hashbangs are only required if you execute the scripts directly, e.g.

    $ ./merge.sh

In this case, it's impossible, because the scripts do not have execute permissions. This is by design, so that you have to explicitly specify the shell and do:

    $ bash merge.sh

In practice, shells other than bash may be usable, but I haven't tested them.

release/summarize_pr.sh
mpenkov: Good catch. I'll document the dependencies. To answer your questions briefly: 

- no, but wget/json/etc packages are easily available for most Linux distros
- There are dependencies like bash and the above utils, but they are really general

piskvorky: Thanks. I can also test this on my environment (OSX, GNU bash).

PR #307, opened 2019-04-23
Fix two S3 bugs

### LINKED ISSUE #304: transport_params: default value for keyword argument is mutable
For some reason, `transport_params.session` defaults to something I didn't set, causing it to ignore the arguments set via url.

Here is my code:
```py
with open('s3://ACCESS_TOKEN:ACCESS_SECRET@s3.example.com:443@bucketname/some_file.mp4', 'wb') as s3_file:
    # code
# end with
```
The logs:
```
"2019-04-22T10:55:45.340767000Z DEBUG"    smart_open.smart_open_lib.open: {'transport_params': {'session': Session(region_name=None)}, 'ignore_ext': False, 'opener': None, 'closefd': True, 'newline': None, 'errors': None, 'encoding': None, 'buffering': -1, 'mode': 'wb', 'uri': 's3://ACCESS_TOKEN:ACCESS_SECRET@s3.example.com:443@bucketname/some_file.mp4'}
"2019-04-22T10:55:45.341045000Z DEBUG"    smart_open.smart_open_lib._s3_open_uri: s3_open_uri: {'transport_params': {'session': Session(region_name=None)}, 'mode': 'wb', 'parsed_uri': Uri(scheme='s3', uri_path=None, bucket_id='bucketname', key_id='some_file.mp4', port=443, host='s3.example.com', ordinary_calling_format=True, access_id='ACCESS_TOKEN', access_secret='ACCESS_SECRET', user=None)}
"2019-04-22T10:55:45.341139000Z WARNING"  smart_open.smart_open_lib._s3_open_uri: ignoring credentials parsed from URL because they conflict with transport_params.session. Set transport_params.session to None to suppress this warning.
```

It shouldn't have any session though, and just use the url params.

### LINKED ISSUE #305: Custom S3 host is ignored when using the url format.
https://github.com/RaRe-Technologies/smart_open/blob/052ff93c110ab9e4ef94bd55667c9b6707d1ceee/smart_open/smart_open_lib.py#L589-L593
This is missing something like `endpoint_url`:
```py
transport_params['session'] = boto3.Session(
        aws_access_key_id=access_key,
        aws_secret_access_key=secret_key,
        endpoint_url='{proto}://{host}:{port}'.format(proto="https", host="s3.example.com" port=443),
)
```
<br><br>

---------
Maybe you could add integrations checks against a self hosted instance.
Minio is quite easy to set up:  [Quick start guide](https://docs.min.io/docs/minio-quickstart-guide.html).
Also they have a public running demo instance at [play.min.io](https://play.min.io). That probably could be used too.

mpenkov:
- Fix #304 
- Fix #305

### DIFF
--- a/integration-tests/test_minio.py
+++ b/integration-tests/test_minio.py
@@ -1,0 +1,63 @@
+import logging
+import boto3
+
+from smart_open import open
+
+#
+# These are publicly available via play.min.io
+#
+KEY_ID = 'Q3AM3UQ867SPQQA43P2F'
+SECRET_KEY = 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG'
+ENDPOINT_URL = 'https://play.min.io:9000'
+
+
+def read_boto3():
+    """Read directly using boto3."""
+    session = get_minio_session()
+    s3 = session.resource('s3', endpoint_url=ENDPOINT_URL)
+
+    obj = s3.Object('smart-open-test', 'README.rst')
+    data = obj.get()['Body'].read()
+    logging.info('read %d bytes via boto3', len(data))
+    return data
+
+
+def read_smart_open():
+    url = 's3://Q3AM3UQ867SPQQA43P2F:zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG@play.min.io:9000@smart-open-test/README.rst'  # noqa
+
+    #
+    # If the default region is not us-east-1, we need to construct our own
+    # session.  This is because smart_open will create a session in the default
+    # region, which _must_ be us-east-1 for minio to work.
+    #
+    tp = {}
+    if get_default_region() != 'us-east-1':
+        logging.info('injecting custom session')
+        tp['session'] = get_minio_session()
+    with open(url, transport_params=tp) as fin:
+        text = fin.read()
+        logging.info('read %d characters via smart_open', len(text))
+        return text
+
+
+def get_minio_session():
+    return boto3.Session(
+        region_name='us-east-1',
+        aws_access_key_id=KEY_ID,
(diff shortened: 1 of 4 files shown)


luckydonald:
@mpenkov Hey, any feedback to the questions above?

mpenkov:
Which questions are you referring to? I don't see any in this PR

luckydonald:
I asked [here](https://github.com/RaRe-Technologies/smart_open/pull/307/files/607de4757a1c1be199f67f17d88fade05e99957a..2219d67da29f8df48e8d59a16007adce2d7fd31f#diff-84d94ca6ab1b619ebc2adc531e8db975R31) (It says pending, so maybe that didn't go through):

> Why is that? Is this specific to minio, or all custom S3 implementations?
Shouldn't that be automatic?

> Needing to set that manually feels like it could cause bugs very easily, later on.
My assumption would be it'll be smart and "just work". Maybe `smart_open` can set the session accordingly?

mpenkov:
I think it's a minio thing, but I don't use it, so my knowledge is limited.

 I don't think it's worth including such details in smart_open library for several reasons:

- Minio is a nonstandard S3 implementation (it's an edge case)
- it's already possible to set and override the default S3 region via boto3 mechanisms
- In the worst case, it's possible to override the region via transport_params

luckydonald:
How did you stumbled upon the fact that it needs to be `us-east-1`?
That's new to me, I didn't even know that, as it just worked<sup>TM</sup> for me.
Does that also apply to other S3 vendors out there, like wasabi etc?

mpenkov:
Trial and error. That's another reason why I'm not confident about including it in the library.

No, I haven't tried other S3 vendors.

PR #315, opened 2019-05-18
Don't use s3 bucket_head to check for bucket existence

### LINKED ISSUE #314: writing to S3 requires unnecessary permissions
I'm trying to write to an S3 bucket for which I have permission to write to a path in the bucket, but I don't have permission to list the root of the bucket, which is a fairly common use case when using S3 to transfer files to/from multiple organizations in the same bucket.  The write fails because of a check on line 426 of s3.py: it tries a head_bucket operation which requires[1] the ListBucket permission, which isn't really needed to write an object.

Commenting out lines 425-428 allows the write to succeed since it no longer does any operations in the root of the bucket.

[1] https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketHEAD.html

caboteria:
This PR fixes #314 .

This code caused problems when you have write permission to an s3
bucket path but not read permission to the bucket root.

This code was added to prevent implicit bucket creation (to fix issue
#154) but it appears that it is no longer needed.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -418,17 +418,12 @@ multipart upload may fail")
             multipart_upload_kwargs = {}
 
         s3 = session.resource('s3', **resource_kwargs)
-
-        #
-        # https://stackoverflow.com/questions/26871884/how-can-i-easily-determine-if-a-boto-3-s3-bucket-resource-exists
-        #
         try:
-            s3.meta.client.head_bucket(Bucket=bucket)
+            self._object = s3.Object(bucket, key)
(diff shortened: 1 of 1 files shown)


caboteria:
test_buffered_writer_wrapper_works in test_s3.py tests the happy path.  test_nonexisting_bucket in test_s3.py tests that the proper exception (ValueError) is thrown when the bucket doesn't exist.  Props for great test coverage!

mpenkov:
OK, thank you, and congrats on your first contribution to smart_open! :1st_place_medal:

PR #318, opened 2019-05-20
Dont list buckets in s3 tests

caboteria:
While working on a previous PR I stubbed my toe on AWS permissions.  The tests needed to be able to list all of the bucket names in my account but I'm not comfortable giving those permissions to Travis so I modified the cleanup_bucket() method to directly delete the bucket instead of listing each bucket to decide whether to delete it or not.

After I made that change I noticed that I was getting more test failures due to race conditions - tests expecting the test bucket to be there when it wasn't or vice versa. To make this happen less often I changed the suite so it creates the test bucket before the tests run and deletes it afterward instead of create/delete per test. This is more stable and saves 20-30s in the s3 tests when running without mocks.

### DIFF
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -35,16 +35,18 @@ def maybe_mock_s3(func):
 
 
 def cleanup_bucket(s3, delete_bucket=False):
-    for bucket in s3.buckets.all():
-        if bucket.name == BUCKET_NAME:
-            for key in bucket.objects.all():
-                key.delete()
+    bucket = s3.Bucket(BUCKET_NAME)
+    try:
+        for key in bucket.objects.all():
+            key.delete()
 
-            if delete_bucket:
-                bucket.delete()
-                return False
-            return True
-    return False
+        if delete_bucket:
+            bucket.delete()
+            return False
+    except s3.meta.client.exceptions.NoSuchBucket:
+        return False
+
+    return True
 
 
 def create_bucket_and_key(
@@ -53,10 +55,8 @@ def create_bucket_and_key(
     # fake (or not) connection, bucket and key
     logger.debug('%r', locals())
     s3 = boto3.resource('s3')
-    bucket_exist = cleanup_bucket(s3)
-
-    if not bucket_exist:
-        mybucket = s3.create_bucket(Bucket=bucket_name)
+    mybucket = s3.create_bucket(Bucket=bucket_name)
+    cleanup_bucket(s3)
 
     #
     # In real life, it can take a few seconds for the bucket to become ready.
@@ -74,7 +74,7 @@ def create_bucket_and_key(
             logger.error('caught %r, retrying', err)
             time.sleep(sleep_time)
 
-    assert False, 'failed to create bucket after %d attempts' % num_attempts
+    assert False, 'failed to create bucket %s after %d attempts' % (bucket_name, num_attempts)
 
 
 def ignore_resource_warnings():
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -18,7 +18,13 @@ import six
 import smart_open
 import smart_open.s3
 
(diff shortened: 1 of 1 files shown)


smart_open/tests/test_s3.py
```diff
@@ -34,29 +34,30 @@ def maybe_mock_s3(func):
...
-
-
-def create_bucket_and_key(
-        bucket_name=BUCKET_NAME, key_name=KEY_NAME, contents=None,
-        num_attempts=12, sleep_time=5):
-    # fake (or not) connection, bucket and key
-    logger.debug('%r', locals())
+@maybe_mock_s3
+def setUpModule():
+    global s3
```
mpenkov: Is it possible to avoid using a global? It's a bit of a code smell.

If it's not possible, then please add a comment about these two functions. Things to mention:

- They're special functions that the unittest library looks for
- setUpModule gets called before any code from the module is run (?)
- tearDownModule gets called ... (when?)
- The motivation (create a new bucket when the module is loaded, and delete the bucket when ...

It's also worth making `s3` more obviously a global. Something like `GLOBAL_S3` would be better. That way, when people touch it, they'll at least know what they're doing.

mpenkov:
> The tests needed to be able to list all of the bucket names in my account but I'm not comfortable giving those permissions to Travis so I modified the cleanup_bucket() method to directly delete the bucket instead of listing each bucket to decide whether to delete it or not.

You shouldn't have to give any of your credentials to travis. My understanding is that it uses **our** credentials, specified in the env.global.secure section in the .travis.yml file. Could you please confirm?

smart_open/tests/test_s3.py
```diff
@@ -65,16 +66,23 @@ def create_bucket_and_key(
...
+            return
         except botocore.exceptions.ClientError as err:
             logger.error('caught %r, retrying', err)
             time.sleep(sleep_time)
 
-    assert False, 'failed to create bucket after %d attempts' % num_attempts
+    assert False, 'failed to create bucket %s after %d attempts' % (BUCKET_NAME, num_attempts)
+
+
+def clean_and_populate_bucket(num_keys=10):
```
mpenkov: Can you please move this function definition down closer to where the populate_bucket function was (L593 in previous version)? It'd make it a lot easier to review.

smart_open/tests/test_s3.py
```diff
@@ -98,14 +106,12 @@ def setUp(self):
 
     def tearDown(self):
         smart_open.s3.DEFAULT_MIN_PART_SIZE = self.old_min_part_size
-        s3 = boto3.resource('s3')
-        cleanup_bucket(s3, delete_bucket=True)
```
mpenkov: I think you should still call `cleanup_bucket` at the end of each test, to ensure that the tests are fully isolated from each other.

smart_open/tests/test_s3.py
```diff
@@ -261,13 +267,8 @@ class BufferedOutputBaseTest(unittest.TestCase):
     def setUp(self):
         ignore_resource_warnings()
 
-    def tearDown(self):
-        s3 = boto3.resource('s3')
-        cleanup_bucket(s3, delete_bucket=True)
```
mpenkov: As above, I think this is actually necessary and desired.

caboteria:
> You shouldn't have to give any of your credentials to travis. My understanding is that it uses our credentials, specified in the env.global.secure section in the .travis.yml file. Could you please confirm?

This is true only when the tests run in the context of the Travis RaRe-Technologies/smart_open project.  Each project has its own unique private/public key pair.  I wanted to be able to trigger builds of my fork so I set up a Travis project for my fork so the private key is different.  I can write a wiki page explaining how to do that if you'd like, it's not hard once you figure out what permissions are needed.

caboteria:
I think I did all of the things you asked for but let me know if I missed something.

mpenkov:
Great work. Thank you for your contribution!

PR #325, opened 2019-06-16
Add version_id transport parameter for fetching a specific S3 object version

### LINKED ISSUE #306: Failed file access should raise IOError not ValueError (S3)
See https://github.com/RaRe-Technologies/smart_open/issues/303#issuecomment-485426630 for details.

I recognise this to be a breaking change (Unless you make a subclass extending both Exception classes) so proper semantic verion should be assigned.

interpolatio:
Added file check by id.
Added check for the existence of the passed id.

Fix #306


smart_open/s3.py
```diff
@@ -26,8 +26,7 @@
 except ImportError:
     warnings.warn("multiprocessing could not be imported and won't be used")
 
-
-DEFAULT_MIN_PART_SIZE = 50 * 1024**2
+DEFAULT_MIN_PART_SIZE = 50 * 1024 ** 2
```
mpenkov: Please revert this superfluous change

smart_open/s3.py
```diff
@@ -97,6 +97,8 @@ def open(
     multipart_upload_kwargs: dict, optional
         Additional parameters to pass to boto3's initiate_multipart_upload function.
         For writing only.
+    version_id: str, optional
+        Version of the object.
```
```suggestion
        Version of the object. If None, will fetch the most recent version.
```

smart_open/s3.py
```diff
@@ -112,6 +114,7 @@ def open(
         fileobj = SeekableBufferedInputBase(
             bucket_id,
             key_id,
+            version_id,
```
mpenkov: Pass keyword parameters by name.

```suggestion
            version_id=version_id,
```

smart_open/s3.py
```diff
@@ -177,7 +186,8 @@ def seek(self, position):
             #
             self._body = io.BytesIO()
         else:
-            self._body = self._object.get(Range=range_string)['Body']
+            getkwargs = dict(Range=range_string)
+            self._body = _get(self._object, self._version_id, **getkwargs)['Body']
```
mpenkov: This is more complicated than it needs to be.

```suggestion
            self._body = _get(self._object, self._version_id, Range=range_string)['Body']
```

The line above is not necessary.

smart_open/s3.py
```diff
@@ -202,6 +212,7 @@ def __init__(self, bucket, key, buffer_size=DEFAULT_BUFFER_SIZE,
         self._object = s3.Object(bucket, key)
         self._raw_reader = RawReader(self._object)
         self._content_length = self._object.content_length
+        self._object.get(VersionId=self.version_id)['ContentLength']
```
mpenkov: This will crash if version_id is None, won't it?

You should probably use the _get function here.

smart_open/s3.py
```diff
@@ -328,24 +339,24 @@ class SeekableBufferedInputBase(BufferedInputBase):
...
-            raise ValueError(
-                '%r does not exist in the bucket %r, '
-                'or is forbidden for access' % (key, bucket)
-            )
-
-        self._raw_reader = SeekableRawReader(self._object, self._content_length)
+            self._content_length = _get(self._object, self._version_id)['ContentLength']
+        except botocore.client.ClientError as error:
+            raise IOError(str(error) + "\n" +
+                          '%r does not exist in the bucket %r, version_id %r'
```
mpenkov: Please use hanging indent (see the previous version of the file for an example).

smart_open/s3.py
```diff
@@ -26,7 +26,8 @@
 except ImportError:
     warnings.warn("multiprocessing could not be imported and won't be used")
 
-DEFAULT_MIN_PART_SIZE = 50 * 1024 ** 2
+
+DEFAULT_MIN_PART_SIZE = 50 * 1024**2
```
mpenkov: What is the purpose of this change? If it's only cosmetic, please remove it.

smart_open/s3.py
```diff
@@ -186,8 +187,7 @@ def seek(self, position):
             #
             self._body = io.BytesIO()
         else:
-            getkwargs = dict(Range=range_string)
-            self._body = _get(self._object, self._version_id, **getkwargs)['Body']
+            self._body = _get(self._object, self._version_id, Range=range_string)['Body']
```
mpenkov: The _get function can raise a ClientError. We should handle it the same way everywhere (wrap in IOError).

smart_open/s3.py
```diff
@@ -212,7 +212,13 @@ def __init__(self, bucket, key, version_id=None, buffer_size=DEFAULT_BUFFER_SIZE
         self._object = s3.Object(bucket, key)
         self._raw_reader = RawReader(self._object)
         self._content_length = self._object.content_length
-        self._object.get(VersionId=self.version_id)['ContentLength']
+        try:
+            self._content_length = _get(self._object, self._version_id)['ContentLength']
+        except botocore.client.ClientError as error:
+            raise IOError(str(error) + "\n" +
```
mpenkov: I think we should move exception handling to the _get function.

smart_open/s3.py
```diff
@@ -352,9 +358,9 @@ def __init__(self, bucket, key, version_id=None, buffer_size=DEFAULT_BUFFER_SIZE
             self._content_length = _get(self._object, self._version_id)['ContentLength']
         except botocore.client.ClientError as error:
             raise IOError(str(error) + "\n" +
-                          '%r does not exist in the bucket %r, version_id %r'
-                          'or is forbidden for access' % (key, bucket, self._version_id)
-                          )
+                '%r does not exist in the bucket %r, version_id %r'
```
mpenkov: OK, we should _definitely_ move the exception handling to _get to avoid duplicating code like this.

smart_open/s3.py
```diff
@@ -127,10 +131,21 @@ def open(
...
 
 
+def _get(s3_object, version=None, **kwargs):
+    try:
+        if version is not None:
+            kwargs['VersionId'] = version
+        return s3_object.get(**kwargs)
+    except botocore.client.ClientError as error:
+        raise IOError(str(error) + "\n" +
+                      '%r does not exist in the bucket %r, version_id %r'
```
mpenkov: Hanging indent

mpenkov: Also, please move the conditional outside the try block. That code will never throw the ClientError

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert False, 'failed to create bucket %s after %d attempts' % (BUCKET_NAME, num_attempts)
+
+
+@maybe_mock_s3
+class TestVersionId(unittest.TestCase):
+
+    def test_good_id(self):
+        """Does version_id into s3 work correctly?"""
+        test_string_ver0 = u"String version 1.0".encode('utf8')
```
mpenkov: You're storing bytes in a variable named "string". This is misleading.

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,101 @@
...
+        test_string_ver1 = u"String version 2.0".encode('utf8')
+        # write into key
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
```
mpenkov: This sort of logic isn't part of the actual test. It should be in a setUp method. That way, your other test cases can reuse it, too.

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,101 @@
...
+            fout.write(test_string_ver0)
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
+            fout.write(test_string_ver1)
+
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+        with smart_open.s3.SeekableBufferedInputBase(BUCKET_NAME, WRITE_KEY_NAME,check_version) as fin:
```
mpenkov: Please run your code through flake8,

README.rst
```diff
@@ -257,7 +257,27 @@ Since going over all (or select) keys in an S3 bucket is a very common operation
   ...     print(key, round(len(content) / 2**20))
   annual/monthly_rain/2010.monthly_rain.nc 13
   annual/monthly_rain/2011.monthly_rain.nc 13
-  annual/monthly_rain/2012.monthly_rain.nc 13
+  annual/monthly_rain/2012.monthly_rain.nc 13 
+
+Specific s3 object version
+--------------------------------------
+
+The ''open'' function has the parameter version_id, which allows you to get the desired version of the file from bucket S3.
```
mpenkov: This is a bit misleading. You're talking about the smart_open.s3.open function, not the smart_open.open function (but you're importing "from smart_open import open", which is even more misleading). While what you say is correct, nobody really uses smart_open.s3.open, because it's mostly an internal function. everybody uses smart_open.open (or the older smart_open.smart_open).

Please rewrite these instructions and use the top-level function in your examples.

README.rst
```diff
@@ -257,7 +257,27 @@ Since going over all (or select) keys in an S3 bucket is a very common operation
...
+
+*Important*: Version control disabled by default. Before starting to use version_id parameter, enable version control in bucket S3. Read https://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html
+
+.. code-block:: python
+
+  >>> from smart_open import open
+  >>> with s3.open('smart-open-bucket', 'test_s3_version.txt', 'rb',version_id='azaMK5Yf_wg03QI1vzKBgzDG8RMin3l1') as file:
```
mpenkov: flake8

smart_open/s3.py
```diff
@@ -97,6 +98,8 @@ def open(
     multipart_upload_kwargs: dict, optional
         Additional parameters to pass to boto3's initiate_multipart_upload function.
         For writing only.
+    version_id: str, optional
+        Version of the object. If None, will fetch the most recent version.
```
mpenkov: This comment makes sense for when we're reading an object via S3.

How does this parameter behave when writing?

smart_open/s3.py
```diff
@@ -127,10 +131,22 @@ def open(
...
     return fileobj
 
 
+def _get(s3_object, version=None, **kwargs):
+    if version is not None:
+        kwargs['VersionId'] = version
+    try:
+        return s3_object.get(**kwargs)
+    except botocore.client.ClientError as error:
+        raise IOError(
```
mpenkov: It's not common to have multiline exception messages in Python. Let's see if we can make this more brief.

What will the user see if we raise:

```python
raise IOError(str(error))
```

?

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,101 @@
...
+        # write into key
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
+            fout.write(test_string_ver0)
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
+            fout.write(test_string_ver1)
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+
+        with self.assertRaises(IOError):
+            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'rb', version_id=check_version+check_version)
```
mpenkov: You can make this test simpler:

```suggestion
            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'rb', version_id='bad-version-does-not-exist')
```

smart_open/s3.py
interpolatio: Return exception
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.

smart_open/s3.py
interpolatio: test result:
AssertionError: OSError not raised

Need to create a new exception?

README.rst
```diff
@@ -257,7 +257,27 @@ Since going over all (or select) keys in an S3 bucket is a very common operation
...
+--------------------------------------
+
+The ``smart_open``'s ``open`` function has the parameter version_id, which allows you to get the desired version of the file from bucket S3.
+
+*Important*: Version control disabled by default. Before starting to use version_id parameter, enable version control in bucket S3. Read https://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html
+
+.. code-block:: python
+
+  >>> from smart_open import open
+  >>> with open('smart-open-bucket', 'test_s3_version.txt', 'rb', version_id='azaMK5Yf_wg03QI1vzKBgzDG8RMin3l1') as file:
```
mpenkov: Are you sure this actually works?

version_id is a parameter that gets passed through the transport_params dict, isn't it?

smart_open/s3.py
mpenkov: From boto3's point of view, does it make sense to pass a version number when writing? Or is it something that should **never** happen?

smart_open/s3.py
mpenkov: OK, let's do:

```python
raise IOError(
    'unable to access bucket: %r key: %r version: %r error: %s' % (
        s3_object.bucket_name, s3_object.key, version, error
    )
)
```

mpenkov:
Also, looks like unit tests are failing, please fix:

```
======================================================================
ERROR: test_readline_tiny_buffer (smart_open.tests.test_s3.SeekableBufferedInputBaseTest)
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/home/travis/build/RaRe-Technologies/smart_open/smart_open/tests/test_s3.py", line 249, in test_readline_tiny_buffer
    with smart_open.s3.BufferedInputBase(BUCKET_NAME, KEY_NAME, buffer_size=8) as fin:
  File "/home/travis/build/RaRe-Technologies/smart_open/smart_open/s3.py", line 222, in __init__
    self._content_length = _get(self._object, self._version_id)['ContentLength']
AttributeError: 'BufferedInputBase' object has no attribute '_version_id'
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,97 @@
...
+        self.test_ver0 = u"String version 1.0".encode('utf8')
+        self.test_ver1 = u"String version 2.0".encode('utf8')
+        # write into key
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
+            fout.write(self.test_ver0)
+        with smart_open.s3.BufferedOutputBase(BUCKET_NAME, WRITE_KEY_NAME) as fout:
+            fout.write(self.test_ver1)
+
+    def test_good_id(self):
+        """Does version_id into s3 work correctly?"""
```
piskvorky: This sentence is not grammatical… some missing words?

(here and elsewhere)

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
```
mpenkov:
```suggestion
        """Does passing the version_id parameter into the s3 submodule work correctly when reading?"""
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+        with smart_open.s3.SeekableBufferedInputBase(BUCKET_NAME, WRITE_KEY_NAME, check_version) as fin:
+            expected = fin.read()
+        self.assertEqual(expected, self.test_ver0)
+
+    def test_bad_id(self):
+        """Does version_id exception into s3 work correctly?"""
```
mpenkov:
```suggestion
        """Does passing an invalid version_id exception into the s3 submodule get handled correctly?"""
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+
+        with self.assertRaises(IOError):
+            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'rb', version_id='bad-version-does-not-exist')
+
+    def test_bad_mode(self):
+        """Check the write mode selection when transferring the version"""
```
mpenkov:
```suggestion
        """Do we correctly handle non-None version when writing?"""
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+        check_version = list(versions)[0].get()['VersionId']
+
+        with self.assertRaises(ValueError):
+            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'wb', version_id=check_version)
+
+    def test_top_level(self):
+        "Checks top level function open"
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+        with open("s3://"+BUCKET_NAME+"/"+WRITE_KEY_NAME, 'rb', check_version) as fin:
```
mpenkov:
```suggestion
        with open("s3://%s/%s" % (BUCKET_NAME, WRITE_KEY_NAME, 'rb', check_version) as fin:
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_bad_mode(self):
+        """Check the write mode selection when transferring the version"""
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+
+        with self.assertRaises(ValueError):
+            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'wb', version_id=check_version)
+
+    def test_top_level(self):
+        "Checks top level function open"
```
mpenkov:
```suggestion
        """Does the top-level open function now accept the version_id parameter?"""
```

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+        except botocore.exceptions.ClientError as err:
+            logger.error('caught %r, retrying', err)
+            time.sleep(sleep_time)
+
+    assert False, 'failed to create bucket %s after %d attempts' % (BUCKET_NAME, num_attempts)
+
+
+@maybe_mock_s3
+class TestVersionId(unittest.TestCase):
+    def setUp(self):
```
mpenkov: This method gets run each time you run the test. So, it gets run around 4 times. Each time you run the test, you're adding two new versions of the file, so by the end of the entire test suite, you have **eight** versions of the file. This is bad, because it means your tests have side effects, and they can interfere with each other.

I think you should move this version creation logic to the top-level code that creates the bucket, so those versions are only created once.

smart_open/tests/test_s3_version.py
```diff
@@ -0,0 +1,104 @@
...
+
+        with self.assertRaises(ValueError):
+            smart_open.s3.open(BUCKET_NAME, WRITE_KEY_NAME, 'wb', version_id=check_version)
+
+    def test_top_level(self):
+        "Checks top level function open"
+        versions = boto3.resource('s3').Bucket(BUCKET_NAME).object_versions.filter(Prefix=WRITE_KEY_NAME)
+        check_version = list(versions)[0].get()['VersionId']
+        with open("s3://"+BUCKET_NAME+"/"+WRITE_KEY_NAME, 'rb', check_version) as fin:
+            expected = fin.read()
```
mpenkov: This test is missing assertions. You need to check the following:

1. Passing in no version at all gives the newest version of the file
2. Passing in the newest version explicitly gives the same as above
3. Passing in the oldest version gives you the oldest version

Get ready for a surprise ;)

README.rst
```diff
@@ -61,6 +61,14 @@ What?
   ...     break
   '<!doctype html>\n'
 
+  >>> from smart_open import open
+  >>> with open('s3://bucket/key.txt', 'rb', transport_params = {'version_id':'need_key_version_id') as file:
```
piskvorky: PEP8: white space. Also, avoid shadowing built-in names (`file`).

piskvorky:
```suggestion
  >>> with open('s3://bucket/key.txt', 'rb', transport_params={'version_id': 'need_key_version_id') as fin:
  ...      for line in fin:
```

README.rst
```diff
@@ -269,15 +277,12 @@ The ``smart_open``'s ``open`` function has the parameter version_id, which allow
 .. code-block:: python
 
   >>> from smart_open import open
-  >>> with open('s3://bucket/key.txt', 'rb', version_id='azaMK5Yf_wg03QI1vzKBgzDG8RMin3l1') as file:
+  >>> with open('s3://bucket/key.txt', 'rb', transport_params = {'version_id':'need_key_version_id') as file:
```
piskvorky: Dtto.

smart_open/s3.py
```diff
@@ -599,6 +599,12 @@ def iter_bucket(bucket_name, prefix='', accept_key=None,
       >>> # limit to 10k files, using 32 parallel workers (default is 16)
       >>> for key, content in iter_bucket(bucket_name, key_limit=10000, workers=32):
       ...     print key, len(content)
+
+      >>> # get the desired version of the file from bucket S3.
+      >>> from smart_open import open
+      >>> with open('s3://bucket/key.txt', 'rb', transport_params={'version_id':'need_key_version'}) as file:
```
mpenkov: Please apply the comments from @piskvorky to this line as well.

README.rst
```diff
@@ -274,6 +274,31 @@ Since going over all (or select) keys in an S3 bucket is a very common operation
   annual/monthly_rain/2011.monthly_rain.nc 13
   annual/monthly_rain/2012.monthly_rain.nc 13
 
+Specific S3 object version
+--------------------------
+
+The ``version_id`` transport parameter enables you to get the desired version of the object from an S3 bucket.
+
+.. Important::
+    Version control disabled by default.
```
piskvorky:
```suggestion
    Version control is disabled on S3 by default.
```

README.rst
```diff
@@ -274,6 +274,31 @@ Since going over all (or select) keys in an S3 bucket is a very common operation
...
+    Before starting to use version_id parameter, enable version control in bucket S3.
```
piskvorky:
```suggestion
    Before using the ``version_id`` parameter, enable version control in your S3 bucket.
```

mpenkov:
@interpolatio Congrats on your very first contribution to smart_open! 🥇 Keep them coming!

PR #342, opened 2019-07-31
Add support for rt, rt+, wt, wt+, at, at+ methods

interpolatio:


smart_open/smart_open_lib.py
```diff
@@ -341,8 +341,10 @@ def open(
     #
     try:
         binary_mode = {'r': 'rb', 'r+': 'rb+',
+                       'rt': 'rb', 'rt+': 'rb+',
```
mpenkov: What about `wt` and `wt+`?

Are there any other modes we're missing?

smart_open/tests/test_smart_open.py
```diff
@@ -354,6 +354,64 @@ def test_read_str_from_bytes(self):
...
 
+    def test_read_str_from_bytes_api_rt(self):
+        """Can we read strings from a byte stream?"""
+        buffer = make_buffer(initial_value=SAMPLE_BYTES)
+        with smart_open.smart_open(buffer, 'rt') as sf:
+            data = sf.read()
+        self.assertEqual(data, SAMPLE_TEXT)
+
+    def test_read_str_from_bytes_api_a(self):
+        """Can we read strings from a byte stream?"""
```
mpenkov:
```suggestion
        """Can we write, append, and read strings from a real text file?"""
```

smart_open/tests/test_smart_open.py
```diff
@@ -354,6 +354,64 @@ def test_read_str_from_bytes(self):
...
+        buffer = make_buffer()
+        with smart_open.smart_open("test", 'wb') as sf:
```
mpenkov: Please use the tempfile module to generate temporary file names. Do this in the setUp method.

Writing to a local file `test` all the time is a bad idea because it increases the potential of coupling between tests.

smart_open/tests/test_smart_open.py
```diff
@@ -354,6 +354,64 @@ def test_read_str_from_bytes(self):
...
         self.assertEqual(data, SAMPLE_TEXT)
 
+    def test_read_str_from_bytes_api_rt(self):
+        """Can we read strings from a byte stream?"""
+        buffer = make_buffer(initial_value=SAMPLE_BYTES)
+        with smart_open.smart_open(buffer, 'rt') as sf:
+            data = sf.read()
+        self.assertEqual(data, SAMPLE_TEXT)
+
+    def test_read_str_from_bytes_api_a(self):
```
mpenkov: Are you really reading from bytes? You're reading from the local file system.

smart_open/tests/test_smart_open.py
```diff
@@ -354,6 +354,64 @@ def test_read_str_from_bytes(self):
...
+        """Can we read strings from a byte stream?"""
+        buffer = make_buffer()
+        with smart_open.smart_open("test", 'wb') as sf:
+            sf.write(SAMPLE_BYTES)
+        with smart_open.smart_open("test", 'ab') as sf:
```
mpenkov: What is `sf`?

I think `fin` or `fout` is more understandable, based on whether you're reading or writing.

smart_open/smart_open_lib.py
interpolatio: Required modes: `rb`, `ab`, `wb`?

Issue #348, opened 2019-09-10
S3 BufferedInputBase fills buffer during every read


terencezl:
Here is a question about the purpose of buffering in the Input file class `BufferedInputBase`. Assuming buffer helps small amount of reads, and loads from S3 remote end less frequently than the reads from caller, the current implementation doesn't achieve the purpose.

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L284

If the request to read is consistently small, the buffer has to be filled at all times. Wouldn't it be nice to fill to a prefilled buffer size (new param) first?

On the other hand, with `BufferOutputBase`, it behaves properly, only uploading to S3 when many writes reach a certain threshold.

https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L521

mpenkov:
I think we should just fill it to `io.DEFAULT_BUFFER_SIZE` or similar constant, unless `size` is larger. What do you think?

terencezl:
That should certainly work.

mpenkov:
Are you able to make a PR?

PR #350, opened 2019-09-11
BufferedInputBase pre-reserves data to DEFAULT_BUFFER_SIZE

terencezl:


terencezl:
https://github.com/RaRe-Technologies/smart_open/issues/348

smart_open/s3.py
```diff
@@ -281,7 +281,8 @@ def read(self, size=-1):
         # Fill our buffer to the required size.
         #
         # logger.debug('filling %r byte-long buffer up to %r bytes', len(self._buffer), size)
-        self._fill_buffer(size)
+        fill_size = size if size > DEFAULT_BUFFER_SIZE else DEFAULT_BUFFER_SIZE
```
```suggestion
        fill_size = max(size, DEFAULT_BUFFER_SIZE)
```

mpenkov:
@terencezl Are you able to finish this PR? It's almost done.

If not, can you please enable "allow edits by maintainers"? If you're unable to finish, I will push it over the line.

terencezl:
On the right side bar it seems already enabled. I did not see the difference in performance when connected to an on-premise object store, and am not sure how to perform this integration test in a general wat.

On second thought if this is useful, we can even use the user provided buffer size instead of just the default value.

smart_open/s3.py
mpenkov: Could you please apply this change?

I get "You do not have permission to push to this repository."

mpenkov:
Closing as stale

PR #356, opened 2019-10-01
add flake8 and make it happy

mpenkov:


smart_open/s3.py
```diff
@@ -593,15 +597,16 @@ def iter_bucket(bucket_name, prefix='', accept_key=None,
...
+      ...                                 accept_key=lambda key: key.endswith('.json')):
       ...     print key, len(content)
 
       >>> # limit to 10k files, using 32 parallel workers (default is 16)
       >>> for key, content in iter_bucket(bucket_name, key_limit=10000, workers=32):
       ...     print key, len(content)
     """
     if accept_key is None:
-        accept_key = lambda key: True
+        accept_key = _accept_all
```
piskvorky: Why this obfuscation?

smart_open/s3.py
```diff
@@ -593,15 +597,16 @@ def iter_bucket(bucket_name, prefix='', accept_key=None,
     --------
 
       >>> # get all JSON files under "mybucket/foo/"
-      >>> for key, content in iter_bucket(bucket_name, prefix='foo/', accept_key=lambda key: key.endswith('.json')):
+      >>> for key, content in iter_bucket(bucket_name, prefix='foo/',
```
piskvorky: Hanging indent please (not vertical).

smart_open/tests/test_smart_open.py
```diff
@@ -123,10 +125,12 @@ def test_s3_handles_querystring(self):
         self.assertEqual(parsed_uri.key_id, "folder/picture1.jpg?bar")
 
     def test_s3_invalid_url_atmark_in_bucket_name(self):
-        self.assertRaises(ValueError, smart_open_lib._parse_uri, "s3://access_id:access_secret@my@bucket@port/mykey")
+        self.assertRaises(ValueError, smart_open_lib._parse_uri,
```
piskvorky: Hanging indents: here and everywhere else.

smart_open/s3.py
mpenkov: PEP8 [discourages](https://pep8.org) from using lambda expressions other when they are unavoidable, like within function calls.

> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier.

PR #357, opened 2019-10-01
fix file preambles for Python scripts

mpenkov:
This follows on from comments on https://github.com/RaRe-Technologies/gensim/pull/2610#issuecomment-536274793

Main points:

- Added author (not sure how necessary this is, `git blame` will list everyone who has ever touched the file)
- Removed shebangs for non-executable files

I mostly copied the header from `smart_open/__init__.py`. Let me know if you want me to change anything (I've got a script that takes care of it, so it's not a problem).

Once you're happy, I can let this script loose on the gensim repo.


integration-tests/test_184.py
```diff
@@ -1,3 +1,13 @@
+#
+# -*- coding: utf-8 -*-
+#
+# Author: Michael Penkov <m@penkov.dev>
+#
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
+#
+# This code is distributed under the terms and conditions
+# from the MIT License (MIT).
+#
```
piskvorky: Blank line instead of blank comment, to separate the blocks.

integration-tests/test_184.py
```diff
@@ -1,3 +1,13 @@
...
+# -*- coding: utf-8 -*-
```
piskvorky: The leading blank comment on top looks a bit weird, but I looked it up and `coding` can indeed be on the first **or** second line, so I guess this is fine: [PEP263](https://www.python.org/dev/peps/pep-0263/)

smart_open/webhdfs.py
```diff
@@ -1,9 +1,12 @@
-#!/usr/bin/env python
+#
 # -*- coding: utf-8 -*-
 #
+# Author: Michael Penkov <m@penkov.dev>
```
piskvorky: The `Author:` doesn't seem always accurate, so let's drop it altogether. Like you say, `git blame` will be more up-to-date.

integration-tests/test_184.py
mpenkov: We can get rid of the leading blank line. It's not necessary.

Personally, I avoid the shebang unless the script is executable, simply because it is unnecessary otherwise.

> I'd prefer to include the Python shebang always (any harm?).

Well, there's no harm (other than it being an eyesore). But consider this analogy: there's no harm in making **all** of our modules executable (chmod u+x), but we don't do it for the same reason (lack of necessity).

integration-tests/test_184.py
mpenkov: I think it's better to keep the entire preamble as a single comment.

If you use blank lines to separate comment blanks, identifying preambles in files (for scripting, etc.) becomes more difficult.

mpenkov:
@piskvorky For the actual logic, see the release/check_preamble.py file.

It's trivial. I think we can use it on gensim as-is.

@menshikh-iv WDYT?

menshikh-iv:
@mpenkov why not? Though I still don't understand, why `copyright` needed in each file.

mpenkov:
@piskvorky Is there a reason to even include a preamble in each file? We already have a separate license file.

integration-tests/test_184.py
piskvorky: It does not separate comment blanks – it separates preamble from code (imports).

piskvorky:
Yes, it's best practice. People copy around files, without necessarily copying entire projects.

Issue #316, opened 2019-05-19
Render user-friendly string representation of open files


piskvorky:
Currently, an open S3 object prints like this:

```python
>>> f = smart_open.open('s3://…')
>>> print(f)
<smart_open.s3.SeekableBufferedInputBase object at 0x10ad02b38>
```

This is not useful, and we can do much better: [we have full info](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L329) about the concrete bucket, key, mode, etc. Compare to boto3's nice native display:

```python
>>> print(f._object)
s3.Object(bucket_name='some-bucket', key='test/bank_form.pdf')
```

Task: render open objects more nicely in `__str__`. This is **great help during logging and debugging** (I consider a human-friendly `str()` best practice, and define them religiously on custom classes).

It seems easy enough to do for S3, but I'm not sure how much work for our other supported storages (HTTP, FTP or whatever), for consistency.

mpenkov:
We could implement `__str__` and `__repr__` for our custom classes, which would solve the problem partially.

What should we do for non-custom wrapper classes? e.g.

```python
import sys
from smart_open import open

print('str read_text: ' + str(open(sys.argv[1])))
```

gives

```
str read_text: <encodings.utf_8.StreamReader object at 0x7fa0b1c14eb8>
```

We don't really have control of that object. What's the best thing to do there?

piskvorky:
I'd leave it as it is. "Falling back to built-in where possible" is a well-communicated property of smart_open.

ZlatSic:
Hi!

I'm new to OSS and am looking to contribute somewhere, mind if I give this a shot?

Just to clarify, the behavior you're looking for is to related to the `__str__` and `__repr__` methods of objects created with the `s3://...` descriptor? Which properties would you like to be shown in the representations? Just the `bucket` and `key` or additionally, some of the other, optional properties?

mpenkov:
Closed via #359

PR #359, opened 2019-10-01
smart_open/s3: Initial implementations of str and repr

ZlatSic:
Initial implementations of str and repr for SeekableBufferedInputBase and BufferedOutputBase.

This wasn't too difficult to implement, so I've gone and made the first version before getting the answers to the [questions at the issue thread](https://github.com/RaRe-Technologies/smart_open/issues/316#issuecomment-537198615). Please let me know if you'd like these outputs to look differently, if I should add other parameters to them or if I should add these to any other objects.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -412,6 +412,14 @@ class SeekableBufferedInputBase(BufferedInputBase):
         """Unsupported."""
         raise io.UnsupportedOperation
 
+    def __str__(self):
+        return "s3.SeekableBufferedInputBase(bucket_name='{}', key='{}')".format(
+            self._object.bucket_name, self._object.key)
+
+    def __repr__(self):
+        return "s3.SeekableBufferedInputBase(bucket_name='{}', key='{}')".format(
+            self._object.bucket_name, self._object.key)
+
 
 class BufferedOutputBase(io.BufferedIOBase):
     """Writes bytes to S3.
@@ -553,6 +561,13 @@ multipart upload may fail")
         else:
             self.close()
 
+    def __str__(self):
+        return "s3.BufferedOutputBase(bucket_name='{}', key='{}')".format(
+            self._object.bucket_name, self._object.key)
+
+    def __repr__(self):
+        return "s3.BufferedOutputBase(bucket_name='{}', key='{}')".format(
+            self._object.bucket_name, self._object.key)
 
 def iter_bucket(bucket_name, prefix='', accept_key=None,
                 key_limit=None, workers=16, retries=3):
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -424,7 +424,7 @@ class SeekableBufferedInputBase(BufferedInputBase):
             "bucket_name=%r, "
             "key=%r, "
             "version_id=%r, "
-            "line_terminator=%r, "
+            "line_terminator=%r)"
         ) % (
             self._object.bucket_name,
             self._object.key,
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -352,6 +352,11 @@ class SeekableBufferedInputBase(BufferedInputBase):
 
     def __init__(self, bucket, key, version_id=None, buffer_size=DEFAULT_BUFFER_SIZE,
                  line_terminator=BINARY_NEWLINE, session=None, resource_kwargs=None):
+
+        self._buffer_size = buffer_size
+        self._session = session
+        self._resource_kwargs = resource_kwargs
+
         if session is None:
             session = boto3.Session()
         if resource_kwargs is None:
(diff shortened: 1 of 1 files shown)


smart_open/s3.py
```diff
@@ -412,6 +412,14 @@ def truncate(self, size=None):
         """Unsupported."""
         raise io.UnsupportedOperation
 
+    def __str__(self):
+        return "s3.SeekableBufferedInputBase(bucket_name='{}', key='{}')".format(
```
mpenkov: Can you please use hanging indents?

```python
return Foo(
    bar,
    baz,
    boz,
)
```

smart_open/s3.py
```diff
@@ -412,6 +412,14 @@ def truncate(self, size=None):
...
+            self._object.bucket_name, self._object.key)
+
+    def __repr__(self):
+        return "s3.SeekableBufferedInputBase(bucket_name='{}', key='{}')".format(
```
mpenkov: I think it makes sense to include the full package name here. It'll be helpful for people outside of smart_open: they'll know what package the class comes from.

```suggestion
        return "smart_open.s3.SeekableBufferedInputBase(bucket_name='{}', key='{}')".format(
```

smart_open/s3.py
```diff
@@ -553,6 +561,13 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         else:
             self.close()
 
+    def __str__(self):
+        return "s3.BufferedOutputBase(bucket_name='{}', key='{}')".format(
```
mpenkov: str should prioritize brevity:

```suggestion
        return "smart_open.s3.BufferedOutputBase(%r, %r)" % (self._object.bucket_name, self._object.key)
```

Also please avoid using the .format() stuff (we don't like it).

smart_open/s3.py
```diff
@@ -553,6 +561,13 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            self._object.bucket_name, self._object.key)
+
+    def __repr__(self):
+        return "s3.BufferedOutputBase(bucket_name='{}', key='{}')".format(
```
mpenkov: Please check out [this link](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) for an explanation of the difference between str and repr.

My understanding is that with repr, the user should be able to copy-paste the result into the Python interpreter and get back a copy of the original object. This may not always be possible, but this is something we should aim for. So, please include all the parameters passed to the constructor.

ZlatSic: Thank you for the feedback!

I've introduced most of the suggested changes, but I'd say that this one is still somewhat incomplete. As you said, sometimes these constructor parameters are not available and here we have a similar case where the values passed to the constructor are used to create other objects and the values themselves aren't stored anywhere.

I could try digging through the implementations of these other objects to find the values, but that seems hacky. Would it be OK if these parameters were stored in SeekableBufferedInputBase or BufferedOutputBase objects as in `self._<parameter name>`?

smart_open/s3.py
```diff
@@ -412,6 +412,26 @@ def truncate(self, size=None):
         """Unsupported."""
         raise io.UnsupportedOperation
 
+    def __str__(self):
+        return ("smart_open.s3.SeekableBufferedInputBase(%r, %r)") % (
```
mpenkov: First set of parens not necessary here.

```suggestion
        return "smart_open.s3.SeekableBufferedInputBase(%r, %r)" % (
```

smart_open/s3.py
```diff
@@ -412,6 +412,26 @@ def truncate(self, size=None):
...
+            self._object.bucket_name,
+            self._object.key
+        )
+
+    def __repr__(self):
+        return (
+            "smart_open.s3.SeekableBufferedInputBase("
```
mpenkov: You're not closing the parenthesis on this line anywhere.

smart_open/s3.py
mpenkov: Sure, I think it makes sense to store them.

smart_open/s3.py
```diff
@@ -553,6 +573,24 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         else:
             self.close()
 
+    def __str__(self):
+        return "smart_open.s3.BufferedOutputBase(%r, %r)" % (
+            self._object.bucket_name,
+            self._object.key
```
```suggestion
            self._object.key,
```

But really, I think the above should fit on one line and still be within limits.

smart_open/s3.py
```diff
@@ -412,6 +417,29 @@ def truncate(self, size=None):
...
+            "session=%r, "
+            "resource_kwargs=%r)"
+        ) % (
+            self._object.bucket_name,
+            self._object.key,
+            self._version_id,
+            self._buffer_size,
+            self._line_terminator,
+            self._session,
+            self._resource_kwargs
```
mpenkov: Please add commas to multiline expressions.

```suggestion
            self._resource_kwargs,
```

This way, if we want to add new parameters later, we only need to change one line, not two.

ZlatSic:
Great!

Thanks for all the feedback, hope it wasn't too annoying going back and forth with the changes 😁

mpenkov:
No problem at all.

Hope you enjoyed the experience, if there's anything we do to improve things for you as contributor, let us know, we'll take it on board.

PR #361, opened 2019-10-08
Support writing any bytes-like object to S3.

### LINKED ISSUE #360: Can't write any bytes-like object to S3
`open` built-in lets you use `write` method on any object as long as it implements [Buffer Protocol](https://docs.python.org/3.7/c-api/buffer.html#bufferobjects). However, `smart_open` restricts writeable types to only [`bytes`, `bytearray` and `memoryview`](https://github.com/RaRe-Technologies/smart_open/blob/d65819014536c111f66ef30c3f232f77ee115eda/smart_open/s3.py#L547). Is there a specific reason behind that? I'm using `smart_open` to upload to S3 files that are created using [`pyarrow`](https://arrow.apache.org/docs/python/). PyArrow has a (new?) type called `Buffer` which implements Buffer Protocol. Therefore, when I try to write it through `smart_open`, it just crashes while it just works with built-in `open` function.

I can easily fix the crash by skipping the type check in `smart_open` which seem to work flawlessly in my case. See #361 

Would you be confident with such change? Not sure why the `isinstance` check was added in the first place. The error thrown by Python is already very explicit:

```
TypeError: a bytes-like object is required, not 'xxx'
```

gilbsgilbs:
Fixes #360

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -536,6 +536,8 @@ multipart upload may fail")
         """Write the given buffer (bytes, bytearray, memoryview or any buffer
         interface implementation) to the S3 file.
 
+        For more information about buffers, see https://docs.python.org/3/c-api/buffer.html
+
         There's buffering happening under the covers, so this may not actually
         do any HTTP transfer right away."""
 
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -10,7 +10,6 @@ import warnings
 import boto3
 import botocore.client
 import six
-import sys
 
 import smart_open.bytebuffer
 
@@ -35,11 +34,6 @@ READ_BINARY = 'rb'
 WRITE_BINARY = 'wb'
 MODES = (READ_BINARY, WRITE_BINARY)
 """Allowed I/O modes for working with S3."""
-
-_BINARY_TYPES = (six.binary_type, bytearray)
-"""Allowed binary buffer types for writing to the underlying S3 stream"""
-if sys.version_info >= (2, 7):
-    _BINARY_TYPES = (six.binary_type, bytearray, memoryview)
 
 BINARY_NEWLINE = b'\n'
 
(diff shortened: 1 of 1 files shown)


mpenkov:
All the tests pass, so I guess that check was not necessary. Good catch!

Could you please update the docstring of that method and mention the Buffer interface?

Is it possible to write a unit test for this without including PyArrow?

piskvorky:
I don't remember that "checking code" that was removed here. Can you please track down the commit / PR that added it, to see if it gives the rationale?

Just to make sure we're not regressing somewhere.

gilbsgilbs:
> I don't remember that "isinstance checking code" that was removed here. Can you please track down the commit / PR that added it, to see if it gives its rationale?

It comes from this commit which I don't think is relevant: https://github.com/RaRe-Technologies/smart_open/commit/5f3b6fa8#diff-6d8b12a1a0fab36149bf68fc0ef770faR315

It was then improved in #293 to support memoryviews.

> Is it possible to write a unit test for this without including PyArrow?

Not sure how :thinking: . I think I'd have to implement the buffer interface in a C extension somehow which sounds overkill. With pyarrow, it'd be quite easy to reproduce, still you'd have to download loads of wheels or even compile it yourself which also sounds overkill for the tests.

```python
import pyarrow as pa
import smart_open

buf = pa.py_buffer(b'foobar')

with open('raw', 'wb') as f:
  f.write(buf)  # Works

with smart_open.open('s3://…', 'wb') as f:
  f.write(buf)  # Crashes
```

> Could you please update the docstring of that method and mention the Buffer interface?

:heavy_check_mark:

gilbsgilbs:
@piskvorky @mpenkov anything else I can do?

smart_open/s3.py
```diff
@@ -539,22 +533,19 @@ def detach(self):
         raise io.UnsupportedOperation("detach() not supported")
 
     def write(self, b):
-        """Write the given bytes (binary string) to the S3 file.
+        """Write the given buffer (bytes, bytearray, memoryview or any buffer
+        interface implementation) to the S3 file.
```
mpenkov: Can you please include a link to the appopriate standard library documentation?

Is it this one? https://docs.python.org/3/c-api/buffer.html

gilbsgilbs: Yes, that's the right one.

mpenkov:
Congrats on your first contribution to smart_open @gilbsgilbs 🥇 !!!

gilbsgilbs:
Thanks for the prompt review and merge! May I ask when you expect to release it? This is quite blocking for me actually.

mpenkov:
We will try to release before the end of October.

Issue #259, opened 2019-02-17
List bucket - provide credentials


botzill:
Hi.

IS it possible to provide custom credentials when listing bucket? I see here https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L537 there is no way to provide params. 

Thx.

mpenkov:
Currently, it is not possible. You have to ensure that boto3 picks up your credentials via is default mechanisms (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#method-parameters).

However, I think it would be relatively simple to implement in a PR. You'd need to pass the session parameters as kwargs to the s3_iter_bucket function. Are you able to make such a PR?

Iamshankhadeep:
@mpenkov can I work on this issue?

mpenkov:
Yes, please.

Iamshankhadeep:
@mpenkov  Can you tell me the file I need to edit? I am fairly new on this any help will be appreciated.

mpenkov:
Closed via https://github.com/RaRe-Technologies/smart_open/pull/372

Issue #366, opened 2019-10-25
s3_iter_bucket for non default profile_name


dilzeem:
Hi, 

I am trying to use s3_iter_bucket with the non default profile name.

Currently when I try to do it, I get a 403 Error.

``` python

import boto3
boto3.setup_default_session(profile_name='non-default-profile', region_name='eu-west-1')


from smart_open import s3_iter_bucket, open

bucket = 'some_bucket'
prefix = 'some_prefix'
key = 'some_prefix' + 'some_key'


session = boto3.DEFAULT_SESSION
client = session.client('s3')

# this works
some_object = client.get_object(Bucket=bucket, Key=key)

# this works
same_object = open('s3://' +  bucket + '/' + key, transport_params=dict(session=session))

# get access denied when I try do iterate over this
s3bucket_objs = s3_iter_bucket(bucket, prefix=prefix)

```

mpenkov:
Please show the full exception with the stack trace.

piskvorky:
+ your `smart_open` version.

@mpenkov how about adding an issue template, similar to the one we have in gensim?

dilzeem:
I am using smart_open version 1.8.4
Windows 10 OS


This is what I get when I do `next(s3bucket_objs)`:

```
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "c:\Users\sivaharan.rajkumar\AppData\Local\Continuum\anaconda3\lib\multiprocessing\pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "c:\Users\sivaharan.rajkumar\AppData\Local\Continuum\anaconda3\lib\site-packages\smart_open\s3.py", line 660, in _download_key
    content_bytes = _download_fileobj(bucket, key_name)
  File "c:\Users\sivaharan.rajkumar\AppData\Local\Continuum\anaconda3\lib\site-packages\smart_open\s3.py", line 677, in _download_fileobj
    bucket.download_fileobj(key_name, buf)
  File "c:\Users\sivaharan.rajkumar\AppData\Local\Continuum\anaconda3\lib\site-packages\boto3\s3\inject.py", line 720, in bucket_download_fileobj
    Callback=Callback, Config=Config)
  File "c:\Users\sivaharan.rajkumar\AppData\Local\Continuum\anaconda3\lib\site-packages\boto3\s3\inject.py", line 678, in download_fileobj
    return future.result()
[...]

PR #372, opened 2019-10-31
Added credentials option for iter_bucket.

derpferd:
This closes issue #259

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -686,10 +686,10 @@ def iter_bucket(
 
     total_size, key_no = 0, -1
     key_iterator = _list_bucket(
-            bucket_name,
-            prefix=prefix,
-            accept_key=accept_key,
-            **session_kwargs)
+        bucket_name,
+        prefix=prefix,
+        accept_key=accept_key,
+        **session_kwargs)
     download_key = functools.partial(
         _download_key,
         bucket_name=bucket_name,
@@ -718,7 +718,8 @@ def _list_bucket(
         prefix='',
         accept_key=lambda k: True,
         **session_kwargs):
-    client = boto3.client('s3', **session_kwargs)
+    session = boto3.session.Session(**session_kwargs)
+    client = boto3.client('s3')
     ctoken = None
 
     while True:
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -614,7 +614,9 @@ def _accept_all(key):
 
 
 def iter_bucket(bucket_name, prefix='', accept_key=None,
-                key_limit=None, workers=16, retries=3):
+                key_limit=None, workers=16, retries=3,
+                aws_access_key_id=None, aws_secret_access_key=None,
+                aws_session_token=None):
     """
     Iterate and download all S3 objects under `s3://bucket_name/prefix`.
 
@@ -634,6 +636,13 @@ def iter_bucket(bucket_name, prefix='', accept_key=None,
         The number of subprocesses to use.
     retries: int, optional
         The number of time to retry a failed download.
+    aws_access_key_id: str, optional
+        The access key to use when creating the AWS client.
+    aws_secret_access_key: str, optional
+        The secret key to use when creating the AWS client.
+    aws_session_token: str, optional
+        The session token to use when creating the AWS client.
+
 
     Yields
     ------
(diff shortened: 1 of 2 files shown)


smart_open/s3.py
```diff
@@ -614,7 +614,9 @@ def _accept_all(key):
 
 
 def iter_bucket(bucket_name, prefix='', accept_key=None,
-                key_limit=None, workers=16, retries=3):
+                key_limit=None, workers=16, retries=3,
```
mpenkov: I think passing in these parameters as three separate things is a bit ugly. This is especially true because we now do this in like 3-4 different functions.

Is it possible to pass around a boto3.Session instead?

smart_open/tests/test_s3.py
```diff
@@ -460,6 +460,30 @@ def test_list_bucket_long(self):
         expected = ['key_%d' % x for x in range(num_keys)]
         self.assertEqual(sorted(keys), sorted(expected))
 
+    @mock.patch('boto3.client')
+    @mock.patch('boto3.session.Session')
+    def test_credentials(self, mock_session, mock_client):
+        # setup mocks
```
mpenkov: This is a useless comment. Please get rid of it.

smart_open/tests/test_s3.py
```diff
@@ -460,6 +460,30 @@ def test_list_bucket_long(self):
...
+        mock_s3 = mock.Mock()
+        mock_client.return_value = mock_s3
+        mock_s3.list_objects_v2.return_value = {'Contents': [{'Key': 'test_key'}]}
+
+        # make call
+        # Note we disable multiprocessing since mocks only work in the thread.
+        list(smart_open.s3.iter_bucket(BUCKET_NAME,
```
mpenkov: Please use a hanging indent, here and everywhere.

smart_open/tests/test_s3.py
```diff
@@ -460,6 +460,30 @@ def test_list_bucket_long(self):
...
+                                       workers=None,
+                                       aws_access_key_id='access_id',
+                                       aws_secret_access_key='access_secret'))
+
+        # asserts
```
mpenkov: Please remove this. It's obvious what you're doing from the source code, so the comment adds zero value.

smart_open/tests/test_s3.py
```diff
@@ -460,6 +460,30 @@ def test_list_bucket_long(self):
...
 
+    @mock.patch('boto3.client')
+    @mock.patch('boto3.session.Session')
+    def test_credentials(self, mock_session, mock_client):
+        # setup mocks
+        mock_s3 = mock.Mock()
+        mock_client.return_value = mock_s3
+        mock_s3.list_objects_v2.return_value = {'Contents': [{'Key': 'test_key'}]}
+
+        # make call
```
mpenkov: "make call" is irrelevant here. The line after it _is_ relevant, however.

smart_open/s3.py
mpenkov: If we pass a Session, we can also close #366, which will be a big win.

derpferd: Per https://github.com/boto/botocore/issues/1246 it would appear that it is not recommended that we can pass boto3.Session around. (Whatever recommended means...)

I totally agree passing separate things around is ugly, but don't know if we should take this risk.

smart_open/tests/test_s3.py
derpferd: I was just following the style I saw used in `iter_bucket`'s function definition. Refactored the function calls to use hanging indent. Would you like the function definitions refactored as well?

smart_open/s3.py
mpenkov: > It is recommended to create a resource instance for each thread / process in a multithreaded or multiprocess application rather than sharing a single instance among the threads / processes.

OK, I see. We need to pass session information across a thread/process boundary. In that case, we shouldn't pass around a single session instance.

Do this instead:

```python
def iter_bucket(..., **session_kwargs):
    """
    session_kwars: dict, optional
        Keyword arguments to pass when creating a new session.
        For a list of available names and values, see:
        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session
    """
    ...

def list_bucket(..., **session_kwargs):
    client = boto3.client('s3', **session_kwargs)

```

smart_open/s3.py
```diff
@@ -694,8 +709,13 @@ def iter_bucket(bucket_name, prefix='', accept_key=None,
     logger.info("processed %i keys, total size %i" % (key_no + 1, total_size))
 
 
-def _list_bucket(bucket_name, prefix='', accept_key=lambda k: True):
-    client = boto3.client('s3')
+def _list_bucket(bucket_name, prefix='', accept_key=lambda k: True,
+                 aws_access_key_id=None, aws_secret_access_key=None,
+                 aws_session_token=None):
+    client = boto3.client('s3',
```
mpenkov: Hanging indent here please.

smart_open/s3.py
```diff
@@ -720,14 +740,18 @@ def _list_bucket(bucket_name, prefix='', accept_key=lambda k: True):
...
+                  aws_access_key_id=None, aws_secret_access_key=None,
+                  aws_session_token=None):
     if bucket_name is None:
         raise ValueError('bucket_name may not be None')
 
     #
     # https://geekpete.com/blog/multithreading-boto3/
     #
-    session = boto3.session.Session()
+    session = boto3.session.Session(aws_access_key_id=aws_access_key_id,
```
mpenkov: ... and here.

smart_open/tests/test_s3.py
mpenkov: Yes, please make sure any _new_ code using hanging indents.

Don't worry about reformatting code you have not changed.

mpenkov:
@derpferd Ping. Are you able to finish this PR?

derpferd:
@mpenkov Sorry for the delay, made the formatting suggestion and changed to use key word args instead.

mpenkov:
Merged. @derpferd Thank you for your effort!

PR #377, opened 2019-11-05
ssh/scp/sftp: enable pipelining of write operations

mrk-its:
#### Motivation

It significantly speeds up writes. From paramiko docs:

> When pipelining is on, paramiko won’t wait for the server response after each write operation. Instead, they’re collected as they come in. At the first non-write operation (including close), all remaining server responses are collected. This means that if there was an error with one of your later writes, an exception might be thrown from within close instead of write.

Popular sftp client library **pysftp** enables pipelining for put operations.

Result of simple [benchmark](https://gist.github.com/mrk-its/04c3f0430ac4e548dfecba6c3a8a41be)

pipelining disabled:

> Total time: 33.55 seconds, sent: 2621440 bytes, speed: 76 kbytes/sec

pipelining enabled:

> Total time: 3.58 seconds, sent: 2621440 bytes, speed: 714 kbytes/sec

### DIFF
--- a/smart_open/ssh.py
+++ b/smart_open/ssh.py
@@ -107,6 +107,7 @@ def open(path, mode='r', host=None, user=None, password=None, port=DEFAULT_PORT,
     conn = _connect(host, user, port, password, transport_params)
     sftp_client = conn.get_transport().open_sftp_client()
     file_obj = sftp_client.open(path, mode)
-    if mode.startswith("w"):
+    is_pipelining_enabled = transport_params.get('write_pipelining', True)
+    if is_pipelining_enabled and mode.startswith("w"):
         file_obj.set_pipelined(True)
     return file_obj
--- a/smart_open/ssh.py
+++ b/smart_open/ssh.py
@@ -106,4 +106,7 @@ def open(path, mode='r', host=None, user=None, password=None, port=DEFAULT_PORT,
 
     conn = _connect(host, user, port, password, transport_params)
     sftp_client = conn.get_transport().open_sftp_client()
-    return sftp_client.open(path, mode)
+    file_obj = sftp_client.open(path, mode)
+    if mode.startswith("w"):
+        file_obj.set_pipelined(True)
+    return file_obj
--- a/smart_open/tests/test_ssh.py
+++ b/smart_open/tests/test_ssh.py
@@ -41,6 +41,25 @@ class SSHOpen(unittest.TestCase):
         )
         mock_connect.assert_called_with("some-host", 22, username="ubuntu", password="pwd")
 
+    @mock_ssh
+    def test_write_pipelining_for_reading(self, mock_connect, get_transp_mock):
+        smart_open.open("ssh://user:pass@some-host/", "r")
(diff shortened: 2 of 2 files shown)


mrk-its:
@mpenkov could you take a look?

smart_open/ssh.py
```diff
@@ -106,4 +106,8 @@ def open(path, mode='r', host=None, user=None, password=None, port=DEFAULT_PORT,
 
     conn = _connect(host, user, port, password, transport_params)
     sftp_client = conn.get_transport().open_sftp_client()
-    return sftp_client.open(path, mode)
+    file_obj = sftp_client.open(path, mode)
+    is_pipelining_enabled = transport_params.get('write_pipelining', True)
```
mpenkov: You shouldn't rely on `transport_params` here. That's a poorly named parameter here, it should really be called `connect_params` or something, because it's only for the connect function.

You should add a new keyword parameter `write_pipelining` and document it in the docstring.

mpenkov:
Thank you for your contribution. Sorry it has taken so long for me to review this.

Please have a look at the comments and get back to me.

smart_open/ssh.py
mrk-its: I don't get it. ssh.connect keyword args are passed through 'connect_kwargs' key in `transport_params` and this is IMHO fine - this allows for introducing other parameters directly in transport_params without affecting ssh.connect arguments (like this `write_pipelining` parameter). I can add `write_pipelining` keyword parameter to ssh.open function signature and initialize it while calling ssh.open (taking value from transport_params dict), but I don't see any advantage of that (this will introduce more backend-related code to smart_open_lib). Could you describe more precisely how do you see it?

mpenkov: The advantage is discovery of your new feature. If you add a new parameter, people will find out about it via the doc string. Otherwise, no-one will know about it.

The existing handling of transport Params by the ssh submodule is broken (look at the current docstring). We will have to fix that in a separate PR.

mrk-its: > The advantage is discovery of your new feature. If you add a new parameter, people will find out about it via the doc string. Otherwise, no-one will know about it.
> 
So maybe backend should be called like `ssh.open(..., **transport_params)`? This way we can have separate keyword parameters in backend's open function (with docstring documentation) and don't have this backend-specific parameter handling on upper level.

> The existing handling of transport Params by the ssh submodule is broken (look at the current docstring). We will have to fix that in a separate PR.

I thought it is implemented as you described here: https://github.com/RaRe-Technologies/smart_open/pull/292#issuecomment-486478592

mpenkov: > So maybe backend should be called like `ssh.open(..., **transport_params)`?

Yes, this.

mpenkov:
@mrk-its Ping. Are you able to finish this PR?

mrk-its:
> @mrk-its Ping. Are you able to finish this PR?

Yes, sure, I'll get back soon with changes.

mpenkov:
Closing as stale

PR #379, opened 2019-11-07
delay s3 request until read

### LINKED ISSUE #362: s3 open seek operation try read rest of file into buffer, which makes following read has timeout risk
[Here](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L200) makes API call to fetch rest of file into buffer when calling `seek`, which makes `seek` very slow. The API call may be put in `read` method instead.

Gapex:
#### Motivation
- Fixes #362

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,6 +91,7 @@ Other examples of URLs that ``smart_open`` accepts::
     file:///home/user/file.bz2
     [ssh|scp|sftp]://username@host//path/file
     [ssh|scp|sftp]://username@host/path/file
+    [ssh|scp|sftp]://username:password@host/path/file
 
 .. _doctools_after_examples:
 
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,8 @@ tests_require = [
     # 'botocore < 1.11.0'
     # Not used directly but allows boto GCE plugins to load.
     # https://github.com/GoogleCloudPlatform/compute-image-packages/issues/262
-    'google-compute-engine==2.8.12'
+    'google-compute-engine==2.8.12',
+    'paramiko',
 ]
 
 install_requires = [
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -30,7 +30,6 @@ import boto
 import boto3
 import six
 
-from boto.compat import urlsplit
 from six.moves.urllib import parse as urlparse
 
 #
@@ -129,6 +128,7 @@ Uri = collections.namedtuple(
         'access_id',
         'access_secret',
         'user',
+        'password',
     )
 )
 """Represents all the options that we parse from user input.
@@ -554,6 +554,8 @@ def _open_binary_stream(uri, mode, transport_params):
                 host=parsed_uri.host,
                 user=parsed_uri.user,
                 port=parsed_uri.port,
+                password=parsed_uri.password,
+                transport_params=transport_params,
             )
             return fobj, filename
         elif parsed_uri.scheme in smart_open_s3.SUPPORTED_SCHEMES:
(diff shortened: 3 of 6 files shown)


vuryleo:
i think adding some test would be better

smart_open/s3.py
```diff
@@ -170,44 +172,58 @@ def __init__(self, s3_object, content_length, version_id=None):
         self._object = s3_object
         self._content_length = content_length
         self._version_id = version_id
+        self._body_reload = True
```
mpenkov: Do you really need this variable?

Isn't is possible to set `self._body = None`, and then expect `self._body is None` to mean the same thing as `self._body_reload is True`?

smart_open/s3.py
```diff
@@ -170,44 +172,58 @@ def __init__(self, s3_object, content_length, version_id=None):
...
+        if self._body_reload:
+            self._load_body()
+
+        try:
+            if size == -1:
+                binary = self._body.read()
+            else:
+                binary = self._body.read(size)
+        except IncompleteReadError:
+            # the underlying connection of streaming body was closed by peer
```
mpenkov: Minor English fixup.

```suggestion
            # The underlying connection of the streaming body was closed by the remote peer.
```

smart_open/s3.py
```diff
@@ -170,44 +172,58 @@ def __init__(self, s3_object, content_length, version_id=None):
...
+            # close the current one explicitly, prepare to open a new and try to read again
```
```suggestion
            # Close the current connection explicitly, prepare to open a new one and try to read again.
```

smart_open/s3.py
```diff
@@ -170,44 +172,58 @@ def __init__(self, s3_object, content_length, version_id=None):
...
+            self._body.close()
+            self._body_reload = True
+            return self.read(size)
```
mpenkov: You're setting up an infinite recursion here, in theory.

Can you think of a way to break the cycle? We could count the number of consecutive IncompleteReadErrors, and then bail out if we see more than e.g. 10.

smart_open/s3.py
Gapex: you're right.

smart_open/s3.py
Gapex: What about adding a member field to count the number of retry, when counter less than the max limit, retry, other wise, terminate the action by raising an error?

mpenkov: That's OK, but you need to make sure you're counting _consecutive_ failures, as opposed to total failures. So, you reset the counter to zero on each successful attempt.

mpenkov:
Looks like your tests are failing because you're dereferencing `None`, please have a look at the buildbot output (https://travis-ci.org/RaRe-Technologies/smart_open/jobs/610158039?utm_medium=notification&utm_source=github_status).

smart_open/s3.py
```diff
@@ -199,13 +201,30 @@ def seek(self, position):
...
         else:
             binary = self._body.read(size)
+        return binary
+
+    def read(self, size=-1):
+        if self._position >= self._content_length:
+            return b''
+
+        if self._body is None:
+            # After seek, or the first read, the self._body is None.
```
vuryleo: how about closed?

smart_open/s3.py
```diff
@@ -199,13 +201,30 @@ def seek(self, position):
...
+            self._load_body()
+
+        try:
+            binary = self._read_from_body(size)
+        except IncompleteReadError:
+            # The underlying connection of the streaming body was closed by the remote peer.
+            self._body.close()
```
vuryleo: do we need this?

smart_open/s3.py
Gapex: After init, the self._body is None, the connection is not built until the first read.

vuryleo: So you can still read from a `closed` reading stream, interesting. Is this work as designed?

smart_open/s3.py
Gapex: I will delete the line.

smart_open/s3.py
Gapex: Yes, we could read from a closed stream, even if after the `close()` operation, but there is no such api offer to users. 
In addition, according to the botocore [reference](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/response.html#botocore-response), streaming body doesn't have the `closed` attribute.

Gapex:
@mpenkov , When executing `python setup.py test`, read from a closed streaming body always returns 0 bytes instead of raising an IncompleteReadError, thus the action of reopening an new streaming body had not been triggered. Thus, the SeekableBufferedInputBase thinks the raw reader reaches it's EOF, and just returns the rest data in the buffer.

This makes the unit test  be failed.

mpenkov:
OK, some questions:

- Who returns 0 bytes when reading from a closed stream?
- What depends on the existing behavior (return 0 bytes)?
- The correct behavior is to raise an exception, right?

We probably want to fix this, as it deviates from the behavior of file streams in the stdlib.

mpenkov:
@Gapex What happened here? Why did you close the PR?

Gapex:
Yes, I will open a new pull request.

PR #389, opened 2019-11-13
Fixes "the connection was closed by the remote peer" error

### LINKED ISSUE #362: s3 open seek operation try read rest of file into buffer, which makes following read has timeout risk

Gapex:
#### Motivation
- Fixes #362 

The current SeekableRawReader sends HTTP request in the [`seek()`](https://github.com/RaRe-Technologies/smart_open/blob/21d98e6a7144191c24ceca862e32910ff740c933/smart_open/s3.py#L200) method, and keeps an continuous connection with the remote peer. But if the SeekableRawReader does not read from the connection immediately after `seek()`, the remote peer may close the connection due to a timeout. Then, the SeekableRawReader attempts to read on a StreamingBody whose underlying connection had been closed, and gets a [botocore.exceptions.IncompleteReadError](), the following example shows the case:

```python
from boto3 import resource

endpoints = [
    'http://xxxxxxxxxxxxx', # real s3
    'http://localhost:8000' # moto s3 server
]

real_s3_resource = resource('s3', endpoint_url=endpoints[0])
moto_s3_resource = resource('s3', endpoint_url=endpoints[1])

# Test read from a closed streaming body.

# For real s3.
obj = real_s3_resource.Object('gyp', 'images.msg')
body = obj.get()['Body']
print(type(body)) # <class 'botocore.response.StreamingBody'>
print(body.read(1)) # b'\xdd'
body.close() # Close the body.
try:
    body.read(1)
except Exception as e:
    print(type(e)) # <class 'botocore.exceptions.IncompleteReadError'>


# For moto s3 server.
moto_s3_resource.Bucket('gyp').create()
moto_s3_resource.Object('gyp', 'test-key').put(Body=b'123456')
obj = moto_s3_resource.Object('gyp', 'test-key')
body = obj.get()['Body']
print(type(body)) # <class 'botocore.response.StreamingBody'>
print(body.read(1)) # b'1'
body.close() # Close the body.
try:
    body.read(1)
except Exception as e:
    print(type(e)) # <class 'botocore.exceptions.IncompleteReadError'>
```

In the above code, the `body` is closed by client actively to stimulate connection closed by the remote peer.

In the [PR](https://github.com/RaRe-Technologies/smart_open/pull/379), I intended to delay the action of sending HTTP request until read, the solution solved the problem of closed connection between a pair of `seek() - read()`, but introduced a similar problem between continuous `read()`, namely, the connection may also be closed by the remote peer after a read, if the second read occurs a long time later.

In this PR, I try to catch the IncompleteReadError in the `read()` directly, if connection is close, and IncompleteReadError is raised,
1. Catch the exception and dereference the current StreamingBody, 
2. Call the `seek()` from the current offset to build a new connection with remote peer. 
3. Finally, read from the new connection as required.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -182,7 +182,7 @@ class SeekableRawReader(object):
         """
         #
         # Close old body explicitly.
-        # When first seek(), self._body is not exist. Catch the exception and do nothing.
+        # When first seek() after __init__(), self._body is not exist.
         #
         if self._body is not None:
             self._body.close()
@@ -190,6 +190,8 @@ class SeekableRawReader(object):
         self._position = position
 
     def _load_body(self):
+        """Build a continuous connection with the remote peer starts from the current postion.
+        """
         range_string = make_range_string(self._position)
         logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
 
@@ -210,9 +212,11 @@ class SeekableRawReader(object):
         return binary
 
     def read(self, size=-1):
+        """Read from the continuous connection with the remote peer."""
         if self._position >= self._content_length:
             return b''
         if self._body is None:
+            # When the first read() after __init__() or seek(), self._body is not exist.
             self._load_body()
 
         try:
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -172,27 +172,28 @@ class SeekableRawReader(object):
         self._object = s3_object
         self._content_length = content_length
         self._version_id = version_id
-        self.seek(0)
+        self._position = 0
+        self._body = None
 
     def seek(self, position):
         """Seek to the specified position (byte offset) in the S3 key.
 
         :param int position: The byte offset from the beginning of the key.
         """
-        self._position = position
-        range_string = make_range_string(self._position)
-        logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
-
         #
         # Close old body explicitly.
         # When first seek(), self._body is not exist. Catch the exception and do nothing.
         #
-        try:
+        if self._body is not None:
             self._body.close()
-        except AttributeError:
-            pass
+        self._body = None
+        self._position = position
 
-        if position == self._content_length == 0 or position == self._content_length:
+    def _load_body(self):
+        range_string = make_range_string(self._position)
+        logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
+
+        if self._position == self._content_length == 0 or self._position == self._content_length:
             #
             # When reading, we can't seek to the first byte of an empty file.
             # Similarly, we can't seek past the last byte.  Do nothing here.
@@ -211,11 +212,14 @@ class SeekableRawReader(object):
     def read(self, size=-1):
         if self._position >= self._content_length:
             return b''
+        if self._body is None:
+            self._load_body()
+
         try:
             binary = self._read_from_body(size)
         except IncompleteReadError:
-            self._body = None
-            self.seek(self._position)
+            # The underlying connection of the self._body was closed by the remote peer.
+            self._load_body()
             binary = self._read_from_body(size)
         self._position += len(binary)
         return binary
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -19,6 +19,8 @@ import six
 
 import smart_open.bytebuffer
 
+from botocore.exceptions import IncompleteReadError
+
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


Gapex:
@mpenkov 
* The moto uses a file object to mock StreamingBody, and reading from a closed file can not actully raises botocore.exceptions.IncompleteReadError.
* So, I need to skip the [test case](https://github.com/RaRe-Technologies/smart_open/pull/389/files#diff-94d24e00c98a55a75441fbd4732ffc7cR118) in 'Mock Environment'.
* I notice that The [travis](https://travis-ci.org/RaRe-Technologies/smart_open/jobs/611258413#L372) unsets the SO_DISABLE_MOCKS env variable, thus all s3 tests were using moto.
* In order to test the modified part of code, a real connection maybe is needed. 


```python
from boto3 import resource

mocker = moto.mock_s3()
mocker.start()
r = resource('s3')
r.Bucket('bucket').create()
r.Object('bucket', 'key').put(Body=b'123456')
obj = r.Object('bucket', 'key')
body = obj.get()['Body']
print(body.read(1)) # b'1'
body.close()
try:
    print(body.read(1))
except Exception as e:
    print(type(e)) # <class 'ValueError'>
    print(e) # I/O operation on closed file.
mocker.stop()
```

vuryleo:
I would think it's an issue related to moto. Maybe some patch is needed :(

smart_open/tests/test_s3.py
```diff
@@ -104,6 +104,26 @@ def ignore_resource_warnings():
     warnings.filterwarnings("ignore", category=ResourceWarning, message="unclosed.*<ssl.SSLSocket.*>")  # noqa
 
 
+class SeekableRawReaderTest(unittest.TestCase):
+
+    def setUp(self):
+        self._local_resoruce = boto3.resource('s3', endpoint_url='http://localhost:5000')
```
mpenkov: Typo:

```suggestion
        self._local_resource = boto3.resource('s3', endpoint_url='http://localhost:5000')
```

smart_open/tests/test_s3.py
```diff
@@ -104,6 +104,26 @@ def ignore_resource_warnings():
     warnings.filterwarnings("ignore", category=ResourceWarning, message="unclosed.*<ssl.SSLSocket.*>")  # noqa
 
 
+class SeekableRawReaderTest(unittest.TestCase):
```
mpenkov: Hide this test behind an environment variable (skip by default).

We shouldn't expect whoever is running the tests to also have moto on port 5000. It's good if they are, but it's not compulsory.

Gapex: Done. Please take a look.

smart_open/s3.py
```diff
@@ -170,27 +172,30 @@ def __init__(self, s3_object, content_length, version_id=None):
...
+        self._position = position
+
+    def _load_body(self):
+        """Build a continuous connection with the remote peer starts from the current postion.
+        """
+        range_string = make_range_string(self._position)
+        logger.debug('content_length: %r range_string: %r', self._content_length, range_string)
 
-        if position == self._content_length == 0 or position == self._content_length:
+        if self._position == self._content_length == 0 or self._position == self._content_length:
```
mpenkov: I wonder if this can be simplified to:

```suggestion
        if self._position == self._content_length:
```

Because you have a situation where your conditional is `if A or B` and `B` is always true whenever `A` is.

What do you think?

smart_open/tests/test_s3.py
mpenkov: Very good. Thank you.

smart_open/s3.py
Gapex: Yeah, because the `or` is a short circuit operation, the `B` will not be evaluated when the `A` is True. But I would think the verbose judgement is helpful for expressing the behaviour of the class, as the following comment stated.
```python
# 
# When reading, we can't seek to the first byte of an empty file.
# Similarly, we can't seek past the last byte.  Do nothing here.
```

mpenkov: OK, you have a point. Let's leave it as is.

mpenkov:
@Gapex Thank you for your patience! This is a very welcome contribution. Keep them coming!

PR #400, opened 2019-12-08
allow use of S3 single part uploads

adrpar:
I had encountered a use case, where I needed to upload many different files into S3 and wanted to use smart_open. The files were a large set of small and some sporadic large files.

Writing all the files to my S3 bucket proved slow, since smart_open is meant for large file and does thus use multipart upload. This resulted in significant overhead.

I am a huge fan of the smart_open interface and ease of use, and though about how such a use case could be supported in smart_open. 

I fixed my issue, by introducing an optional parameter hinting at the size of the input stream (often this is known) and then added the possibility for the S3 writer to use direct upload if better suited.

This PR is far from done (no handling of what happens if the input stream proves larger than claimed) and I just wanted to propose and start a discussion on such a feature. Further, there are many different ways on how smart_open could be guided to use direct upload. Maybe this is not even wanted...

Therefore, thanks for having a look at this and let's start a discussion and I am happy to adjust my implementation :)

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -515,8 +515,12 @@ multipart upload may fail")
             self._object = s3.Object(bucket, key)
             self._min_part_size = min_part_size
             self._mp = self._object.initiate_multipart_upload(**self._upload_kwargs)
-        except botocore.client.ClientError:
-            raise ValueError('the bucket %r does not exist, or is forbidden for access' % bucket)
+        except botocore.client.ClientError as error:
+            raise ValueError(
+                'the bucket %r does not exist, or is forbidden for access (%r)' % (
+                    bucket, error
+                )
+            )
 
         self._buf = io.BytesIO()
         self._total_bytes = 0
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -125,13 +125,6 @@ def open(
     if mode not in MODES:
         raise NotImplementedError('bad mode: %r expected one of %r' % (mode, MODES))
 
-    if resource_kwargs is None:
-        resource_kwargs = {}
-    if multipart_upload_kwargs is None:
-        multipart_upload_kwargs = {}
-    if object_kwargs is None:
-        object_kwargs = {}
-
     if (mode == WRITE_BINARY) and (version_id is not None):
         raise ValueError("version_id must be None when writing")
 
@@ -152,7 +145,7 @@ def open(
                 key_id,
                 min_part_size=min_part_size,
                 session=session,
-                multipart_upload_kwargs=multipart_upload_kwargs,
+                upload_kwargs=multipart_upload_kwargs,
                 resource_kwargs=resource_kwargs,
             )
         else:
@@ -160,7 +153,7 @@ def open(
                 bucket_id,
                 key_id,
                 session=session,
-                singlepart_upload_kwargs=singlepart_upload_kwargs,
+                upload_kwargs=singlepart_upload_kwargs,
                 resource_kwargs=resource_kwargs,
             )
     else:
@@ -496,10 +489,9 @@ class MultipartWriter(io.BufferedIOBase):
             min_part_size=DEFAULT_MIN_PART_SIZE,
             session=None,
             resource_kwargs=None,
-            multipart_upload_kwargs=None,
+            upload_kwargs=None,
             ):
 
-        self._multipart_upload_kwargs = multipart_upload_kwargs
 
         if min_part_size < MIN_MIN_PART_SIZE:
             logger.warning("S3 requires minimum part size >= 5MB; \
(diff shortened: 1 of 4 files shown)


mpenkov:
Hi, sorry it has taken me so long to get around to this.

I think the motivation behind your PR is good. Single-part uploads make sense under some conditions.

I think the execution of the idea could be a little better. First, we should leave it up to the user to decide how small is too small for a multipart upload, instead of hardcoding a threshold ourselves. So the call to smart_open would look like this instead:

```python
with open(..., 'w', transport_params={'multipart_upload': True}) as fout:
    ...
```

On the implementation side, I see you've gone with conditionals inside the existing class. This is a bit too difficult to maintain, as it complicates the code unnecessarily.

I suggest you implement a separate class to take care of single-part uploading. This can be entirely separate to the existing class. This class would be extremely simple, because you'd be loading the whole single part into memory, and wouldn't need much of the logic of the existing class.

The above approach would hit two birds with one stone:

1. Keep the existing, multi-part class as-is.
2. Satisfy your use case.

Let me know if you're up to it.

adrpar:
Hi, thanks for looking at the PR. 

I will redesign the whole approach and use a separate class as you proposed, no problem. The current way was the simplest solution for solving the use case we had and I was not sure if you would agree to have the feature. Therefore I just wanted to get your feedback and guidance first :)

I would then have the `multipart_upload` transport parameter default to `true` and if set to `false` use the single-part upload class.

I am up for this and hope to have the updated PR ready soon.

adrpar:
I split the single part upload into its own class and hope that I covered the single part upload fully with tests. I am awaiting your comments and hope, that's how you were thinking about how this should be implemented.

smart_open/s3.py
```diff
@@ -100,6 +101,12 @@ def open(
     multipart_upload_kwargs: dict, optional
         Additional parameters to pass to boto3's initiate_multipart_upload function.
         For writing only.
+    multipart_upload: bool, optional
+        Default: `True`
+        If set to `True`, will use multipart upload for writing to S3. If set
+        to `false`, S3 upload will use the S3 Single-Part Upload API, which
```
```suggestion
        to `False`, S3 upload will use the S3 Single-Part Upload API, which
```

smart_open/s3.py
```diff
@@ -127,14 +134,23 @@ def open(
...
+                session=session,
+                multipart_upload_kwargs=multipart_upload_kwargs,
+                resource_kwargs=resource_kwargs,
+            )
+        else:
+            fileobj = BufferedSinglePartOutputBase(
+                bucket_id,
+                key_id,
+                session=session,
+                multipart_upload_kwargs=multipart_upload_kwargs,
```
mpenkov: Why are we passing multi-part arguments to the single-part writer?

smart_open/s3.py
```diff
@@ -479,8 +495,8 @@ def __repr__(self):
         )
 
 
-class BufferedOutputBase(io.BufferedIOBase):
-    """Writes bytes to S3.
+class BufferedMultiPartOutputBase(io.BufferedIOBase):
```
mpenkov: Please call this class MultipartWriter.

smart_open/s3.py
```diff
@@ -658,6 +674,130 @@ def __repr__(self):
         )
 
 
+class BufferedSinglePartOutputBase(io.BufferedIOBase):
```
mpenkov: Please call this class SinglepartWriter.

smart_open/s3.py
```diff
@@ -658,6 +674,130 @@ def __repr__(self):
...
+    """Writes bytes to S3 using the single part API.
+
+    Implements the io.BufferedIOBase interface of the standard library."""
```
mpenkov: You should mention that this class buffers all of its input, and only writes to S3 when `close` is called.

mpenkov:
Please fix flake8 warnings in your code @adrpar

smart_open/s3.py
adrpar: This should be `singlepart_upload_kwargs` to be passed onto the S3.Object.put method. Will change accordingly

adrpar:
Rebased and addressed your comments @mpenkov.

smart_open/s3.py
```diff
@@ -644,6 +665,134 @@ def __repr__(self):
...
+        s3 = session.resource('s3', **resource_kwargs)
+        try:
+            self._object = s3.Object(bucket, key)
+            s3.meta.client.head_bucket(Bucket=bucket)
+        except botocore.client.ClientError:
+            raise ValueError('the bucket %r does not exist, or is forbidden for access' % bucket)
+
+        self._buf = io.BytesIO()
+        self._total_bytes = 0
+        self._closed = False
```
mpenkov: Do we need a separate variable for this? Can we not use `self._object is None` to mean closed, and assume open otherwise.

smart_open/s3.py
```diff
@@ -644,6 +665,134 @@ def __repr__(self):
...
+    This class buffers all of its input in memory until its `close` method is called. Only then will
+    the data be written to S3 and the buffer is released."""
+
+    def __init__(
+            self,
+            bucket,
+            key,
+            session=None,
+            resource_kwargs=None,
+            singlepart_upload_kwargs=None,
```
mpenkov: I think the singlepart thing is a bit redundant here, because this is the SinglepartWriter. `upload_kwargs` is enough.

smart_open/s3.py
```diff
@@ -644,6 +665,134 @@ def __repr__(self):
...
+    def closed(self):
+        return self._closed
+
+    def writable(self):
+        """Return True if the stream supports writing."""
+        return True
+
+    def tell(self):
+        """Return the current stream position."""
+        return self._total_bytes
```
mpenkov: Why don't we use `self._buf.tell()` here? That way, we could get rid of the `_total_bytes` attribute entirely.

mpenkov: Nevermind, let's keep it for symmetry.

mpenkov:
Thank you for the effort @adrpar . I'm merging this.

PR #401, opened 2019-12-25
Add test data in package via MANIFEST.in

jayvdb:
The tests are included in the sdist, but they fail without the data.

Also update docs filenames.


mpenkov:
Merged. Thank you for your work! And congrats on your first contribution to smart_open 🥇

Issue #198, opened 2018-06-12
Support for GCS


Arvinds-ds:
Do you have plans to support the Google Cloud Storage protocol `gcs://`

piskvorky:
Sounds like a good idea to me. No concrete plans yet -- do you want to implement this?

petedannemann:
I am working on adding GCS support and wanted to ask a question before I get too deep into the code. I can create a new issue and repost the question there if this is the wrong place to ask it.

Will future smart-open releases support Python 2.7?

mpenkov:
I don't think so. There is little value of spending effort supporting a language that even the language creators no longer support ([Py2.7 end of life is Jan 1 2020](https://pythonclock.org/)).

That said, the library doesn't depend on the Python version _that_ much, so the effort of maintaining support for 2.7 is probably much less than for e.g. gensim.

Why do you ask?

petedannemann:
GCS support will just require a substantial amount of code to be added (similar to the amount of S3) and I wasn't sure how much I should focus on backwards compatibility. We can revisit this once I've actually completed the code and submitted a PR.

mpenkov:
In general, aim for Py3. It's unlikely you'll finish this before the end of the year, and Py2 will be gone by then.

petedannemann:
Can we close this?

mpenkov:
Yes, thank you for following up.

PR #404, opened 2020-01-05
Google Cloud Storage (GCS)

petedannemann:
#### Motivation
- Adds GCS Support #198 

#### Checklist

We will need to figure out how we plan to deal with integration testing on GCP. Would RaRe be willing to host the bucket? We will need to update Travis to include those tests if so.

EDIT: Removed comment about the testing timeout issue. Since fixing the memory issue with reads, it has gone away.

### DIFF
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 import io
 import os
-import subprocess
+
+import google.cloud.storage
+from six.moves.urllib import parse as urlparse
 
 import smart_open
 
@@ -10,7 +12,14 @@ assert _GCS_URL is not None, 'please set the SO_GCS_URL environment variable'
 
 
 def initialize_bucket():
-    subprocess.check_call(['gsutil' 'rm', '-r', _GCS_URL])
+    client = google.cloud.storage.Client()
+    parsed = urlparse.urlparse(_GCS_URL)
+    bucket_name = parsed.netloc
+    prefix = parsed.path
+    bucket = client.get_bucket(bucket_name)
+    blobs = bucket.list_blobs(prefix=prefix)
+    for blob in blobs:
+        blob.delete()
 
 
 def write_read(key, content, write_mode, read_mode, **kwargs):
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -2,7 +2,7 @@
 import io
 import os
 
-from google.cloud import storage
+import google.cloud.storage
 
 import smart_open
 
@@ -12,7 +12,7 @@ assert _GCS_BUCKET is not None, 'please set the SO_GCS_BUCKET environment variab
 
 
 def initialize_bucket():
-    storage_client = storage.Client()
+    storage_client = google.cloud.storage.Client()
     bucket = storage_client.get_bucket(_GCS_BUCKET)
     blobs = bucket.list_blobs()
     for blob in blobs:
@@ -116,5 +116,5 @@ def test_gcs_performance_small_reads(benchmark):
     with smart_open.open(key, 'wb') as fout:
         fout.write(one_megabyte_of_msgs)
 
-    actual = benchmark(read_length_prefixed_messages, key, 'rb', buffer_size=ONE_MIB)
+    actual = benchmark(read_length_prefixed_messages, key, 'rb', buffering=ONE_MIB)
     assert actual == one_megabyte_of_msgs
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -23,7 +23,7 @@ def write_read(key, content, write_mode, read_mode, encoding=None, **kwargs):
     with smart_open.open(key, write_mode, encoding=encoding, **kwargs) as fout:
         fout.write(content)
     with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
-        actual = fin.read()
+        return fin.read()
     return actual
 
 
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -44,7 +44,7 @@ def test_gcs_readwrite_text(benchmark):
 
     key = _GCS_URL + '/sanity.txt'
     text = 'с гранатою в кармане, с чекою в руке'
-    actual = benchmark(write_read, key, text, 'w', 'r', 'utf-8')
+    actual = benchmark(write_read, key, text, 'w', 'r', encoding='utf-8')
     assert actual == text
 
 
@@ -53,7 +53,7 @@ def test_gcs_readwrite_text_gzip(benchmark):
 
     key = _GCS_URL + '/sanity.txt.gz'
     text = 'не чайки здесь запели на знакомом языке'
-    actual = benchmark(write_read, key, text, 'w', 'r', 'utf-8')
+    actual = benchmark(write_read, key, text, 'w', 'r', encoding='utf-8')
     assert actual == text
 
 
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -30,7 +30,7 @@ def write_read(key, content, write_mode, read_mode, encoding=None, **kwargs):
 def read_length_prefixed_messages(key, read_mode, encoding=None, **kwargs):
     with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
         actual = b''
-        length_byte = fin.read(1);
+        length_byte = fin.read(1)
         while len(length_byte):
             actual += length_byte
             msg = fin.read(ord(length_byte))
@@ -106,9 +106,9 @@ def test_gcs_performance_small_reads(benchmark):
 
     ONE_MIB = 1024**2
     one_megabyte_of_msgs = io.BytesIO()
-    msg = b'\x0f' + b'0123456789abcde' # a length-prefixed "message"
+    msg = b'\x0f' + b'0123456789abcde'  # a length-prefixed "message"
     for _ in range(0, ONE_MIB, len(msg)):
-            one_megabyte_of_msgs.write(msg)
+        one_megabyte_of_msgs.write(msg)
     one_megabyte_of_msgs = one_megabyte_of_msgs.getvalue()
 
     key = _GCS_URL + '/many_reads_performance.bin'
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -19,7 +19,7 @@ def initialize_bucket():
         blob.delete()
 
 
-def write_read(key, content, write_mode, read_mode, encoding=None, **kwargs):
+def write_read(key, content, write_mode, read_mode, **kwargs):
     with smart_open.open(key, write_mode, encoding=encoding, **kwargs) as fout:
         fout.write(content)
     with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -30,8 +30,9 @@ def write_read(key, content, write_mode, read_mode, **kwargs):
 
 
 def read_length_prefixed_messages(key, read_mode, **kwargs):
+    result = io.BytesIO()
+
     with smart_open.open(key, read_mode, **kwargs) as fin:
-        result = io.BytesIO()
         length_byte = fin.read(1)
         while len(length_byte):
             result.write(length_byte)
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -29,14 +29,14 @@ def write_read(key, content, write_mode, read_mode, encoding=None, **kwargs):
 
 def read_length_prefixed_messages(key, read_mode, encoding=None, **kwargs):
     with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
-        actual = b''
+        result = io.BytesIO()
         length_byte = fin.read(1)
         while len(length_byte):
-            actual += length_byte
+            result.write(length_byte)
             msg = fin.read(ord(length_byte))
-            actual += msg
+            result.write(msg)
             length_byte = fin.read(1)
-    return actual
+    return result.getvalue()
 
 
 def test_gcs_readwrite_text(benchmark):
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -1,22 +1,16 @@
 # -*- coding: utf-8 -*-
 import io
 import os
-
-import google.cloud.storage
+import subprocess
 
 import smart_open
 
-_GCS_BUCKET = os.environ.get('SO_GCS_BUCKET')
-_GCS_URL = 'gs://' + _GCS_BUCKET
-assert _GCS_BUCKET is not None, 'please set the SO_GCS_BUCKET environment variable'
+_GCS_URL = os.environ.get('SO_GCS_URL')
+assert _GCS_URL is not None, 'please set the SO_GCS_URL environment variable'
 
 
 def initialize_bucket():
-    storage_client = google.cloud.storage.Client()
-    bucket = storage_client.get_bucket(_GCS_BUCKET)
-    blobs = bucket.list_blobs()
-    for blob in blobs:
-        blob.delete()
+    subprocess.check_call(['gsutil' 'rm', '-r', _GCS_URL])
 
 
 def write_read(key, content, write_mode, read_mode, **kwargs):
--- a/integration-tests/test_gcs.py
+++ b/integration-tests/test_gcs.py
@@ -20,15 +20,14 @@ def initialize_bucket():
 
 
 def write_read(key, content, write_mode, read_mode, **kwargs):
-    with smart_open.open(key, write_mode, encoding=encoding, **kwargs) as fout:
+    with smart_open.open(key, write_mode, **kwargs) as fout:
         fout.write(content)
-    with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
+    with smart_open.open(key, read_mode, **kwargs) as fin:
         return fin.read()
-    return actual
 
 
-def read_length_prefixed_messages(key, read_mode, encoding=None, **kwargs):
-    with smart_open.open(key, read_mode, encoding=encoding, **kwargs) as fin:
+def read_length_prefixed_messages(key, read_mode, **kwargs):
+    with smart_open.open(key, read_mode, **kwargs) as fin:
         result = io.BytesIO()
         length_byte = fin.read(1)
         while len(length_byte):
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -23,11 +23,13 @@ logger = logging.getLogger(__name__)
 
 _READ_BINARY = 'rb'
 _WRITE_BINARY = 'wb'
+
 _MODES = (_READ_BINARY, _WRITE_BINARY)
 """Allowed I/O modes for working with GCS."""
 
 _BINARY_TYPES = (six.binary_type, bytearray)
 """Allowed binary buffer types for writing to the underlying GCS stream"""
+
 if sys.version_info >= (2, 7):
     _BINARY_TYPES = (six.binary_type, bytearray, memoryview)
 
@@ -49,10 +51,13 @@ DEFAULT_BUFFER_SIZE = 256 * 1024
 
 START = 0
 """Seek to the absolute start of a GCS file"""
+
 CURRENT = 1
 """Seek relative to the current positive of a GCS file"""
+
 END = 2
 """Seek relative to the end of a GCS file"""
+
 _WHENCE_CHOICES = (START, CURRENT, END)
 
 _SUCCESSFUL_STATUS_CODES = (200, 201)
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -475,7 +475,7 @@ class BufferedOutputBase(io.BufferedIOBase):
             'Content-Range': _make_range_string(start, stop, end)
         }
         data = self._buf
-        response = self._session.put(self._resumeable_upload_url, data=data, headers=headers)
+        response = self._session.put(self._resumeable_upload_url, data=self._buf, headers=headers)
 
         if response.status_code not in _SUCCESSFUL_STATUS_CODES:
             logger.error("upload failed with status %s", response.status_code)
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -380,7 +380,7 @@ class BufferedOutputBase(io.BufferedIOBase):
         self._credentials = self._client._credentials
         self._bucket = self._client.bucket(bucket)  # type: google.cloud.storage.Bucket
         self._blob = self._bucket.blob(blob)  # type: google.cloud.storage.Blob
-        assert min_part_size % _MIN_MIN_PART_SIZE == 0, 'min part size must be a multiple of 256KB'
+        assert min_part_size % _REQUIRED_CHUNK_MULTIPLE == 0, 'min part size must be a multiple of 256KB'
         self._min_part_size = min_part_size
 
         self._total_size = 0
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -345,6 +345,22 @@ class SeekableBufferedInputBase(io.BufferedIOBase):
                 logger.debug('reached EOF while filling buffer')
                 self._eof = True
 
+    def __str__(self):
+        return "(%s, %r, %r)" % \
+              (self.__class__.__name__, self._bucket.name, self._blob.name)
+
+    def __repr__(self):
+        return (
+            "%s("
+            "bucket=%r, "
+            "blob=%r, "
+            "buffer_size=%r)"
+        ) % (
+            self._bucket.name,
+            self._blob.name,
+            self._buffer_size,
+        )
+
 
 class BufferedOutputBase(io.BufferedIOBase):
     """Writes bytes to GCS.
@@ -445,11 +461,11 @@ class BufferedOutputBase(io.BufferedIOBase):
         content_length = self._buf.tell()
         start = self._total_size - content_length
         stop = self._total_size - 1
-        if content_length != self._min_part_size:
+        if stop - start + 1 == content_length:
             end = content_length
         else:
             end = _UNKNOWN_FILE_SIZE
-            if content_length != _REQUIRED_CHUNK_MULTIPLE:
+            if content_length % _REQUIRED_CHUNK_MULTIPLE != 0:
                 stop = content_length // _REQUIRED_CHUNK_MULTIPLE * _REQUIRED_CHUNK_MULTIPLE - 1
 
         self._buf.seek(0)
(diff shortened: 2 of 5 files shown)


README.rst
```diff
@@ -174,6 +175,14 @@ More examples
     with open('s3://bucket/key.txt', 'wb', transport_params=transport_params) as fout:
         fout.write(b'here we stand')
 
+    # stream from GCS
+    for line in open('gs://my_bucket/my_file.txt'):
+        print(line)
+
+    # stream content *into* GCS (write mode):
+    with open('gs://my_bucket/my_file.txt') as fout:
+        fout.write(b'hello world')
```
mpenkov: Why are you writing bytes here? Isn't `fout` open in text mode? Also, you're not specifying write mode here. This example looks broken.

integration-tests/test_gcs.py
```diff
@@ -0,0 +1,120 @@
...
+def initialize_bucket():
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(_GCS_BUCKET)
+    blobs = bucket.list_blobs()
+    for blob in blobs:
+        blob.delete()
+
+
+def write_read(key, content, write_mode, read_mode, encoding=None, **kwargs):
+    with smart_open.smart_open(key, write_mode, encoding=encoding, **kwargs) as fout:
```
mpenkov: Please use the `open` function, here and everywhere. The `smart_open` function is deprecated.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+
+from google.cloud import storage, exceptions
+import google.auth.transport.requests as google_requests
+import six
+
+import smart_open.bytebuffer
+
+logger = logging.getLogger(__name__)
+
+READ_BINARY = 'rb'
```
mpenkov: These can probably be internal.

Please review all module-level variables and functions. If they aren't needed from outside this module, then make them internal. If they are needed outside, add descriptive docstrings.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+_BINARY_TYPES = (six.binary_type, bytearray)
+"""Allowed binary buffer types for writing to the underlying GCS stream"""
+if sys.version_info >= (2, 7):
+    _BINARY_TYPES = (six.binary_type, bytearray, memoryview)
+
+_UNKNOWN_FILE_SIZE = '*'
+
+_RESUME_INCOMPLETE = 308
+
+BINARY_NEWLINE = b'\n'
```
mpenkov: If these are not internal, add docstrings for them.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+"""Google requires you to upload in multiples of 256 KB, except for the last part."""
+
+MIN_MIN_PART_SIZE = 256 * 1024
+"""The absolute minimum permitted by Google."""
+DEFAULT_BUFFER_SIZE = 256 * 1024
+
+START = 0
+CURRENT = 1
+END = 2
+WHENCE_CHOICES = [START, CURRENT, END]
```
mpenkov: This can be a tuple.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+        The name of the bucket this object resides in.
+    blob_id: str
+        The name of the blob within the bucket.
+    mode: str
+        The mode for opening the object.  Must be either "rb" or "wb".
+    buffer_size: int, optional
+        The buffer size to use when performing I/O.
+    min_part_size: int, optional
+        The minimum part size for multipart uploads.  For writing only.
+    client: object, optional
```
mpenkov: Be more precise about expected data types:

```suggestion
    client: google.cloud.storage.Client, optional
```

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+        )
+    else:
+        raise NotImplementedError('GCS support for mode %r not implemented' % mode)
+
+
+class RawReader(object):
+    """Read an GCS blob."""
+    def __init__(self,
+                 gcs_blob,  # type: storage.Blob
+                 ):
```
mpenkov: This can fit on a single line, no need to break across multiple lines.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+            bucket_id,
+            blob_id,
+            min_part_size=min_part_size,
+            client=client,
+        )
+    else:
+        raise NotImplementedError('GCS support for mode %r not implemented' % mode)
+
+
+class RawReader(object):
```
mpenkov: This looks like a useless class to me. You can replace it easily with:

```python
alternative = io.BytesIO(gcs_blob.download_as_string())
```

and it will behave the same way.

BTW, how large are these blobs? Is it acceptable to store them entirely in memory? I suspect the answer is no.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+    ):
+        if not client:
+            client = storage.Client()
+
+        bucket = client.get_bucket(bucket)  # type: storage.Bucket
+
+        self._blob = bucket.get_blob(key)   # type: storage.Blob
+        self._size = self._blob.size if self._blob.size else 0
+
+        self._raw_reader = RawReader(self._blob)
```
mpenkov: This loads the whole thing into memory, is this really what we want?

smart_open/smart_open_lib.py
```diff
@@ -831,6 +846,16 @@ def _unquote(text):
     return text and urlparse.unquote(text)
 
 
+def _parse_uri_gcs(parsed_uri):
+    assert parsed_uri.scheme in smart_open_gcs.SUPPORTED_SCHEMES
+
+    bucket_id, blob_id = parsed_uri.netloc, parsed_uri.path[1:]
+
+    return Uri(
```
mpenkov: This can fit on one line.

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
...
+    def _fill_buffer(self, size=-1):
+        size = size if size >= 0 else self._buffer._chunk_size
+        while len(self._buffer) < size and not self._eof:
+            bytes_read = self._buffer.fill(self._raw_reader)
+            if bytes_read == 0:
+                logger.debug('reached EOF while filling buffer')
+                self._eof = True
+
+
+class SeekableBufferedInputBase(BufferedInputBase):
```
mpenkov: Why do you need two separate classes?

1. BufferedInputBase
2. SeekableBufferedInputBase

Can't you get away with just one of them?

smart_open/gcs.py
```diff
@@ -0,0 +1,524 @@
+# -*- coding: utf-8 -*-
+"""Implements file-like objects for reading and writing to/from GCS."""
+
+import io
+import logging
+import sys
+
+from google.cloud import storage, exceptions
```
mpenkov: Please do this instead:

```suggestion
import google.cloud.exceptions
import google.cloud.storage
```

It makes your code more difficult to write, because it is more verbose.
However, it is easier to read, because the reader doesn't have to remember where `storage` and `exceptions` are coming from.
Readability is more important than writeability in this case.

petedannemann:
> Hi, thank you for your contribution.
> 
> I skimmed the reader part and left you some comments here and there. Please have a look.
> 
> My main concern is that you're loading the whole blob into memory when reading. That's what the `download_as_string` method appears to do. We definitely want to avoid this, because the remote object could be much larger than what can fit into memory.

Thanks for the review. A lot of these issues came from me reusing a lot of stuff based on how it was written in the s3 module, but it looks like that's not the approach we want to take. I'll work on addressing these issues, especially the one regarding memory

mpenkov:
That module evolved a bit over time, so you don't necessarily have to mimic everything it does.

Try to work around the memory issue (from memory, the S3 module doesn't suffer from this problem, so that particular part may provide some hints for you) and let me know when you're ready for another round of review.

Thanks!

integration-tests/test_gcs.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+import io
+import os
+
+import google.cloud.storage
+
+import smart_open
+
+_GCS_BUCKET = os.environ.get('SO_GCS_BUCKET')
+_GCS_URL = 'gs://' + _GCS_BUCKET
```
piskvorky: Why `gs://` when the service is GCS?

petedannemann: `gs://` is the most common scheme for Google Cloud Storage, but `gcs://` is also used occasionally

piskvorky: Is there some specification for this? Where does this convention come from, where is it used?

petedannemann: [Google's tools](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage#gcs-uri) (gsutil, Google Cloud Storage console uris) all use `gs`, I actually had not seen `gcs` before the issue that requested this feature, and think maybe we should drop `gcs` for simplicity.

piskvorky: OK, Google itself = the originator of the protocol is a source that's authoritative enough :)

petedannemann: I agree, it's very confusing because they use the GCS acronym to refer to the Google Cloud Storage service. I wouldn't be surprised is their reasoning was one less letter to send in network calls, sounds about right for Google haha.

To clarify, where did we land with this? Just support `gs` or both `gs` and `gcs`?

piskvorky: I'm +1 on `gs` only, if that's the established standard. @mpenkov thoughts?

smart_open/gcs.py
petedannemann: Great suggestion, RawReader has been removed

smart_open/gcs.py
petedannemann: Another great point, BufferedInputBase has been removed

petedannemann:
@mpenkov I believe I have addressed the issues you brought up. Please review when you get the chance

integration-tests/test_gcs.py
mpenkov: Yes, `gs` is fine initially. If people complain, we can revisit later.

mpenkov:
OK, I will review your changes during the week. In the meanwhile, the tests are failing for some reason in CI, please make them pass.

smart_open/gcs.py
```diff
@@ -141,28 +125,34 @@ def seek(self, position):
...
-        if size == -1:
-            binary = self._body
+            binary = b''
+        elif size == -1:
+            binary = self._blob.download_as_string(start=position)
         else:
-            binary = self._body[:size]
-        self._position += len(binary)
+            start, end = position, position + size
+            binary = self._blob.download_as_string(start=start, end=end)
```
piskvorky: What happens when trying to read past the end of the file? `end > self._size`

smart_open/smart_open_lib.py
```diff
@@ -684,6 +693,8 @@ def _parse_uri(uri_as_string):
     Supported URI schemes are:
 
       * file
+      * gcs
+      * gs
```
piskvorky: One line (if the same protocol).

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,660 @@
...
+            return
+        except google.cloud.exceptions.NotFound as err:
+            logger.error('caught %r, retrying', err)
+            time.sleep(sleep_time)
+
+    assert False, 'failed to create bucket %s after %d attempts' % (BUCKET_NAME, num_attempts)
+
+
+def mock_gcs(func):
+    '''Mocks the function and provides additional required arguments.'''
```
piskvorky:
```suggestion
    '''Mock the function and provide additional required arguments.'''
```

PEP257: Docstrings in imperative mode.

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,660 @@
...
+def maybe_mock_gcs(func):
+    if DISABLE_MOCKS:
+        return func
+    else:
+        return mock_gcs(func)
+
+
+@maybe_mock_gcs
+def setUpModule():
+    '''Called once by unittest when initializing this module.  Sets up the
```
piskvorky:
```suggestion
    '''Called once by unittest when initializing this module.  Set up the
```

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,660 @@
...
+    test GCS bucket.
+
+    '''
+    storage_client.create_bucket(BUCKET_NAME)
+
+
+@maybe_mock_gcs
+def tearDownModule():
+    '''Called once by unittest when tearing down this module.  Empties and
```
piskvorky:
```suggestion
    '''Called once by unittest when tearing down this module. Empty and
```

setup.py
```diff
@@ -59,14 +59,15 @@ def read(fname):
     'boto >= 2.32',
     'requests',
     'boto3',
+    'google-cloud-storage',
```
piskvorky: How heavy is this dependency? (download size, subdependencies, any compilation?, etc)

smart_open/gcs.py
```diff
@@ -0,0 +1,521 @@
...
+import six
+
+import smart_open.bytebuffer
+
+logger = logging.getLogger(__name__)
+
+_READ_BINARY = 'rb'
+_WRITE_BINARY = 'wb'
+_MODES = (_READ_BINARY, _WRITE_BINARY)
+"""Allowed I/O modes for working with GCS."""
```
piskvorky: Why are these docstrings? I'd expect a comment *above* each variable, instead of a docstring *below* it.

mpenkov: Sphinx picks them up that way.

Not that we use Sphinx for this project, but it's a convention.

petedannemann: This was done in response to @mpenkov's [comment](https://github.com/RaRe-Technologies/smart_open/pull/404#discussion_r363068746). The convention used in the S3 module is docstrings below the constant.

setup.py
petedannemann: The wheel is [72.7 kB on pypi](https://pypi.org/project/google-cloud-storage/#files). It has [three other GCP dependencies](https://github.com/googleapis/google-cloud-python/blob/master/storage/setup.py), I do not see any compilation required. I think optional installation for smart_open is a good idea, and can move it into `extras_require` in the setup.py

piskvorky: Nah. If it's light (and reasonably stable), I'd prefer a direct dependency, to keep things simple.

smart_open/gcs.py
piskvorky: OK, but I find that thoroughly confusing. In the following block of code, which comment relates to which variable?

```python

START = 0
"""Seek to the absolute start of a GCS file"""
CURRENT = 1
"""Seek relative to the current positive of a GCS file"""
END = 2
"""Seek relative to the end of a GCS file"""
_WHENCE_CHOICES = (START, CURRENT, END)

```

petedannemann:
> OK, I will review your changes during the week. In the meanwhile, the tests are failing for some reason in CI, please make them pass.

I made some fixes and now the only error is from Python 2.7 not recognizing the `test_gcs.py` file when it runs `python setup.py test`. I'm pretty stumped on why this is happening, any ideas?

petedannemann:
> OK, I will review your changes during the week. In the meanwhile, the tests are failing for some reason in CI, please make them pass.

The tests are passing, but the build for Python 2.7 is failing when it runs `python setup.py test` due to the error below:
```bash
>> AttributeError: 'module' object has no attribute 'test_gcs'
```

Either of you know how to get Python 2.7 to discover my new test file?

smart_open/gcs.py
petedannemann: `google.cloud.storage.Blob.download_as_string` appears to handle this for us. I added a test to verify.

```python
@maybe_mock_gcs
    def test_read_past_end(self):
        content = b'englishman\nin\nnew\nyork\n'
        put_to_bucket(contents=content)

        with smart_open.gcs.SeekableBufferedInputBase(BUCKET_NAME, BLOB_NAME) as fin:
            data = fin.read(100)

        self.assertEqual(data, content)
```

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+        raise io.UnsupportedOperation
+
+    def read(self, size=-1):
+        """Read up to size bytes from the object and return them."""
+        if size == 0:
+            return b''
+        elif size < 0:
+            from_buf = self._read_from_buffer()
+            self._current_pos = self._size
+            return from_buf + self._raw_reader.read()
```
mpenkov: You don't need to declare variables here.

```suggestion
            return self._read_from_buffer() + self._raw_reader.read()
```

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+                the_line.write(self._read_from_buffer(next_newline + 1))
+                break
+            else:
+                the_line.write(self._read_from_buffer())
+                self._fill_buffer()
+        return the_line.getvalue()
+
+    def terminate(self):
+        """Do nothing."""
+        pass
```
mpenkov: Is this OK?

Theoretically, you can continue reading after calling terminate, which doesn't sound very useful. Why is this function even needed?

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+    def _fill_buffer(self, size=-1):
+        size = size if size >= 0 else self._buffer._chunk_size
+        while len(self._buffer) < size and not self._eof:
+            bytes_read = self._buffer.fill(self._raw_reader)
+            if bytes_read == 0:
+                logger.debug('reached EOF while filling buffer')
+                self._eof = True
+
+    def __str__(self):
+        return "(%s, %r, %r)" % \
```
mpenkov: One line.

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+            min_part_size=_DEFAULT_MIN_PART_SIZE,
+            client=None,  # type: google.cloud.storage.Client
+    ):
+        if client is None:
+            client = google.cloud.storage.Client()
+        self._client = client
+        self._credentials = self._client._credentials
+        self._bucket = self._client.bucket(bucket)  # type: google.cloud.storage.Bucket
+        self._blob = self._bucket.blob(blob)  # type: google.cloud.storage.Blob
+        assert min_part_size % _MIN_MIN_PART_SIZE == 0, 'min part size must be a multiple of 256KB'
```
mpenkov: You really mean this, right?

```suggestion
        assert min_part_size % _REQUIRED_CHUNK_MULTIPLE == 0, 'min part size must be a multiple of 256KB'
```

mpenkov: Also, you need to check `min_part_size >= _MIN_MIN_PART_SIZE`.

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+        pass
+
+    #
+    # Override some methods from io.IOBase.
+    #
+    def close(self):
+        logger.debug("closing")
+        if self._total_size == 0:  # empty files
+            self._upload_empty_part()
+        if self._buf.tell():
```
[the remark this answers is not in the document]
mpenkov: Do you need to clean up other class members here?

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+        self._client = client
+        self._credentials = self._client._credentials
+        self._bucket = self._client.bucket(bucket)  # type: google.cloud.storage.Bucket
+        self._blob = self._bucket.blob(blob)  # type: google.cloud.storage.Blob
+        assert min_part_size % _MIN_MIN_PART_SIZE == 0, 'min part size must be a multiple of 256KB'
+        self._min_part_size = min_part_size
+
+        self._total_size = 0
+        self._total_parts = 0
+        self._buf = io.BytesIO()
```
mpenkov: I think a better name would be:

```suggestion
        self._current_part = io.BytesIO()
```

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+    # Internal methods.
+    #
+    def _upload_next_part(self):
+        part_num = self._total_parts + 1
+        logger.info("uploading part #%i, %i bytes (total %.3fGB)",
+                    part_num, self._buf.tell(), self._total_size / 1024.0 ** 3)
+        content_length = self._buf.tell()
+        start = self._total_size - content_length
+        stop = self._total_size - 1
+        if stop - start + 1 == content_length:
```
mpenkov: What is this fidgeting here?

LHS:

- `stop - start + 1`
- `((self._total_size - 1) - (self._total_size - content_length) + 1` (substituting)
- `self._total_size - 1 - self._total_size + content_length + 1` (expand parentheses)
- `content_length` (simplify, cancel out)

So, the LHS is always equal to `content_length`, which is equal to RHS.

What am I missing here?

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+                stop = content_length % _REQUIRED_CHUNK_MULTIPLE - 1
+
+        self._buf.seek(0)
+
+        headers = {
+            'Content-Length': str(content_length),
+            'Content-Range': _make_range_string(start, stop, end)
+        }
+        data = self._buf
+        response = self._session.put(self._resumeable_upload_url, data=data, headers=headers)
```
mpenkov: You don't need to declare `data` just to use it in the line below.

```suggestion
        response = self._session.put(self._resumeable_upload_url, data=self._buf, headers=headers)
```

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,659 @@
...
+        self.name = name
+        self._bucket = bucket  # type: FakeBucket
+        self._exists = False
+        self.__contents = io.BytesIO()
+
+        if create:
+            self._create_if_not_exists()
+
+    def create_resumable_upload_session(self):
+        resumeable_upload_url = self.RESUMABLE_SESSION_URI_TEMPLATE.format(
```
mpenkov: Please don't use .format, we avoid it in our source code. Use `formatstr % dict(foo='bar')`.

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,659 @@
...
+
+    def list_blobs(self):
+        return self.blobs
+
+    def _delete_blob(self, blob):
+        self.blobs.remove(blob)
+
+
+class FakeBlob(object):
+    RESUMABLE_SESSION_URI_TEMPLATE = (
```
mpenkov: Move this to module scope.

smart_open/tests/test_gcs.py
```diff
@@ -0,0 +1,659 @@
...
+logger = logging.getLogger(__name__)
+
+
+def ignore_resource_warnings():
+    if six.PY2:
+        return
+    warnings.filterwarnings("ignore", category=ResourceWarning, message="unclosed.*<ssl.SSLSocket.*>")  # noqa
+
+
+class FakeBucket(object):
```
mpenkov: It looks like you're implementing your own mocks. Is there really no library to mock GCS? (looks like [there isn't](https://github.com/googleapis/google-cloud-python/issues/5816)).

If so, then we need to write tests for these mocks as well, to ensure they're behaving as expected. The risk is that the mock implementation is buggy in a particular way that hides bugs in our GCS tests.

petedannemann: I was unable to find a mocking library for GCS. I will write tests for the mocks. Do you think these should all reside in the same file or should there be something like one file for the mocks, `gcs_utils.py` and another for the mock tests `test_gcs_utils`?

integration-tests/test_gcs.py
```diff
@@ -0,0 +1,120 @@
...
+    for _ in range(0, ONE_MIB, len(msg)):
+        one_megabyte_of_msgs.write(msg)
+    one_megabyte_of_msgs = one_megabyte_of_msgs.getvalue()
+
+    key = _GCS_URL + '/many_reads_performance.bin'
+
+    with smart_open.open(key, 'wb') as fout:
+        fout.write(one_megabyte_of_msgs)
+
+    actual = benchmark(read_length_prefixed_messages, key, 'rb', buffering=ONE_MIB)
```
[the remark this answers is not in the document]
petedannemann: > Shouldn't buffering be wrapped in a transport_kwargs dict? It's a GCS-only parameter, right?

Buffering is a kwarg for the built-in `open` function ([source](https://python-reference.readthedocs.io/en/latest/docs/functions/open.html)) and I thought it would be simpler for it to be named the same instead of making it specific to GCS. If you feel otherwise I can wrap it in transport_kwargs.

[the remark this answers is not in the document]
petedannemann: > Did you try running these tests against a real GCS bucket?

Yes, I have ran both the integration and unit tests against a real bucket and they are currently passing.

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+CURRENT = 1
+"""Seek relative to the current positive of a GCS file"""
+END = 2
+"""Seek relative to the end of a GCS file"""
+_WHENCE_CHOICES = (START, CURRENT, END)
+
+_SUCCESSFUL_STATUS_CODES = (200, 201)
+
+
+def clamp(value, minval, maxval):
```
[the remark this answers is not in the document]
petedannemann: Changed so clamp is imported from S3 instead of copied here

smart_open/gcs.py
```diff
@@ -0,0 +1,523 @@
...
+
+        self._buf.write(b)
+        self._total_size += len(b)
+
+        if self._buf.tell() >= self._min_part_size:
+            self._upload_next_part()
+
+        return len(b)
+
+    def terminate(self):
```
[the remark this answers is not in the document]
petedannemann: This was something I copied from the s3 module. I think having some way to abort an upload seems valuable, although it would probably be used very infrequently. I will document it

smart_open/tests/test_gcs.py
mpenkov: Same file is fine.

petedannemann:
@mpenkov @piskvorky I've been trying to think through the GCS mocks and imitate how moto works and unfortunately I think this will be a significantly larger endeavor than I originally thought. I also am worried about the maintainability of them in the long run and coupling them with the smart_open library doesn't seem like the right way of doing it. I see two options:

1. Develop a Python package for mocking GCS that lives outside of smart_open (this will probably take me months and will be a risk for smart_open)
2. Refactor all gcs tests to simply assert calls to the `google.cloud.storage library`. These tests will not be as good as the current ones, but will still yield some testing benefits

What do you think?

mpenkov:
I agree that 1) is probably too much for any of us to take on right now.

I think 2) is more of a step in the right direction. If the goal is to prevent regression, then your existing tests and mocks are sufficient. To prove correctness, we will have to use actual GCS buckets (integration testing).

petedannemann:
@mpenkov I retract my last comment and think the mocks are actually the best testing methodology for this situation. I think I got caught up in scope creep from looking at all of the functionality of moto, but the mocks here only need functionality as it relates to the tests they're being used with. I added tests for the mocks and am ready for another review.

mpenkov:
OK, I'll have a look within a week and let you know.

smart_open/gcs.py
```diff
@@ -0,0 +1,530 @@
...
+        super(UploadFailedError, self).__init__(message)
+        self.status_code = status_code
+        self.text = text
+
+
+def open(
+        bucket_id,
+        blob_id,
+        mode,
+        buffering=DEFAULT_BUFFER_SIZE,
```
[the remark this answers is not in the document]
petedannemann: Changed to `buffer_size` to avoid this issue and mimic the s3 module

petedannemann:
> Looking very good. Left you some minor comments.
> 
> Also, you're missing tests for the top-level module (smart_open_lib). These don't have to be super-detailed, but you want to make sure that `smart_open.open("gs://bucket/key")` actually works.

Thanks @mpenkov! I added your comments and added those tests.

petedannemann:
Thanks for your help and patience!

wookayin:
How does it handle authentication? gcloud and gsutil are well-configured and working well but `open('gs://...')` does not work.

mpenkov:
@wookayin Can you please open a new ticket to describe your problem? Be sure to fill out the full issue template.

petedannemann:
@wookayin it uses Google's [google-cloud-storage](https://pypi.org/project/google-cloud-storage/) package so please refer to that package's documentation on how to configure authentication. The preferred method is to create a service account key file and set your GOOGLE_APPLICATION_CREDENTIALS environment variable to the path of that file.

mpenkov:
@petedannemann Is that documented anywhere, like the readme or howto.md? If not, it'd be good to add some documentation to point people in the right direction (the above comment is already a solid start).

petedannemann:
It is not. I will work on a PR to add this missing documentation soon and consider adding some examples of how you can use other authentication methods by passing in your own `google.cloud.storage.Client` object to transport_kwargs.

PR #408, opened 2020-01-11
Medium-scale refactoring

mpenkov:
Motivation:

- smart_open_lib.py was cluttered
- parsing logic was messy

Main changes:

- smart_open.parse_uri is now a publicly available function
- Parsers for the individual schemes, e.g. HDFS, S3 now live in the
  corresponding submodule
- Moved compression to a separate submodule
- Moved transport registration to a separate submodule
- **Finally** dropped Py2.7 support

I haven't changed any of the unit tests, in order to demonstrate that
everything still works as expected from the outside.

#### Checklist


smart_open/s3.py
```diff
@@ -13,13 +13,17 @@
 import logging
 import warnings
 
+import boto
```
piskvorky: Do we still need both `boto` and `boto3`? What's the relationship between the two (I forgot).

extending.md
```diff
@@ -0,0 +1,69 @@
...
+
+def open_uri(uri_as_str, mode, transport_params):
+    """Return a file-like object pointing to the URI."""
+    ...
+```
+
+Have a look at the existing mechanisms to see how they work.
+You may define other functions and classes as necessary for your implementation.
+
+Once your module is working, register it in the `smart_open/smart_open_lib.py` file.
```
petedannemann: I think it would be helpful to document what other documentation needs updating when adding a new transport mechanism. The instances I am aware of are README examples, the setup.py description, and smart_open.open.__doc__.

smart_open/compression.py
```diff
@@ -0,0 +1,116 @@
...
+    return sorted(_COMPRESSOR_REGISTRY.keys())
+
+
+def register_compressor(ext, callback):
+    """Register a callback for transparently decompressing files with a specific extension.
+
+    Parameters
+    ----------
+    ext: str
+        The extension.
```
menshikh-iv: `"should starts with '.', like .gz`

smart_open/compression.py
```diff
@@ -0,0 +1,116 @@
...
+    callback: callable
+        The callback.  It must accept two position arguments, file_obj and mode.
```
menshikh-iv: `"Function that will be called when you process file with concrete extension"`

extending.md
```diff
@@ -0,0 +1,69 @@
+# Extending `smart_open`
```
menshikh-iv: link from README to this file

smart_open/hdfs.py
```diff
@@ -18,8 +18,35 @@
 import logging
 import subprocess
 
+from six.moves.urllib import parse as urlparse
```
menshikh-iv: offtopic: any plans for dropping py2 support in `smart_open`?

smart_open/s3.py
```diff
@@ -55,6 +61,140 @@
 WHENCE_CHOICES = [START, CURRENT, END]
 
 
+def _my_urlsplit(url):
```
menshikh-iv: any better name?

smart_open/hdfs.py
mpenkov: Yes, let's do it in a separate PR, as soon as possible. The rationale is the same. People who cannot upgrade to Py3 can continue to use existing versions of smart_open.

CC @piskvorky

smart_open/s3.py
mpenkov: We don't really need `boto`, IMHO. It's entirely separate to `boto3`.

Let's remove it in a separate PR.

smart_open/hdfs.py
piskvorky: Yes, let's. With good communication it shouldn't be an issue.

Are there any stats on smart_open vs Python version?

mpenkov: Should be a way to do it via https://packaging.python.org/guides/analyzing-pypi-package-downloads/, I'll have a look at it during the week.

mpenkov:
@piskvorky @menshikh-iv I've made more changes and made sure the build passes. Can you please re-review?

smart_open/utils.py
```diff
@@ -1,3 +1,11 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 Radim Rehurek <me@radimrehurek.com>
+#
+# This code is distributed under the terms and conditions
+# from the MIT License (MIT).
+#
+"""Helper functions for documentation, etc."""
```
piskvorky: Newline before + after docstring please.

extending.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. `smart_open` will be able to open any URI supported by your module
+2. The docstring for the `smart_open.open` function will contain a section
+   detailing the parameters for your transport module.
+3. The docstring for the `parse_uri` function will include the schemas and
+   examples supported by your module.
+
+You can confirm the documentation changes by running:
+
+    python -c 'help("smart_open")'
```
menshikh-iv: link to submodule should be here, I guess?

extending.md
```diff
@@ -0,0 +1,132 @@
...
+
+For example:
+
+```python
+def _handle_xz(file_obj, mode):
+    import lzma
+    return lzma.LZMAFile(filename=file_obj, mode=mode, format=lzma.FORMAT_XZ)
+
+
+register_compressor('.xz', _handle_xz)
```
menshikh-iv: is that "global" setting? I.e. can I do that in "runtime" on my side?

setup.py
```diff
@@ -18,7 +17,7 @@ def _get_version():
     curr_dir = os.path.dirname(os.path.abspath(__file__))
     with open(os.path.join(curr_dir, 'smart_open', 'version.py')) as fin:
         #
-        # __version__ = '1.8.4'
+        # example: __version__ = '1.8.4'
```
menshikh-iv: ?

extending.md
```diff
@@ -0,0 +1,132 @@
+# Extending `smart_open`
+
+This document targets potential contributors to `smart_open`.
```
menshikh-iv: What's about actions in case if I want to add new format, but without pushing them to smart_open (for example, reader for proprietary stuff in the company, useless for open-source).

smart_open/ssh.py
```diff
@@ -38,6 +41,36 @@
 
 DEFAULT_PORT = 22
 
+URI_EXAMPLES = (
+    '[ssh|scp|sftp]://username@host//path/file',
```
menshikh-iv: comparing to other URI examples looks strange (not sure than all understand `[X|Y|Z]` form)

smart_open/doctools.py
```diff
@@ -156,3 +163,51 @@ def extract_examples_from_readme_rst(indent='    '):
...
+                    uri_examples.write('    * %s\n' % e)
+                seen_examples.add(e)
+
+    substrings['codecs'] = '\n'.join(
+        ['    * %s' % e for e in compression.get_supported_extensions()]
+    )
+    substrings['examples'] = extract_examples_from_readme_rst()
+
+    #
+    # The docstring can be None if -OO was passed to the interpreter.
```
menshikh-iv: Maybe warning in that case (If you can detect that.)?

smart_open/s3.py
```diff
@@ -64,10 +68,150 @@
...
+    # - Bucket names can contain lowercase letters, numbers, and hyphens.
+    # - Each label must start and end with a lowercase letter or a number.
+    #
+    # We use the above as a guide only, and do not perform any validation.  We
+    # let boto3 take care of that for us.
+    #
+    split_uri = _safe_urlsplit(uri_as_string)
+    assert split_uri.scheme in SCHEMES
+
+    port = DEFAULT_PORT
```
menshikh-iv: what's about non-default port (for example for case when smbd use https://min.io/ with s3-like AP on non-443I)?

smart_open/tests/test_bytebuffer.py
```diff
@@ -9,23 +9,25 @@
...
 CHUNK_SIZE = 1024
 
 
+def int2byte(i):
+    return bytes((i, ))
+
+
 def random_byte_string(length=CHUNK_SIZE):
-    rand_bytes = [six.int2byte(random.randint(0, 255)) for _ in range(length)]
+    rand_bytes = [int2byte(random.randint(0, 255)) for _ in range(length)]
```
menshikh-iv: non-determenistic tests, not sure then this is a good idea

menshikh-iv:
Amazing work @mpenkov :star:

extending.md
mpenkov: Yes

setup.py
mpenkov: Contents of the version.py for reference

extending.md
```diff
@@ -0,0 +1,132 @@
...
+- `smart_open.file`
+- `smart_open.s3`
+- `smart_open.ssh`
+- ... and others
+
+So, to implement a new transport mechanism, you need to create a new module.
+Your module must expose the following:
+
+```python
+SCHEMA = ...
```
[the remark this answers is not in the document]
mpenkov: Even a minimum example is a bit too involved. I'll add a link to one of the simpler submodules, like http.

smart_open/s3.py
mpenkov: https://github.com/RaRe-Technologies/smart_open/pull/408/files#diff-6d8b12a1a0fab36149bf68fc0ef770faR144

smart_open/tests/test_bytebuffer.py
mpenkov: Yeah, I agree that having random in the tests is a bit of a code smell, but it appears to be a core feature of those tests. I'm not sure how we'd go about removing them. Let's handle that in a separate PR.

smart_open/s3.py
menshikh-iv: wrong link?

mpenkov: Yes, sorry. Trying again: https://github.com/RaRe-Technologies/smart_open/blob/c1be8deaf4abad08a447dd6be3436a11f4fecd08/smart_open/s3.py#L143

PR #421, opened 2020-02-16
Gcs multiple writes

### LINKED ISSUE #417: GCS Writing multiple times does not work
#### Problem description
When I open a GCS bucket object and write multiple times only the data of the first write is stored in the object
The logs must show that it uploaded multiple parts (2+) like this `INFO:smart_open.gcs:uploading part #2`

Be sure your description clearly answers the following questions:

- What are you trying to achieve?
Upload a 1 MB file to a google bucket object

- What is the expected result?
I expect the file to be fully uploaded to the GCS

- What are you seeing instead?
Only the first part was uploaded, usually part size is about 200 to 300 KB

```Python3
import os
import io
import logging
from smart_open import open

logging.basicConfig(level=logging.INFO)

url = "gs://smart-open-784235/test-object"
size = 99 * 1024

data = io.BytesIO()
with open(url, "wb") as f:
    for _ in range(10):
        buffer = os.urandom(size)
        data.write(buffer)
        f.write(buffer)

data = bytes(data.getbuffer())
with open(url, "rb") as f:
    remote_data = f.read()

print("Local size:", len(data))
print("Remote size:", len(remote_data))
```
Output:
```
INFO:smart_open.gcs:uploading part #1, 304128 bytes (total 0.000GB)
INFO:smart_open.gcs:uploading part #2, 304128 bytes (total 0.001GB)
INFO:smart_open.gcs:uploading part #3, 304128 bytes (total 0.001GB)
INFO:smart_open.gcs:uploading part #4, 101376 bytes (total 0.001GB)
INFO:google.resumable_media.requests.download:No MD5 checksum was returned from the service while downloading https://storage.googleapis.com/download/storage/v1/b/smart-open-784235/o/test-object?generation=1581104519907572&alt=media
(which happens for composite objects), so client-side content integrity
checking is not being performed.
Local size: 1013760
Remote size: 304128
```

#### Versions

Linux-5.4.13-201.fc31.x86_64-x86_64-with-fedora-31-Thirty_One
Python 3.7.6 (default, Dec 19 2019, 22:52:49) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
smart_open 1.9.0

I have downloaded smart-open from github yesterday and used it instead of the repo version as there is not GCS support in the repos
The latest commit at the time of download was https://github.com/RaRe-Technologies/smart_open/commit/72818ca6d3a0a99e1717ab31db72bf109ac5ce65

#### Checklist

petedannemann:
#### Motivation

- Fixes #417 

#### Checklist

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -420,7 +420,6 @@ class BufferedOutputBase(io.BufferedIOBase):
         self._total_size = 0
         self._total_parts = 0
         self._bytes_uploaded = 0
-        self._closed = False
         self._current_part = io.BytesIO()
 
         self._session = google_requests.AuthorizedSession(self._credentials)
@@ -448,12 +447,12 @@ class BufferedOutputBase(io.BufferedIOBase):
                 self._upload_empty_part()
             else:
                 self._upload_final_part()
-            self._closed = True
+            self._client = None
         logger.debug("successfully closed")
 
     @property
     def closed(self):
-        return self._closed
+        return self._client is None
 
     def writable(self):
         """Return True if the stream supports writing."""
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -400,6 +400,7 @@ class BufferedOutputBase(io.BufferedIOBase):
 
         self._total_size = 0
         self._total_parts = 0
+        self._bytes_uploaded = 0
         self._current_part = io.BytesIO()
 
         self._session = google_requests.AuthorizedSession(self._credentials)
@@ -424,7 +425,7 @@ class BufferedOutputBase(io.BufferedIOBase):
         logger.debug("closing")
         if self._total_size == 0:  # empty files
             self._upload_empty_part()
-        if self._current_part.tell():
+        else:
             self._upload_final_part()
         logger.debug("successfully closed")
 
@@ -454,6 +455,7 @@ class BufferedOutputBase(io.BufferedIOBase):
         self._current_part.write(b)
         self._total_size += len(b)
 
+        # TODO: Change this back to if statement
         while self._current_part_size >= self._min_part_size:
             self._upload_next_part()
 
@@ -473,6 +475,7 @@ class BufferedOutputBase(io.BufferedIOBase):
         part_num = self._total_parts + 1
 
         # we can only upload chunks in multiples of 256kB, except for the final part
+        # TODO: Upload the largest chunk possible
         content_length = self._min_part_size
         total_size = self._total_size - (self._total_size % self._min_part_size)
 
@@ -482,8 +485,8 @@ class BufferedOutputBase(io.BufferedIOBase):
             content_length,
             total_size / 1024.0 ** 3
         )
-        start = total_size - content_length
-        stop = total_size - 1
+        start = self._bytes_uploaded
+        stop = start + content_length - 1
 
         self._current_part.seek(0)
 
@@ -491,6 +494,7 @@ class BufferedOutputBase(io.BufferedIOBase):
             'Content-Length': str(content_length),
             'Content-Range': _make_range_string(start, stop, _UNKNOWN_FILE_SIZE),
         }
+        logger.info('Sent headers ' + str(headers))
         response = self._session.put(
             self._resumable_upload_url,
             data=self._current_part.read(content_length),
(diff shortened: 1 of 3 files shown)


petedannemann:
@mpenkov do you think you will have the chance to review this soon? Fixing this bug would be great if we plan to release the GCS feature (https://github.com/RaRe-Technologies/smart_open/issues/423)

smart_open/gcs.py
```diff
@@ -399,6 +400,8 @@ def __init__(
 
         self._total_size = 0
         self._total_parts = 0
+        self._bytes_uploaded = 0
+        self._closed = False
```
mpenkov: Do we really need this as a separate variable? Can we not use `self._client is None` to mean we're closed, and assume open otherwise?

smart_open/gcs.py
```diff
@@ -453,7 +462,7 @@ def write(self, b):
         self._current_part.write(b)
         self._total_size += len(b)
 
-        if self._current_part.tell() >= self._min_part_size:
+        if self._current_part_size > self._min_part_size:
```
mpenkov: Why is this necessary? Why is this better?

smart_open/gcs.py
```diff
@@ -470,52 +479,136 @@ def terminate(self):
...
+            data=self._current_part.read(content_length),
+            headers=headers,
+        )
+
+        if response.status_code != _UPLOAD_INCOMPLETE_STATUS_CODE:
             msg = (
                 "upload failed ("
-                "status code: %i"
-                "response text=%s, "
+                "status code: %i "
```
mpenkov: This formatting looks a little awkward, and you perform it in multiple places. Do:

```python
raise UploadFailedError.from_response(response, ...)
```

and implement the above as a `classmethod`.

smart_open/gcs.py
```diff
@@ -470,52 +479,136 @@ def terminate(self):
...
         headers = {'Content-Length': '0'}
         response = self._session.put(self._resumable_upload_url, headers=headers)
-        assert response.status_code in _SUCCESSFUL_STATUS_CODES
+        assert response.status_code in _UPLOAD_COMPLETE_STATUS_CODES
 
         self._total_parts += 1
 
+    @property
+    def _current_part_size(self):
+        return self._current_part.seek(0, io.SEEK_END)
```
mpenkov: This property has a nasty side effect - it seeks to the end of the current part, and this side effect is completely obscured by its name. This could cause problems in the future, for example, if someone uses this property in log statements or otherwise.

I think the way forward is to make this a regular method (or even a pure function, as it doesn't really use the full object) and rename this as `_finalize_current_part` to make it obvious that it modifies the buffer significantly.

You could also undo the side effect before returning, but I think the above alternative is better.

smart_open/tests/test_gcs.py
```diff
@@ -172,6 +170,7 @@ def exists(self, client=None):
         return self._exists
 
     def upload_from_string(self, str_):
+        self.__contents = io.BytesIO()
```
mpenkov: You can be more terse:

```suggestion
        self.__contents = io.BytesIO(str_)
```

Also, the method name is misleading. You're talking about strings, but you're actually writing bytes. If this is part of Google's API, then we can't really do much, other than leave a comment here to console people working on this in the future.

smart_open/gcs.py
petedannemann: Once we have uploaded the final part (signified by a content range with an end), any additional put requests lead to a hanging response that takes a while to timeout. Saving state of the completeness of the upload was done to avoid this hanging response.

mpenkov: OK, then please set `self._client = None` instead of using `self._closed = True`.

smart_open/tests/test_gcs.py
petedannemann: This is part of Google's API, the method names are misleading on their end. I added comments and improved variable names to document this

smart_open/gcs.py
```diff
@@ -91,6 +91,25 @@ def __init__(self, message, status_code, text):
         self.status_code = status_code
         self.text = text
 
+    @classmethod
+    def from_response(cls, response, part_num, content_length, total_size, headers):
+        msg = (
+            "upload failed ("
```
mpenkov: This line splitting is a bit excessive.

You can also simplify the formatting code by setting the required local variables and then using `locals()`.

smart_open/gcs.py
```diff
@@ -561,7 +570,7 @@ def _upload_final_part(self):
           "%i bytes (total %.3fGB)"
           "headers %r",
           part_num,
-          part_size,
+          content_length,
```
mpenkov: The formatstring can fit on one line.

Please use indendation that is a multiple of 4, here and everywhere.

smart_open/gcs.py
```diff
@@ -453,7 +465,7 @@ def write(self, b):
         self._current_part.write(b)
         self._total_size += len(b)
 
-        if self._current_part.tell() >= self._min_part_size:
+        if self._current_part.tell() > self._min_part_size:
```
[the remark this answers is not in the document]
petedannemann: The headers for a final part and a non-final part are different. Headers for final parts are set in `_upload_final_part` while headers for non-final parts are set in `_upload_next_part`. If we want to upload exactly 256kB we will actually want to use `_upload_final_part` and this greater than makes sure that happens. I tried to document this [here](https://github.com/RaRe-Technologies/smart_open/pull/421/files#diff-9d5a27c1afe22c7747d606a2f131904eR486)

mpenkov:
```suggestion
        #
        # If the size of this part is precisely equal to the minimum part size,
        # we don't perform the actual write now, and wait until we see more data.
        # We do this because the very last part of the upload must be handled slightly
        # differently (see comments in the _upload_next_part method).
        #
        if self._current_part.tell() > self._min_part_size:
```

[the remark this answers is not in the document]
mpenkov: Alright, I've added a comment, because the connection wasn't immediately obvious.

mpenkov:
Thank you for your patience. Merging.

PR #426, opened 2020-03-09
implemented efficient readline for ByteBuffer

### LINKED ISSUE #371: BufferedInputBase readline reading buffer twice
Hi,

`BufferedInputBase` may run through the buffer multiple times while running `readline`:

https://github.com/RaRe-Technologies/smart_open/blob/1770807b0c8da2a34f749afa17ff70a715bc85de/smart_open/s3.py#L307-L315

I think we can remove this inefficiency if we use `find` instead of `index`.  E.g.:

```python
while not (self._eof and len(self._buffer) == 0):
    remaining_buffer = self._buffer.peek()
    terminator_index = remaining_buffer.find(self._line_terminator)
    if terminator_index > -1:
        the_line.write(self._read_from_buffer(terminator_index + 1))
        break
    else:
        the_line.write(self._read_from_buffer())
        self._fill_buffer()
```

Does this make sense or am I missing something?

-- Eric

mpenkov:
#### Motivation

The previous implementation searched for the terminator in the buffer twice. This was unnecessary.

The new implementaiton performs the search only once.

- Fixes #371 

#### Checklist

### DIFF
--- a/smart_open/bytebuffer.py
+++ b/smart_open/bytebuffer.py
@@ -159,3 +159,22 @@ class ByteBuffer(object):
 
         self._bytes += new_bytes
         return len(new_bytes)
+
+    def readline(self, terminator):
+        """Efficiently reads a line from this buffer.
+
+        A line is a contiguous sequence of bytes that ends with either:
+
+        1. The ``terminator`` character
+        2. The end of the buffer itself
+
+        :param byte terminator: The line terminator character.
+        :rtype: bytes
+
+        """
+        index = self._bytes.find(terminator, self._pos)
+        if index == -1:
+            size = len(self)
+        else:
+            size = index - self._pos + 1
+        return self.read(size)
--- a/smart_open/bytebuffer.py
+++ b/smart_open/bytebuffer.py
@@ -161,7 +161,7 @@ class ByteBuffer(object):
         return len(new_bytes)
 
     def readline(self, terminator):
-        """Efficiently reads a line from this buffer.
+        """Read a line from this buffer efficiently.
 
         A line is a contiguous sequence of bytes that ends with either:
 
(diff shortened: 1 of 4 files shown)


smart_open/bytebuffer.py
```diff
@@ -159,3 +159,22 @@ def fill(self, source, size=-1):
 
         self._bytes += new_bytes
         return len(new_bytes)
+
+    def readline(self, terminator):
+        """Efficiently reads a line from this buffer.
```
piskvorky:
```suggestion
        """Read a line from this buffer efficiently.
```

(docstrings in imperative mode)

smart_open/bytebuffer.py
```diff
@@ -159,3 +159,22 @@ def fill(self, source, size=-1):
 
         self._bytes += new_bytes
         return len(new_bytes)
+
+    def readline(self, terminator):
+        """Read a line from this buffer efficiently.
+
+        A line is a contiguous sequence of bytes that ends with either:
+
+        1. The ``terminator`` character
```
piskvorky: How does this play with Python's "universal newline" functionality? Is it fully compatible?

IIRC Python considers multiple byte terminators as possible line ending: `\n`, `\r`, `\r\n`.

mpenkov: We don't support universal newlines at all ().

```python
    if mode not in ('rb', 'rb+', 'wb', 'wb+', 'ab', 'ab+'):
        #
        # This should really be a ValueError, but for the sake of compatibility
        # with older versions, which raise NotImplementedError, we do the same.
        #
        raise NotImplementedError('unsupported mode: %r' % mode)
```

mpenkov:
Please check out the benchmarks above. With 5 runs to read 1 million lines, the minimum time hasn't decreased all that much (from 38 to 36s), but the max time changed significantly (from 58 to 40s).

PR #427, opened 2020-03-09
improve buffering efficiency

### LINKED ISSUE #348: S3 BufferedInputBase fills buffer during every read

mpenkov:
#### Motivation

This fixes an edge case where the user performs multiple sequential reads of a small number of bytes, e.g. one byte at a time. The previous implementation would fill the buffer one byte a time, which negates the benefit of using a buffer.

The new implementation fixes the above problem by always reading in chunks that are larger than a sensible threshold (currently equal to io.DEFAULT_BUFFER_SIZE).

- Fixes #348

#### Checklist

### DIFF
--- a/integration-tests/test_s3_buffering.py
+++ b/integration-tests/test_s3_buffering.py
@@ -1,4 +1,3 @@
-import sys
 from smart_open import open
 
 
--- a/integration-tests/test_s3_buffering.py
+++ b/integration-tests/test_s3_buffering.py
@@ -1,0 +1,24 @@
+import sys
+from smart_open import open
+
+
+def read_bytes(url, limit):
+    bytes_ = []
+    with open(url, 'rb') as fin:
+        for i in range(limit):
+            bytes_.append(fin.read(1))
+
+    return bytes_
+
+
+def test(benchmark):
+    #
+    # This file is around 850MB.
+    #
+    url = (
+        's3://commoncrawl/crawl-data/CC-MAIN-2019-51/segments/1575541319511.97'
+        '/warc/CC-MAIN-20191216093448-20191216121448-00559.warc.gz'
+    )
+    limit = 1000000
+    bytes_ = benchmark(read_bytes, url, limit)
+    assert len(bytes_) == limit
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -314,10 +314,6 @@ class Reader(io.BufferedIOBase):
         if self._eof:
             return self._read_from_buffer()
 
-        #
-        # Fill our buffer to the required size.
-        #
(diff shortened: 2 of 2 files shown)


mpenkov:
Probably for a while now, as long as I can remember, anyway.

I'm not sure what the actual impact of this bug is. I'll run some benchmarks later and let you know if anything interesting comes up.

mpenkov:
I haven't been able to measure any significant benefit in doing this. I suspect something else outside of our control is performing its own buffering.

mpenkov:
Despite the lack of performance improvement, I think we should merge this anyway, as the new way of doing things makes more sense.

Let me know if you think otherwise.

piskvorky:
What's the performance before/after?

mpenkov:
It's in the commit message for the "add benchmarks" commit.

```
Before:

------------------------------------------- benchmark: 1 tests -------------------------------------------
Name (time in s)        Min      Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------
test                 4.8925  10.1093  5.9906  2.3032  5.0104  1.3963       1;1  0.1669       5           1
----------------------------------------------------------------------------------------------------------

After:

------------------------------------------- benchmark: 1 tests ------------------------------------------
Name (time in s)        Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------
test                 4.9611  9.7707  5.9822  2.1190  5.0280  1.3168       1;1  0.1672       5           1
---------------------------------------------------------------------------------------------------------
```

piskvorky:
Is boto is doing some buffering internally? How else could reading one-byte-at-a-time be the same speed? Strange.

If so, maybe we shouldn't buffer at all, and leave it to boto instead.

mpenkov:
Yes, I also suspect boto3 does its own buffering. We could investigate, and remove our own buffering and compare performance, but that's a much larger change than the current PR. I'd rather deal with that separately, when we have more time.

PR #433, opened 2020-03-15
fix WebHDFS read method

### LINKED ISSUE #388: WebHDFS readline() for single line not working
The read() function in webhdfs.py doen't handle a certain case:
```python
    def read(self, size=None):
        if size is None:
            self._buf, retval = b'', self._buf + self._response.raw.read()
            return retval
        elif size < len(self._buf):
            self._buf, retval = self._buf[size:], self._buf[:size]
            return retval
          <<<<<<<<<<< case not handled where size > _buf


        try:
            while len(self._buf) < size:
                self._buf += self._response.raw.readline()
        except StopIteration:
            pass

        self._buf, retval = self._buf[size:], retval[:size] <<<<<<<<<<<< causes retval to not be set as a variable and can't be used here
        return retval
```

When I used readline(), the code goes through the read() function and looking at it, my size was larger than the buffer. This caused the retval to never get set, so later on line ~93, retval is used but never set.

Output:
```
  File "helper.py", line 88, in convert_csv_to_json
    header = csv_file.readline()
  File "/usr/lib/python3.6/codecs.py", line 557, in readline
    data = self.read(readsize, firstline=True)
  File "/usr/lib/python3.6/codecs.py", line 497, in read
    newdata = self.stream.read(size)
  File "/usr/local/lib/python3.6/dist-packages/smart_open/webhdfs.py", line 93, in read
    self._buf, retval = self._buf[size:], retval[:size]
UnboundLocalError: local variable 'retval' referenced before assignment
```

Versions:
Python 3.6
smart_open 1.9

mpenkov:
Fix #388 

The existing WebHDFS implementation was poorly tested. I fixed the bug, and added some unit tests.

#### Checklist

### DIFF
--- a/smart_open/webhdfs.py
+++ b/smart_open/webhdfs.py
@@ -114,11 +114,11 @@ class BufferedInputBase(io.BufferedIOBase):
 
         try:
             while len(self._buf) < size:
-                self._buf += self._response.raw.readline()
+                self._buf += self._response.raw.read(io.DEFAULT_BUFFER_SIZE)
         except StopIteration:
             pass
 
-        self._buf, retval = self._buf[size:], retval[:size]
+        self._buf, retval = self._buf[size:], self._buf[:size]
         return retval
 
     def read1(self, size=-1):
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ script:
   # The creds won't be available unless this is a branch belonging to the
   # RaRe-Technologies organization.
   #
-  - if [[ ${TRAVIS_SECURE_ENV_VARS} && ${RUN_BENCHMARKS} ]]; then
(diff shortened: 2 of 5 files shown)


smart_open/webhdfs.py
```diff
@@ -114,11 +114,11 @@ def read(self, size=None):
 
         try:
             while len(self._buf) < size:
-                self._buf += self._response.raw.readline()
+                self._buf += self._response.raw.read(io.DEFAULT_BUFFER_SIZE)
         except StopIteration:
             pass
 
-        self._buf, retval = self._buf[size:], retval[:size]
+        self._buf, retval = self._buf[size:], self._buf[:size]
```
piskvorky: Wow. What was this doing before?

mpenkov: I think it worked in the base case only (read all content from beginning to end).

PR #436, opened 2020-03-21
Add pathlib monkeypatch with replacement of `pathlib.Path.open`

menshikh-iv:
## Problem

If you use `pathlib.Path` instead of `str` in your code, you should use the internal `open` method like

```python
with mypath.open("r") as infile:
    ...
```
of course, you can pass them to `smart_open` like this

```python
with smart_open.open(mypath, 'r') as infile:
    ...
```
but this doesn't looks "good enough"

## Solution

I implement "mokeypatch" function that replaces `pathlib.Path.open` to `smart_open.open`

## Example

```python
from pathlib import Path
from smart_open.smart_open_lib import patch_pathlib

patch_pathlib()

pth = Path("/dev/random")

with pth.open("rb") as infile:
    print(pth.read(4))
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -395,13 +395,13 @@ Now you can natively use ``smart_open.open`` with your ``Path`` objects
   >> patch_pathlib()  # replace `Path.open` with `smart_open.open`
   >>
   >> path = Path("smart_open/tests/test_data/crime-and-punishment.txt.gz")
-  >>>
+  >>
   >> with path.open("r") as infile:
   ..     # not possible with standard `Path.open` (because gzipped),
   ..     # but works perfectly with "patched" version by `smart_open`
   ..     for line in infile:
   ..         print(line)
-
+  ..         break
 
 Comments, bug reports
 =====================
--- a/README.rst
+++ b/README.rst
@@ -392,7 +392,7 @@ Now you can natively use ``smart_open.open`` with your ``Path`` objects
   >>> from pathlib import Path
   >>> from smart_open.smart_open_lib import patch_pathlib
   >>>
-  >>> patch_pathlib()  # replace `Path.open` with `smart_open.open`
+  >>> _ = patch_pathlib()  # replace `Path.open` with `smart_open.open`
   >>>
   >>> path = Path("smart_open/tests/test_data/crime-and-punishment.txt.gz")
   >>>
@@ -400,8 +400,9 @@ Now you can natively use ``smart_open.open`` with your ``Path`` objects
   ...     # not possible with standard `Path.open` (because gzipped),
   ...     # but works perfectly with "patched" version by `smart_open`
   ...     for line in infile:
-  ...         print(line)
+  ...         print(repr(line))
   ...         break
+  'В начале июля, в чрезвычайно жаркое время, под вечер, один молодой человек вышел из своей каморки, которую нанимал от жильцов в С -- м переулке, на улицу и медленно, как бы в нерешимости, отправился к К -- ну мосту.\n'
 
 Comments, bug reports
 =====================
--- a/README.rst
+++ b/README.rst
@@ -389,19 +389,19 @@ Now you can natively use ``smart_open.open`` with your ``Path`` objects
 
 .. code-block:: python
 
-  >> from pathlib import Path
-  >> from smart_open.smart_open_lib import patch_pathlib
-  >>
-  >> patch_pathlib()  # replace `Path.open` with `smart_open.open`
-  >>
-  >> path = Path("smart_open/tests/test_data/crime-and-punishment.txt.gz")
-  >>
-  >> with path.open("r") as infile:
-  ..     # not possible with standard `Path.open` (because gzipped),
-  ..     # but works perfectly with "patched" version by `smart_open`
-  ..     for line in infile:
-  ..         print(line)
-  ..         break
+  >>> from pathlib import Path
+  >>> from smart_open.smart_open_lib import patch_pathlib
+  >>>
+  >>> patch_pathlib()  # replace `Path.open` with `smart_open.open`
+  >>>
+  >>> path = Path("smart_open/tests/test_data/crime-and-punishment.txt.gz")
+  >>>
+  >>> with path.open("r") as infile:
+  ...     # not possible with standard `Path.open` (because gzipped),
+  ...     # but works perfectly with "patched" version by `smart_open`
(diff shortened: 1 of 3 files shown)


mpenkov:
Looks good to me! Merged. 

Thank you for your contribution @menshikh-iv !!

smart_open/smart_open_lib.py
```diff
@@ -928,3 +928,28 @@ def _encoding_wrapper(fileobj, mode, encoding=None, errors=None):
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        _patch_pathlib(self.old_impl)
+
+
+def _patch_pathlib(func):
+    """Replace `Path.open` with `func`"""
+    pathlib = sys.modules.get("pathlib", None)
+
+    if not pathlib:
+        raise RuntimeError("Can't patch 'pathlib.Path.open', you should import 'pathlib' first")
```
piskvorky: Why not import it ourselves?

smart_open/tests/test_smart_open.py
```diff
@@ -287,6 +289,35 @@ def test_gs_uri_contains_slash(self):
...
+
+    def test_pathlib_monkeypath_read_gz(self):
+        path = pathlib.Path(CURR_DIR) / 'test_data' / 'crime-and-punishment.txt.gz'
+
+        # Check that standart implementation can't work with gzip
+        with path.open("r") as infile:
+            with self.assertRaises(Exception) as context:
+                lines = infile.readlines()
+
+        # Check that out implementation works with gzip
```
piskvorky: Typo: 

```
# Check that our implementation works with gzip
```

smart_open/tests/test_smart_open.py
```diff
@@ -287,6 +289,35 @@ def test_gs_uri_contains_slash(self):
...
+        obj = patch_pathlib()
+        with path.open("r") as infile:
+            lines = infile.readlines()
+
+        _patch_pathlib(obj.old_impl)
+
```
piskvorky: Can you add a test for the context manager? That should be the preferred way of restoring the original, instead of calling `_patch_pathlib(obj.old_impl)`.

piskvorky:
Sorry for late review, I see this was already merged. Needs a little language + API clean up.

smart_open/smart_open_lib.py
menshikh-iv: We shouldn't do that ourselves, this is bad tone.

BTW - this doesn't help to the user, because the user will not have `Path` class available after our import

smart_open/tests/test_smart_open.py
menshikh-iv: I added context manager for testing purposes, I see no cases for "real usage" of context manager in this case

README.rst
```diff
@@ -382,6 +382,25 @@ If your file object doesn't have one, set the ``.name`` attribute to an appropri
 Furthermore, that value has to end with a **known** file extension (see the ``register_compressor`` function).
 Otherwise, the transparent decompression will not occur.
 
+Drop-in replacement of ``pathlib.Path.open``
+--------------------------------------------
+
+Now you can natively use ``smart_open.open`` with your ``Path`` objects.
```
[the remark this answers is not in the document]
menshikh-iv: OK with first part, but definitely agains of context-manager usage in this case, because:

*your variant*
```python
with patch_pathlib():
    with Path("smart_open/tests/test_data/crime-and-punishment.txt.gz").open("r") as infile:
        print(infile.readline()[:41])
```

*as I'll do that*
```python
with smart_open("smart_open/tests/test_data/crime-and-punishment.txt.gz") as infile:
    print(infile.readline()[:41])
```

shorter and simpler

smart_open/smart_open_lib.py
piskvorky: I don't follow – imported modules are singletons, it doesn't matter who imported it.

README.rst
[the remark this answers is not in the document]
piskvorky: That's true.

smart_open/smart_open_lib.py
menshikh-iv: that's true, but user has no access to `pathlib` itself (without digging into `sys.modules`), like 

```python
In [1]: def f(): 
   ...:     import pathlib 
   ...:                                                                         

In [2]: pathlib                                                                 
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-7bc182bbe5df> in <module>
----> 1 pathlib

NameError: name 'pathlib' is not defined
```

I expect that user UNDERSTAND what they do, for this reason, I expect than user firstly import `pathlib` and after apply monkeypathing.

piskvorky: I still don't get it. What's the point of failing with "import pathlib first!", when we can import it ourselves?

piskvorky: …plus `pathlib` already seems to be pre-imported internally:

```python
>>> import sys
>>> sys.modules.get("pathlib", None)
<module 'pathlib' from '/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pathlib.py'>
```

So when can this `if` condition actually happen? Is it some Python version compatibility thing?
If so, needs better comments (and a different exception message).

piskvorky:
@mpenkov the `smart_open` badge shows "build failing" after this merge – can you have a look? Thanks.

smart_open/smart_open_lib.py
mpenkov: In my environment:

```
(smart_open) sergeyich:smart_open misha$ python
Python 3.7.6 (default, Dec 30 2019, 19:38:28)
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.modules.get('pathlib', 'not there')
'not there'
```

However, `smart_open` already imports pathlib by itself: https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/smart_open_lib.py#L51

So, given that we're already doing it, it's probably simpler to patch it directly.

mpenkov:
@menshikh-iv @piskvorky Please see #437, I've incorporated the post-merge comments into there.

smart_open/smart_open_lib.py
menshikh-iv: Ok @mpenkov, if we do that anyway, I agree.

PR #437, opened 2020-03-22
Fixup

mpenkov:
Continued from https://github.com/RaRe-Technologies/smart_open/pull/436

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -945,6 +945,8 @@ class patch_pathlib(object):
 
 def _patch_pathlib(func):
     """Replace `Path.open` with `func`"""
+    if not PATHLIB_SUPPORT:
+        raise RuntimeError('install pathlib (or pathlib2) before using this function')
     old_impl = pathlib.Path.open
     pathlib.Path.open = func
     return old_impl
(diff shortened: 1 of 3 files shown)


smart_open/smart_open_lib.py
```diff
@@ -945,11 +945,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 def _patch_pathlib(func):
     """Replace `Path.open` with `func`"""
-    pathlib = sys.modules.get("pathlib", None)
```
menshikh-iv: are you sure about it? as I see in head of the file

https://github.com/RaRe-Technologies/smart_open/blob/dfcaf2e7134a220954ab69e3e064d297f47cd76d/smart_open/smart_open_lib.py#L47-L55

i.e. `pathlib` can be not defined

menshikh-iv: `pathlib2` (for py2?) installed only in test env 

https://github.com/RaRe-Technologies/smart_open/blob/90d4f72065c28ef9bd462a18f6d86950e1d7fdaf/setup.py#L42-L45

so, in case if person use `py2` and install smart_open from pypi (with "normal" mode),  `pathlib` can be undefined. For `py3.5+` no issues here (batteries inside).

PR #438, opened 2020-03-22
Add doctest to our CI workflow

mpenkov:
This will automatically run `doctest` on README.rst for each build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
 script:
   - flake8 --max-line-length=110
   - python check_keys.py
-  - python -m doctest README.rst
+  - python -m doctest README.rst -v
 
   - if [[ ${SO_ENABLE_MOTO_SERVER} = 1 ]]; then
       sh -c "moto_server -p5000 2> /dev/null &";
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,15 @@ install:
 script:
   - flake8 --max-line-length=110
   - python check_keys.py
-  - python -m doctest README.rst -v
+
+  #
+  # Our doctests require valid boto3 credentials.  They won't work without them.
+  # This is by design: we don't want to cram our README.rst with programming
+  # logic to handle the absence of credentials.
+  #
+  - if [[ "${TRAVIS_SECURE_ENV_VARS}" == "true" ]]; then
+      python -m doctest README.rst -v;
+    fi
 
   - if [[ ${SO_ENABLE_MOTO_SERVER} = 1 ]]; then
       sh -c "moto_server -p5000 2> /dev/null &";
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ install:
 script:
   - flake8 --max-line-length=110
   - python check_keys.py
+  - python -m doctest README.rst
 
   - if [[ ${SO_ENABLE_MOTO_SERVER} = 1 ]]; then
       sh -c "moto_server -p5000 2> /dev/null &";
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -41,6 +41,7 @@ install:
 script:
   - flake8 --max-line-length=110
   - python check_keys.py
+  - python -m doctest README.rst
```
menshikh-iv:
```suggestion
  - python -m doctest README.rst -v
```

I propose to add verbosity here to 
- better understanding, what's going on on this stage
- make sure than a fresh "piece of code" was interpreted as docstring (to avoid the situation that happened yesterday).

README.rst
```diff
@@ -133,15 +133,22 @@ More examples
 
 .. code-block:: python
 
-    >>> import boto3
+    >>> import os, sys, boto3
```
menshikh-iv: also, I thought that this is a good idea to get rid `repr` function that used for strings (we don't really need them), see my "pathlib-patching" example

README.rst
```diff
@@ -40,18 +40,12 @@ How?
   >>> from smart_open import open
   >>>
   >>> # stream lines from an S3 object
-  >>> for line in open('s3://commoncrawl/robots.txt'):
-  ...    print(repr(line))
-  ...    break
+  >>> open('s3://commoncrawl/robots.txt').readline()
```
piskvorky: Strong -1: show line iteration in the examples, as prominently as possible.

Why this change?

README.rst
```diff
@@ -133,15 +123,22 @@ More examples
 
 .. code-block:: python
 
-    >>> import boto3
+    >>> import os, sys, boto3
+    >>>
+    >>> def get_session():
+    ...     # This is here so we can doctest this example under TravisCI
```
piskvorky: Mild -1: the purpose of README is clear communication, not doctests.

If the two goals clash, clarity of message takes precedence. Doctests less important.

README.rst
```diff
@@ -313,7 +310,7 @@ The ``version_id`` transport parameter enables you to get the desired version of
 
   >>> # If you don't specify a version, smart_open will read the most recent one
   >>> with open('s3://%s/%s' % (bucket, key)) as fin:
-  ...     print(repr(fin.read()))
+  ...     fin.read()
```
piskvorky: Why this change? Intent less clear.

README.rst
menshikh-iv: btw if we have broken example in the docstring, that's bad for users, for this reason, I propose to make something similar as in gensim we done: some "internal util functions" inside package that "make dirty work" and just re-use them in docstring (similar stuff in gensim are `datapath`, small corpora & dictionaries, etc).

README.rst
mpenkov: Two reasons:

1. Brevity
2. A for-loop that breaks after the first iteration is redundant

README.rst
mpenkov: Again, brevity. `print(repr(...))` is the default behavior of a REPL session.

README.rst
mpenkov: We show line iteration in separate examples (e.g. the example below), so there's little value in showing it again here, in my opinion.

piskvorky: No, iteration should be the top-most (most visible, most prominent) example.

Also showing it "below" is fine too, to drive the point home.

README.rst
piskvorky: Our README goal is clarity, not brevity.

There is no assumption the code goes into REPL. Keep the `print` please.

README.rst
mpenkov: OK, reverted.

README.rst
mpenkov: OK, reverted.

Issue #216, opened 2018-08-07
Improve tooling for repository


menshikh-iv:
We need to improve the tooling of the repository to increase code-quality and simplify test process
- [x] Add style-checker (pycodestyle, flake8, any)
- [x] Add tox that should allow
  - [x] Run tests with all python version
  - [x] Run style-checkers
  - [x] (optional) Run integration tests

**Discussion:** @piskvorky @mpenkov wdyt to add automatic code-formatter like  https://github.com/google/yapf or https://github.com/ambv/black, I want to drop all discussions about code-style at all (honestly, I want to do it for Gensim, but `smart_open` are good candidate for test this approach, because it simple and small).

Why I think this is useful:
- auto code-formatting (no need to remember about style always)
- no comments about code-formatting
- always clean code

piskvorky:
Fine with me (of course), good idea. As long as we can enforce, or at least allow, hanging indent (not vertical) and lines longer than 80 chars.

PR #358, opened 2019-10-01
[WIP] Add tox

menshikh-iv:
Fix #216 

Add tox to the current project

Missing
- Execute the benchmark on CI (tox issues, commented for now)

PR #439, opened 2020-03-22
Add tox

menshikh-iv:
Continuation of https://github.com/RaRe-Technologies/smart_open/pull/358
Fix https://github.com/RaRe-Technologies/smart_open/issues/216

In addition, I fixed warnings from `.tavis` file validator


tox.ini
```diff
@@ -0,0 +1,80 @@
...
+    .[test]
+
+    integration: numpy
+
+    benchmark: pytest_benchmark
+    benchmark: awscli
+
+setenv =
+    SO_DISABLE_MOCKS={env:SO_DISABLE_MOCKS:}
+    SO_DISABLE_MOTO_SERVER={env:SO_DISABLE_MOTO_SERVER:}
```
mpenkov: This variable is now SO_ENABLE_MOTO_SERVER: https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/tests/test_s3.py#L35

Also, is this environment variable being used anywhere?

tox.ini
```diff
@@ -0,0 +1,80 @@
...
+    TRAVIS_SECURE_ENV_VARS={env:TRAVIS_SECURE_ENV_VARS:}
+    RUN_BENCHMARKS={env:RUN_BENCHMARKS:}
+
+commands =
+    test: pytest smart_open/tests/
+
+    integration: pytest integration-tests/test_http.py integration-tests/test_207.py
+    integration: bash -c 'if [[ "$TRAVIS_SECURE_ENV_VARS" = "true" ]]; then pytest integration-tests/test_s3_ported.py; fi'
+
+    benchmark: bash -c 'if [[ "$TRAVIS_SECURE_ENV_VARS" = "true" && "$RUN_BENCHMARKS" = "true" ]]; then export SO_S3_URL=$SO_S3_URL/$(python -c "from uuid import uuid4;print(uuid4())"); set -e; pytest integration-tests/test_s3.py --benchmark-save=`git rev-parse HEAD`; set +e; aws s3 cp .benchmarks/*/*.json $SO_S3_RESULT_URL; aws s3 rm --recursive $SO_S3_URL; fi'
```
mpenkov: Is there any way to make this prettier? Having it as a single long line is pain.

You could move out the git-rev-parse bit to a separate environment variable. I have a suspicion travis may already expose that for you.

The same thing goes for generating SO_S3_URL.

mpenkov: Also, can we move the conditionals somewhere else?

tox.ini
menshikh-iv: fixed 73476c5

tox.ini
menshikh-iv: > Is there any way to make this prettier? Having it as a single long line is pain.

`tox ` have issues with multiline *sh commands, sorry :( I already tried to make a multi-line version, but this doesn't work

> You could move out the git-rev-parse bit to a separate environment variable. I have a suspicion travis may already expose that for you.

Yes, I can, but in that case that will work only in travis (I can't fallback to `git rev-parse` in `setenv` section)

> Also, can we move the conditionals somewhere else?

Why?

mpenkov: It just looks a bit ugly and difficult to maintain.

How about making a separate bash script `ci/run_benchmarks.sh`? You could invoke it directly with `bash ci/run_benchmarks.sh` and it would read environment variables, etc. and do the hard work.

Similarly, with test_s3_ported, we could move the conditional into that script. Check that the creds are available in os.environ, and skip the tests if they are not.

menshikh-iv: I see 3 variants here:

Note: by "check env" I mean `if [[ ... = "true" && ]] then ... fi;` construction

1. Check env directly in `sh` script and call them from tox
2. Chek env in `tox.ini` and call `.sh`
3. Check env in `.travis` file and call .sh`

what are you prefer? I dislike (3), but (1) and (2) look fine for me.

mpenkov: I prefer 1).

mpenkov:
Good work @menshikh-iv ! Thank you!

PR #442, opened 2020-03-23
Fix error when calling str() or repr() on GCS SeekableBufferedInputBase

robcowie:
Fix `__repr__` and `__str__` on `gcs.SeekableBufferedInputBase` by storing the bucket as `_bucket` instead of `bucket`.

This is now consistent with [`gcs.BufferedOutputBase`](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/gcs.py#L399) and [`s3. _SeekableRawReader`](https://github.com/RaRe-Technologies/smart_open/blob/master/smart_open/s3.py#L197) (which stores its S3 client as `_object`)

#### Motivation

We were seeing logging errors originating from smart_open:

```python
  File "/usr/local/lib/python3.8/logging/__init__.py", line 1081, in emit msg = self.format(record)
  File "/usr/local/lib/python3.8/logging/__init__.py", line 925, in format return fmt.format(record)
  File "/usr/local/lib/python3.8/logging/__init__.py", line 664, in format record.message = record.getMessage()
  File "/usr/local/lib/python3.8/logging/__init__.py", line 369, in getMessage msg = msg % self.args
  File "/usr/local/lib/python3.8/gzip.py", line 220, in __repr__ s = repr(self.fileobj)
  File "/usr/local/lib/python3.8/site-packages/smart_open/gcs.py", line 379, in __repr__ self.__class__.__name__, self._bucket.name, self._blob.name, self._current_part_size, AttributeError: 'SeekableBufferedInputBase' object has no attribute '_bucket'
```

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -213,7 +213,8 @@ class SeekableBufferedInputBase(io.BufferedIOBase):
         self._blob = client.get_bucket(bucket).get_blob(key)  # type: google.cloud.storage.Blob
 
         if self._blob is None:
-            raise google.cloud.exceptions.NotFound('blob {} not found in {}'.format(key, bucket))
+            raise google.cloud.exceptions.NotFound('blob %s not found in %s' % (key, bucket))
+
         self._size = self._blob.size if self._blob.size is not None else 0
 
         self._raw_reader = _SeekableRawReader(self._blob, self._size)
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -209,11 +209,11 @@ class SeekableBufferedInputBase(io.BufferedIOBase):
     ):
         if client is None:
             client = google.cloud.storage.Client()
-        bucket = client.get_bucket(bucket)  # type: google.cloud.storage.Bucket
+        self._bucket = client.get_bucket(bucket)  # type: google.cloud.storage.Bucket
 
-        self._blob = bucket.get_blob(key)
+        self._blob = self._bucket.get_blob(key)
         if self._blob is None:
-            raise google.cloud.exceptions.NotFound('blob {} not found in {}'.format(key, bucket))
+            raise google.cloud.exceptions.NotFound('blob {} not found in {}'.format(key, self._bucket.name))
         self._size = self._blob.size if self._blob.size is not None else 0
 
         self._raw_reader = _SeekableRawReader(self._blob, self._size)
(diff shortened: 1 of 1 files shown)


smart_open/gcs.py
```diff
@@ -209,11 +209,11 @@ def __init__(
     ):
         if client is None:
             client = google.cloud.storage.Client()
-        bucket = client.get_bucket(bucket)  # type: google.cloud.storage.Bucket
+        self._bucket = client.get_bucket(bucket)  # type: google.cloud.storage.Bucket
```
piskvorky: We don't want to store complex objects (`bucket`) as internal attributes, unless we have to. Why not just store the bucket name = string?

Do we really need the full `bucket` for `__str__` or other methods?

robcowie: That sounds reasonable. We do not need the bucket object in the current code. It's available on the blob anyway (`Blob.bucket`).

I'll store the user-provided string as `self.bucket_name` and use that for rendering the repr. 

I'll also make the corresponding change in `BufferedOutputBase`

piskvorky: If it's already in the blob, can't we just use `self._blob.bucket.name`?

smart_open/gcs.py
```diff
@@ -209,11 +209,13 @@ def __init__(
...
         if client is None:
             client = google.cloud.storage.Client()
+
+        self._bucket_name = bucket
         bucket = client.get_bucket(bucket)  # type: google.cloud.storage.Bucket
 
         self._blob = bucket.get_blob(key)
         if self._blob is None:
-            raise google.cloud.exceptions.NotFound('blob {} not found in {}'.format(key, bucket))
+            raise google.cloud.exceptions.NotFound('blob {} not found in {}'.format(key, self._bucket_name))
```
piskvorky: Unrelated to your PR but we avoid `.format()` in our code (style preference). Not sure how this sneaked in, but let's take this opportunity to fix it:

```suggestion
            raise google.cloud.exceptions.NotFound('blob %s not found in %s' % (key, self._blob.bucket.name))
```

smart_open/gcs.py
robcowie: How do you feel about the following one-liner? By not assigning the bucket object,`bucket` remains the user-provided string and can be used in the `NotFound` exception.

```python
self._blob = client.get_bucket(bucket).get_blob(key)  # type: google.cloud.storage.Blob
```

piskvorky: Sure. Re-assigning the `bucket` variable to a different value and different type wasn't the best choice anyway :)

mpenkov:
Merged. Thank you @robcowie for your contribution!

PR #448, opened 2020-03-29
Minor cleanup in GCS code

mpenkov:
#### Motivation

- Reduce code duplication
- Shorter, more descriptive class names

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -492,7 +492,7 @@ class Writer(io.BufferedIOBase):
         # of parts being multiples of 256kB, except for the last one.
         #
         # A final upload of 0 bytes does not work, so we need to guard against
-        # this edge case this results in occasionally keeping an additional
+        # this edge case. This results in occasionally keeping an additional
         # 256kB in the buffer after uploading a part, but until this is fixed
         # on Google's end there is no other option.
         #
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -395,7 +395,6 @@ class Writer(io.BufferedIOBase):
         if client is None:
             client = google.cloud.storage.Client()
         self._client = client
-        self._credentials = self._client._credentials  # noqa
         self._blob = self._client.bucket(bucket).blob(blob)  # type: google.cloud.storage.Blob
         assert min_part_size % _REQUIRED_CHUNK_MULTIPLE == 0, 'min part size must be a multiple of 256KB'
         assert min_part_size >= _MIN_MIN_PART_SIZE, 'min part size must be greater than 256KB'
@@ -406,7 +405,7 @@ class Writer(io.BufferedIOBase):
         self._bytes_uploaded = 0
         self._current_part = io.BytesIO()
 
-        self._session = google.auth.transport.requests.AuthorizedSession(self._credentials)
+        self._session = google.auth.transport.requests.AuthorizedSession(client._credentials)
 
         #
         # https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload#start-resumable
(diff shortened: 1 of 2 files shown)


mpenkov:
@petedannemann Can you please have a look?

smart_open/gcs.py
```diff
@@ -39,17 +40,19 @@
 DEFAULT_BUFFER_SIZE = 256 * 1024
 """Default buffer size for working with GCS"""
 
-_UPLOAD_INCOMPLETE_STATUS_CODE = 308
+_UPLOAD_INCOMPLETE_STATUS_CODES = (308, )
```
petedannemann: Why this change?

smart_open/gcs.py
```diff
@@ -39,17 +40,19 @@
...
 _UPLOAD_COMPLETE_STATUS_CODES = (200, 201)
 
 
-def _make_range_string(start, stop=None, end=_UNKNOWN_FILE_SIZE):
+def _make_range_string(start, stop=None, end=None):
```
petedannemann: Is None as a default arg really cleaner than the actual default?

petedannemann:
Related to above, does it make sense long term for me to be the one running those tests? Would RaRe-Technologies be willing to create a Google Cloud account to run the tests on? I would be happy to help in any way with that, but I don't think it makes sense to have the account linked to me. I believe you get $300 in credits for the first year and billing information isn't even required during that time. I can't imagine these tests costing much money to run after that. I'm happy to open a new issue to discuss this further if that is more appropriate.

smart_open/gcs.py
mpenkov: To simplify the check against expected values in the upload_part function.

smart_open/gcs.py
mpenkov: Yeah, I think so. That asterisk is an implementation detail of the `make_range_string` function. You don't want to leak it to callers of the function. It's also more Pythonic to use None as opposed to a constant to indicate the absence of a value.

smart_open/gcs.py
```diff
@@ -482,103 +484,71 @@ def terminate(self):
...
-
-        # a final upload of 0 bytes does not work, so we need to guard against this edge case
-        # this results in occasionally keeping an additional 256kB in the buffer after uploading a part,
-        # but until this is fixed on Google's end there is no other option
+        #
+        # Here we upload the largest amount possible given GCS's restriction
+        # of parts being multiples of 256kB, except for the last one.
+        #
+        # A final upload of 0 bytes does not work, so we need to guard against
+        # this edge case this results in occasionally keeping an additional
```
piskvorky:
```suggestion
        # this edge case. This results in occasionally keeping an additional
```

smart_open/tests/test_gcs.py
```diff
@@ -371,7 +371,7 @@ def test_unfinished_put_does_not_write_to_blob(self):
             'Content-Length': str(4),
         }
         response = self.session.put(self.upload_url, data, headers=headers)
-        self.assertEqual(response.status_code, smart_open.gcs._UPLOAD_INCOMPLETE_STATUS_CODE)
+        self.assertEqual(response.status_code, smart_open.gcs._UPLOAD_INCOMPLETE_STATUS_CODES[0])
```
piskvorky: Why not:

```suggestion
        self.assertIn(response.status_code, smart_open.gcs._UPLOAD_INCOMPLETE_STATUS_CODES)
```

piskvorky:
@petedannemann we have bad experience with free time-limited trial accounts, because when they expire it's chaos.

Unfortunately we have zero budget for open source, this is all 100% voluntary work (which we sink a lot of our time into).

smart_open/gcs.py
```diff
@@ -13,7 +14,7 @@
 
 import google.cloud.exceptions
 import google.cloud.storage
-import google.auth.transport.requests as google_requests
+import google.auth.transport.requests
```
petedannemann: We need to update https://github.com/RaRe-Technologies/smart_open/blob/22761352d1013ed4340eec6c592f35eaadf2ed04/smart_open/tests/test_gcs.py#L463 to match this for mocks to work:

```
'google.auth.transport.requests',
```

mpenkov:
OK, thanks for your help!

PR #451, opened 2020-03-29
Allow accept_key function to use content metadata.

matt2000:
#### Title

Allow accept_key function to use content metadata.

#### Motivation

The goal here is to allow users to provide a more intelligent filtering function to iter_bucket that can consider the content metadata, and not just the key. For example, one could implement a function to include all files under a certain size, or updated since some date.

This is an alternative implementation to PR#324, which is likely more compatible (it should pass your test suite!) and may be better all-around; it's certainly more pythonic, IMO.

#### Checklist

- [n/a] Linked to any existing issues that your PR will be solving

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -955,7 +955,8 @@ def iter_bucket(
     prefix: str, optional
         Limits the iteration to keys starting wit the prefix.
     accept_key: callable, optional
-        This is a function that accepts a key name (unicode string) and
+        This is a function that accepts a key name (unicode string), and optionally as a 
+        second argument, a dictionary of content metadata provided by S3, and 
         returns True/False, signalling whether the given key should be downloaded.
         The default behavior is to accept all keys.
     key_limit: int, optional
@@ -1061,7 +1062,14 @@ def _list_bucket(
         else:
             for c in content:
                 key = c['Key']
-                if accept_key(key):
+                # Enable accept_key() to optionally take `content` as a second argument.
+                try:
+                    is_included = accept_key(key)
+                except TypeError:
+                    # Python raises TypeError when the number or arguments doesn't match the required number.
+                    is_included = accept_key(key, content)
+           
+                if is_included:
                     yield key
         ctoken = response.get('NextContinuationToken', None)
         if not ctoken:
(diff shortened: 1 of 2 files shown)


smart_open/tests/test_s3.py
```diff
@@ -538,6 +538,16 @@ def test_iter_bucket(self):
         results = list(smart_open.s3.iter_bucket(BUCKET_NAME))
         self.assertEqual(len(results), 10)
 
+    def test_iter_bucket_accept_key(self):
+        """ Ensure accept_key function can have 1 or 2 arguments."""
+        populate_bucket()
+        results = list(smart_open.s3.iter_bucket(BUCKET_NAME,
+            accept_key=lambda key: False))
```
mpenkov: No need to break lines here, our hard limit is 110 chars.

smart_open/s3.py
```diff
@@ -1061,7 +1062,14 @@ def _list_bucket(
         else:
             for c in content:
                 key = c['Key']
-                if accept_key(key):
+                # Enable accept_key() to optionally take `content` as a second argument.
```
mpenkov: I think it would be better to probe the accept_key function once, outside of the loops, and then call it as appropriate.

For the probe mechanism, I think checking the number of parameters explicitly is better than relying on TypeError, because it's possible (although unlikely) that the user's code can raise TypeError.

matt2000: OK, I agree. This makes the code a bit less readable, IMO, but it is probably significantly more performant. How about this approach?:
```
            if accept_key.__code__.co_argcount == 1:
                chooser = lambda k,c: accept_key(k)
            else:
                chooser = accept_key
            for c in content:
                key = c['Key']
                if chooser(key, c):
                    yield key
```

If this seems acceptable to you, I'll put it in and address your other comments.

mpenkov: I would do this:

```
def _wrap_for_compatibility(accept_key):
    def inner(key, content):
        return accept_key(key)
    return inner

#
# Handle legacy accept_key functions: they only accepted a single param.
#
if accept_key.__code__.co_argcount == 1:
    accept_key = _wrap_for_compatibility(accept_key)

for c in content:
    if accept_key(c['Key'], c):
        yield c['Key]
```

In such cases, it's better to avoid lambda. Lambda makes things a bit more magical here, and it's also [against PEP8](https://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def).

mpenkov:
Closing as stale

PR #454, opened 2020-04-01
Move optional dependencies to extras

### LINKED ISSUE #443: Don't package all cloud dependencies at once
#### Problem description

Smart open is a really useful library, but I find it a bit annoying to have all dependencies packaged with it. Most of the time one doesnt need to manipulate s3 and gcs and maybe azure storage if it were to be integrated in smart-open. 

#### What I wish I could do:

pip install smart-open would be the same behaviour as now
pip install smart-open[s3] would only install boto3 dependencies 
pip install smart-open[gcs] same for gcs
...

#### Note:

If you find it interesting I can assign this to myself and work on the project

BTW: i think it's the same behaviour for gensim, it packages boto3 but is not needed for common nlp tasks

Amertz08:
**IMPORTANT: this PR requires a migration path**

We should think about how and when we will expose users to this change.

#### Motivation

- Fixes #443 

#### Checklist

#### Workflow

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,7 +106,10 @@ Installation
 ------------
 ::
 
-    pip install smart_open
+    pip install smart_open  // Install with no cloud dependencies
+    pip install smart_open[aws] // Install AWS deps
+    pip install smart_open[gcp] // Install GCP deps
+    pip install smart_open[all] // Installs all cloud dependencies
 
 Or, if you prefer to install from the `source tar.gz <http://pypi.python.org/pypi/smart_open>`_::
 
@@ -117,6 +120,11 @@ To run the unit tests (optional), you'll also need to install some other depende
 The tests are also run automatically with `Travis CI <https://travis-ci.org/RaRe-Technologies/smart_open>`_ on every commit push & pull request.
 
 If you're upgrading from ``smart_open`` versions 1.8.0 and below, please check out the `Migration Guide <MIGRATING_FROM_OLDER_VERSIONS.rst>`_.
+
+Version `x.x.x` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
+was introduced in version `x.x.x`. If you want to maintain backwards compatibility (installing all dependencies) install this package via `smart_open[all]` now
+and once the change is made you should not have any issues. If all you care about is AWS dependencies for example you can install via `smart_open[aws]` and
+once the dependency change is made you will simply drop the unwanted dependencies.
 
 
 Built-in help
--- a/README.rst
+++ b/README.rst
@@ -124,7 +124,7 @@ If you're upgrading from ``smart_open`` versions 1.8.0 and below, please check o
 Version ``x.x.x`` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
 was introduced in version ``x.x.x``. If you want to maintain backwards compatibility (installing all dependencies) install this package via ``smart_open[all]`` now
 and once the change is made you should not have any issues. If all you care about is AWS dependencies for example you can install via ``smart_open[aws]`` and
-once the dependency change is made you will simply drop the unwanted dependencies.
+once the dependency change is made you will simply drop the unwanted dependencies. You can read more about the motivations `Here <https://github.com/RaRe-Technologies/smart_open/issues/443>`_
 
 
 Built-in help
--- a/README.rst
+++ b/README.rst
@@ -121,9 +121,9 @@ The tests are also run automatically with `Travis CI <https://travis-ci.org/RaRe
 
 If you're upgrading from ``smart_open`` versions 1.8.0 and below, please check out the `Migration Guide <MIGRATING_FROM_OLDER_VERSIONS.rst>`_.
 
-Version `x.x.x` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
-was introduced in version `x.x.x`. If you want to maintain backwards compatibility (installing all dependencies) install this package via `smart_open[all]` now
-and once the change is made you should not have any issues. If all you care about is AWS dependencies for example you can install via `smart_open[aws]` and
+Version ``x.x.x`` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
+was introduced in version ``x.x.x``. If you want to maintain backwards compatibility (installing all dependencies) install this package via ``smart_open[all]`` now
+and once the change is made you should not have any issues. If all you care about is AWS dependencies for example you can install via ``smart_open[aws]`` and
 once the dependency change is made you will simply drop the unwanted dependencies.
 
 
--- a/README.rst
+++ b/README.rst
@@ -124,7 +124,7 @@ If you're upgrading from ``smart_open`` versions 1.8.0 and below, please check o
 Version ``x.x.x`` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
 was introduced in version ``x.x.x``. If you want to maintain backwards compatibility (installing all dependencies) install this package via ``smart_open[all]`` now
 and once the change is made you should not have any issues. If all you care about is AWS dependencies for example you can install via ``smart_open[aws]`` and
-once the dependency change is made you will simply drop the unwanted dependencies. You can read more about the motivations `Here <https://github.com/RaRe-Technologies/smart_open/issues/443>`_
+once the dependency change is made you will simply drop the unwanted dependencies. You can read more about the motivations `here <https://github.com/RaRe-Technologies/smart_open/issues/443>`_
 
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -95,6 +104,7 @@ def read(fname):
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
+        'Programming Language :: Python :: 2.7',
```
mpenkov, line 107: We intentionally don't support 2.7 anymore.

piskvorky: Is this public? I see no mention of this (important, critical) information in the README.

If true, let's point to the "last version that works with 2.7" in the README and release notes, clearly and prominently.

Amertz08: Dropped that classifier.

Amertz08: Whatever you guys decide lmk and I can make a change. I did not indent on this PR covering that issue.

setup.py
```diff
@@ -53,10 +53,13 @@ def read(fname):
...
 install_requires = [
     'requests',
-    'boto3',
-    'google-cloud-storage',
 ]
 
+aws_deps = ['boto3']
+gcp_deps = ['google-cloud-storage']
+
+all_deps = install_requires + aws_deps + gcp_deps
```
mpenkov, line 61: GCP is still young, we can afford to quietly remove it without annoying too many people.

```suggestion
all_deps = install_requires + aws_deps
```

README.rst
```diff
@@ -118,6 +121,11 @@ The tests are also run automatically with `Travis CI <https://travis-ci.org/RaRe
 
 If you're upgrading from ``smart_open`` versions 1.8.0 and below, please check out the `Migration Guide <MIGRATING_FROM_OLDER_VERSIONS.rst>`_.
 
+Version ``x.x.x`` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
```
mpenkov, line 124:
```suggestion
Version ``2.0`` will introduce a backwards incompatible installation method with regards to the cloud dependencies. A migration path to minimize breaking
```

setup.py
```diff
@@ -81,10 +84,16 @@ def read(fname):
...
+    # Concatenating the lists together is temporary and will
+    # eventually simply be install_requires dropping the cloud
+    # dependencies from being installed without explicitly being declared.
+    install_requires=install_requires + aws_deps + gcp_deps,
     tests_require=tests_require,
     extras_require={
         'test': tests_require,
+        'aws': aws_deps,
+        'gcp': gcp_deps,
+        'all': all_deps
```
mpenkov, line 96:
```suggestion
        'all': all_deps,
```

Tyrannas:
I'm not sure about this but I think the imports in the code should now be in try except ? Like if someone installs the smart_open[gcs] the smart_open.s3 will throw an import error

Amertz08:
Test coverage keeps failing in 3.8. @Tyrannas technically not yet because this doesn't not install the AWS deps when you do `smart_open[gsc]` as they are still in the base requirements.

Amertz08:
We can still add it just so it's not forgotten though.

Tyrannas:
> We can still add it just so it's not forgotten though.

I was thinking something like:
```python
try:
    import boto3
except ImportError:
    logging.getLogger('smart_open').info("Boto3 dependency is not installed, smart_open s3 will therefore not be working.")
```

Using info instead of warning to avoid regular loggers to be spammed since logging level is by default warning.

What do you think?

mpenkov:
@Tyrannas I think we already do this:

https://github.com/RaRe-Technologies/smart_open/blob/29092ba823ba61028598dbe94324b750f19244a1/smart_open/transport.py#L43

The next step is to get the backends to automatically tell the users which packages they need, e.g. "to enable S3 support, run `pip install smart_open[s3]`".  Not sure what the best way to do that is, but it's probably out of scope for this PR.

Tyrannas:
> @Tyrannas I think we already do this:
> 
> https://github.com/RaRe-Technologies/smart_open/blob/29092ba823ba61028598dbe94324b750f19244a1/smart_open/transport.py#L43
> 
> The next step is to get the backends to automatically tell the users which packages they need, e.g. "to enable S3 support, run `pip install smart_open[s3]`". Not sure what the best way to do that is, but it's probably out of scope for this PR.

@mpenkov yeah but for example you are importing .s3 in smart_open __init__ file and .s3 imports boto3, so this will lead to an Import Error

Amertz08:
We can just use deprecation warnings. It's built into python.
https://docs.python.org/3.6/library/warnings.html

smart_open/s3.py
```diff
@@ -6,16 +6,23 @@
...
-import boto3
-import botocore.client
-import botocore.exceptions
+try:
+    import boto
+    import boto3
+    import botocore.client
+    import botocore.exceptions
+except ImportError:
+    sys.stderr.write("Install via smart_open[aws] or smart_open[all] to use this module")
```
Tyrannas, line 24: Why this instead of a logger.info() ? Otherwise the message will be displayed all the time for every library that is not installed

Tyrannas: or a least the s3 import needs to be taken out from __init__.py otherwise the actual version will sys.exit(1) without boto3 installed

Amertz08: The logger will only print if the root logger is configured which is no guarantee.

As far as that import path in __init__.py goes it would seem that it would need to be removed. Once those underlying dependencies are dropped from being installed I believe the import will cause issues. If sys.exit is dropped and the shortcut path maintained it would cause the import error to be suppressed here and thrown elsewhere. Whereas by maintaining the sys.exit the program exits exactly where the message to fix the issue is printed vs exiting with an import error elsewhere that doesn't immediately tell you what the issue is.

Amertz08: We would probably want a deprecation warning for the import path in this PR.

Amertz08: I was unaware of that import path so thank you for pointing that out.

Amertz08: Yeah so didn't see that in `transport.py` sys.exit will cause an issue every time. So yeah this is a bigger issue than just the short import path in `__init__.py`

Amertz08: Maybe move `register_transport('smart_open.s3')` to after the `try/except` block in each of the cloud modules?

Amertz08: ^ moving the register after try/except did not work so I moved it back. Also I modified the try/except block to reraise the `ImportError` as the `register_transport` would never see it w/o it.

mpenkov: We have two separate problems.

1. We import s3.iter_bucket in __init__.py. This will explode if boto3 is not installed. Currently, we _always_ install boto3 (for backwards compatibility) so it's not an acute problem.
2. When we try to register all the transports, we log a warning for each submodule we couldn't import. This will be annoying for people who intentionally didn't install dependencies for a particular submodule.

I think the solution for (1) is simple. When we stop installing boto3 by default, we can try-catch around the culprit import. We can also remove the import altogether, and direct people to use `smart_open.s3.iter_bucket` via a warning.

I think the solution for (2) is to group the submodules that failed to register together, and then log a single warning like:

```
The following submodules could not be imported due to missing dependencies: abs, s3, gcs. To use those submodules, you must install their dependencies.  See <URL> for more info.
```

Once people get used to the new extra handling (a few versions later), we can reduce the verbosity of the log message.

@Amertz08 I'd like to include your PR in a soon-to-be bugfix release of smart_open (we need to do one because the Google Cloud deps are messing things up for some people). Can you please roll it back to the state it was in when I approved it? I will then merge. We can continue our discussion and work in a later PR.

mpenkov: (Please roll back all commits after 0d86f9f. Don't force push, it will axe our discussions).

setup.py
```diff
@@ -53,10 +53,13 @@ def read(fname):
...
 install_requires = [
     'requests',
-    'boto3',
-    'google-cloud-storage',
 ]
 
+aws_deps = ['boto3']
+gcp_deps = ['google-cloud-storage']
+
+all_deps = install_requires + aws_deps
```
mpenkov, line 61:
```suggestion
all_deps = install_requires + aws_deps + gcp_deps
```

setup.py
```diff
@@ -81,10 +84,16 @@ def read(fname):
     license='MIT',
     platforms='any',
 
-    install_requires=install_requires,
+    # Concatenating the lists together is temporary and will
+    # eventually simply be install_requires dropping the cloud
+    # dependencies from being installed without explicitly being declared.
+    install_requires=install_requires + aws_deps + gcp_deps,
```
mpenkov, line 90:
```suggestion
    install_requires=install_requires + aws_deps,
```

mpenkov:
Merging. Thank you for the hard work @Amertz08 .

I've made additional issues #463 and #464 to continue the discussion @Tyrannas .

PR #460, opened 2020-04-04
Correctly handle GCS paths that contain '?' char 

### LINKED ISSUE #458: Cannot read files  with '?' character in the path when using google storage
#### Problem description

Cannot open files from google storage  when the file path has '?' character in it (which are legal in google storage). 

This was addressed for s3 in https://github.com/RaRe-Technologies/smart_open/issues/285 via PR https://github.com/RaRe-Technologies/smart_open/pull/291
The same fix is needed for Google Storage as well. 

#### Versions

Please provide the output of:

```
Linux-4.19.102+-x86_64-with-debian-9.11
Python 3.6.10 (default, Jan  4 2020, 03:29:08)
[GCC 6.3.0 20170516]
smart_open 1.10.0
```

chakruperitus:
#### Motivation

GCS (and s3) paths support the character '?' . Use of urllib.parse.urlsplit was splitting the path at '?' due to which such files cannot be opened.  This was already handled for S3 paths correctly. 
Moved the existing utility function in s3 to utils so that it can be called from both s3 and gcs implementations.

- Fixes #458 

#### Checklist

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -10,7 +10,6 @@
 
 import io
 import logging
-import urllib.parse
 
 import google.cloud.exceptions
 import google.cloud.storage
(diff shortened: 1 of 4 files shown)


mpenkov:
Good catch, thank you! And congratulations on your first PR to smart_open! 🥇

PR #476, opened 2020-04-16
Prevent smart_open from writing to logs on import

### LINKED ISSUE #474: Unable to load smart_open.gcs
I have an issue similar to this :  https://stackoverflow.com/questions/61126546/docker-compose-up-d-cannot-finish-because-an-error

Installation in docker works fine but when importing this error crops up.

The issue does not occur with 1.10, only with 1.11

Apologies for the total lack of details / reproducibility. I will try to add those tomorrow if possible.

### LINKED ISSUE #475: Importing gensim prints stuff and configures logging
#### Problem description

Not sure if this is a ticket for `smart_open` or `gensim`, but importing gensim prints out some (unwanted) stuff: `unable to import 'smart_open.gcs', disabling that module`.

I checked the `smart_open` source code and this is actually printed from a logger:

```
logger.warning('unable to import %r, disabling that module', submodule)
```

This is a big no-no – libraries have no business configuring the logging subsystem. That's up to the application. If an app tried to `import gensim` and then call `logging.basicConfig` then the latter will be ignored, because logging will already have been set up. Not good: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

Originally found via https://github.com/RaRe-Technologies/gensim/issues/2794 .

Screen Shot 2020-04-14 at 19 14 53

#### Versions

Please provide the output of:

```python
import platform, sys, smart_open
print(platform.platform())
print("Python", sys.version)
print("smart_open", smart_open.__version__)
```

```
Darwin-15.6.0-x86_64-i386-64bit
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)]
smart_open 1.11.1
```

#### Checklist

mpenkov:
#### Motivation

- Fix #474
- Fix #475

#### Checklist

### DIFF
--- a/smart_open/__init__.py
+++ b/smart_open/__init__.py
@@ -25,6 +25,10 @@ The main functions are:
 import logging
 from smart_open import version
 
+logger = logging.getLogger(__name__)
+if len(logger.handlers) == 0:
+    logger.addHandler(logging.NullHandler())
+
 from .smart_open_lib import open, parse_uri, smart_open, register_compressor
 from .s3 import iter_bucket as s3_iter_bucket
 
@@ -36,8 +40,4 @@ __all__ = [
     'smart_open',
 ]
 
-
 __version__ = version.__version__
-
-logger = logging.getLogger(__name__)
-logger.addHandler(logging.NullHandler())
--- a/smart_open/__init__.py
+++ b/smart_open/__init__.py
@@ -29,8 +29,8 @@ logger = logging.getLogger(__name__)
 if len(logger.handlers) == 0:
     logger.addHandler(logging.NullHandler())
 
-from .smart_open_lib import open, parse_uri, smart_open, register_compressor
-from .s3 import iter_bucket as s3_iter_bucket
+from .smart_open_lib import open, parse_uri, smart_open, register_compressor  # noqa: E402
+from .s3 import iter_bucket as s3_iter_bucket  # noqa: E402
 
 __all__ = [
     'open',
--- a/smart_open/__init__.py
+++ b/smart_open/__init__.py
@@ -29,8 +29,8 @@ logger = logging.getLogger(__name__)
 if len(logger.handlers) == 0:
     logger.addHandler(logging.NullHandler())
 
-from .smart_open_lib import open, parse_uri, smart_open, register_compressor  # noqa: E402
-from .s3 import iter_bucket as s3_iter_bucket  # noqa: E402
+from .smart_open_lib import open, parse_uri, smart_open, register_compressor
+from .s3 import iter_bucket as s3_iter_bucket
 
 __all__ = [
     'open',
--- a/smart_open/__init__.py
+++ b/smart_open/__init__.py
@@ -25,10 +25,6 @@ The main functions are:
 import logging
 from smart_open import version
 
-logger = logging.getLogger(__name__)
-if len(logger.handlers) == 0:
-    logger.addHandler(logging.NullHandler())
-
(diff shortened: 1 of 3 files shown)


piskvorky:
I tested this to wrap my head around what's happening. Recording my thoughts here:

**Before this PR:**
1. app imports lib
2. lib calls `logger.info` on import
3. logging prints the raw message to stderr
4. app configures logging with `logging.basicConfig()`
5. app calls `logger.info`
6. message is correctly formatted to log

**After this PR:**
1. app imports lib
2. lib adds NullHandler
3. lib calls `logger.info` on import
4. logging ignores the message
5. app configures logging with `logging.basicConfig()`
6. app calls `logger.info`
7. message is correctly formatted to log

I was surprised by "Before 6)". I thought "Before 4)" would be ignored because "Before 4)" already configured the logging system, but that's not the case. "Before 4)" still works.

So #475 is less severe than I thought. The application's logging was not messed up – users just saw some cryptic messages in their stderr.

piskvorky:
Btw with this PR, users will not see these `unable to import 'smart_open.gcs', disabling that module.` messages at all. Is that what we want? And if it is, why are they there in the first place? 

The global (import-time) logging is weird. Perhaps this should be a `warning`, or not import-time?

mpenkov:
@radim @menshikh-iv OK, I tried a different approach. Instead of logging the import error, we handle it silently, and instead raise an exception when the user tries to use the module that raised the import error. For example, here's what they'll see if they try to open a GCS URL without the google-cloud-core library installed:

```
$ python -c 'import smart_open;smart_open.open("gs://foo/bar")'
  File "/home/misha/git/smart_open/smart_open/transport.py", line 74, in get_transport
    submodule = _REGISTRY[scheme]
KeyError: 'gs'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/home/misha/git/smart_open/smart_open/smart_open_lib.py", line 224, in open
    binary = _open_binary_stream(uri, binary_mode, transport_params)
  File "/home/misha/git/smart_open/smart_open/smart_open_lib.py", line 398, in _open_binary_stream
    submodule = transport.get_transport(scheme)
  File "/home/misha/git/smart_open/smart_open/transport.py", line 76, in get_transport
    raise NotImplementedError(message)
NotImplementedError: Unable to handle scheme 'gs', expected one of ('', 'file', 'hdfs', 'http', 'https', 's3', 's3a', 's3n', 's3u', 'scp', 'sftp', 'ssh', 'webhdfs'). Extra dependencies required by 'gs' may be missing. See <https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst> for details.
```

radim:
Wrong Radim 😉

mpenkov:
I mean @piskvorky, whoops

smart_open/__init__.py
```diff
@@ -40,4 +36,8 @@
     'smart_open',
 ]
 
+
 __version__ = version.__version__
+
+logger = logging.getLogger(__name__)
```
piskvorky: Why this revert? Isn't it safer to add the handler early on, to avoid issues like this in the future?

mpenkov: I reverted it because it was no longer a necessary part of the solution.

If you think it's worth keeping for future-proofing the existing solution, then I can add it back in.

piskvorky: Not critical, but it makes sense to add the handler as soon as possible. Who knows what the contributed modules will do in the future (a missed logging call on import…).

Issue #491, opened 2020-04-27
Investigate building wheels for smart_open


mpenkov:
This has certain benefits:

1. The pip client no longer has to build wheels itself when installing
2. The install process is marginally faster
3. Any others?

Also, are there any compelling reasons to avoid building wheels?

@menshikh-iv @piskvorky @gojomo

piskvorky:
Increased complexity. Smart_open doesn't compile anything, why would we need wheels?

menshikh-iv:
The wheel doesn't contain any "binary" stuff for the pure-python project.
Wheel building has no difference with default tarball if the project contains ONLY python files. 

You just do `python setup.py bdist_wheel` instead of `python setup.py sdist`, where is complexity? Also, you need an only a single wheel (same as tarball) for all stuff (instead of compiling separate wheel for each python X each OS version as in gensim). See https://packaging.python.org/guides/distributing-packages-using-setuptools/#pure-python-wheels for more information.

menshikh-iv:
So, here is all needed changes https://github.com/RaRe-Technologies/smart_open/pull/492

piskvorky:
> Wheel building has no difference with default tarball if the project contains ONLY python files.

Exactly – I see no point of complicated things with wheels.

menshikh-iv:
I don't see any "complication" here, can you point to exact "complicated things", please?

piskvorky:
Extra features / files = extra complications, extra maintenance. It's common sense.

To summarize:

Pros:
1. The pip client no longer has to build wheels itself when installing. (*that's an observation; what's the actual benefit?*)
2. The install process is marginally faster

Cons:
1. Extra maintenance.

Both Pros and Cons seem quite inconsequential to me, minor. A low priority ticket.

gojomo:
I'd prefer not using a wheel unless necessary for some user convenience, as my understanding is:

(1) it makes the download a little larger;
(2) in includes more opaque things (in this case `.pyc`) that are fully determined by the actual source. 

Aesthetically and as a matter of "minimal surface area for errors/confusion", I'd prefer delivering the tiniest, most-transparent set of things as the package – given that they can become the full package on demand at any users' environment.

menshikh-iv:
> (1) it makes the download a little larger;

why? I see the opposite on practice

``` 
ivan@P50:~/release/smart_open$ python setup.py sdist bdist_wheel
ivan@P50:~/release/smart_open$ cd dist/
[...]

PR #492, opened 2020-04-27
[EXAMPLE] Build pure python wheel

menshikh-iv:
Demonstration for #491

### DIFF
--- a/release/push_pypi.sh
+++ b/release/push_pypi.sh
@@ -22,7 +22,7 @@ set -u
 
 cd ..
 pip install twine
-python setup.py sdist
+python setup.py sdist bdist_wheel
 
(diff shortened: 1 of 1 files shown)


sdolenc:
FWIW: I tried creating a whl file for smart_open. For some reason `python setup.py bdist_wheel` was generating a tiny (unusuable) .whl file for me, but `pip wheel .` from within the cloned repo did the trick. It may have been a fluke, but I thought I'd share just in case

Hopefully smart_open gets official wheels soon

mpenkov:
I did this https://github.com/RaRe-Technologies/smart_open/commit/e42a7c7df49fc5cbd8022666876e3a555ff913d0 here, thanks for the idea.

PR #495, opened 2020-05-07
Avoid redundant GetObject API queries in s3.Reader

### LINKED ISSUE #494: Redundant GetObject requests when fetching file contents from S3
#### Problem description

When fetching the contents of a file from S3, smart_open sends two GetObject API requests -- one to fetch the file length, and one to fetch the contents. From an EC2 instance in the same region as the bucket, this doubles the cost of fetching the file.

The double request happens because there is one call to `_get` in `Reader.__init__`, solely to fetch the content length, and a second call to `_get` in `_SeekableRawReader._load_body`.

Here is a demo of the problem, enabling logging in botocore to see the requests sent:

```
In [1]: from smart_open import open
In [2]: import boto3
In [3]: import logging
   ...: boto3.set_stream_logger('botocore.endpoint', logging.DEBUG)
In [4]: with open('s3://my-bucket/my-object') as f:
   ...:     contents = f.read()
   ...:
2020-05-05 15:50:34,840 botocore.endpoint [DEBUG] Making request for OperationModel(name=GetObject) with params: ...{'url_path': '/my-bucket/my-object' 
...
2020-05-05 15:50:35,020 botocore.endpoint [DEBUG] Making request for OperationModel(name=GetObject) with params: {'url_path': '/my-bucket/my-object'
...
```

Performing the same load directly with boto3 triggers only a single API request:

```
In [6]: contents = boto3.resource('s3').Object('my-bucket', 'my-object').get()['Body'].read()
2020-05-05 15:54:28,232 botocore.endpoint [DEBUG] Making request for OperationModel(name=GetObject) with params: {'url_path': ...
```

#### Versions

```
Linux-4.14.173-137.229.amzn2.x86_64-x86_64-with-glibc2.2.5
Python 3.7.6 (default, Feb 26 2020, 20:54:15)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-6)]
smart_open 2.0.0
```

#### Checklist

jcushman:
#### Motivation

Fixes #494

As documented in #494, simple uses of `open` with an S3 resource result in two redundant (and billable) GetObject calls instead of one. For example, this uses two API calls:

```open("s3://my-bucket/my-object").read()```

The first query is used solely to populate `Reader._content_length`.

This PR avoids the redundant queries by making `Reader._content_length` a `@property` provided by `Reader._raw_reader._content_length`, which is in turn populated by `_raw_reader._load_body()`.

There are two user-visible changes:

* `open("s3://my-bucket/my-object")._content_length` will return `None` until `.seek()` or `.read()` is called.
* `open("s3://my-bucket/nonexistent-object")` will not throw `KeyNotFound` until `.seek()` or `.read()` is called.

Both changes could be avoided by calling `seek(0)` inside `__init__`, which would still avoid redundant API queries in most cases. However, that would result in a redundant API query if the file is opened and seek is immediately called by the user.

This PR also includes a change to the s3 unit tests so that tests can assert what API queries were called. I only added those assertions for `SeekableBufferedInputBaseTest`, but it would probably make sense to add them to the other tests in `test_s3.py` as well.

#### Checklist

### CODE REVIEW

File: smart_open/s3.py [hunk 1]
```diff
@@ -279,47 +279,92 @@ class _SeekableRawReader(object):
...
+        self._content_length = None
         self._version_id = version_id
         self._position = 0
         self._body = None
         self._object_kwargs = object_kwargs if object_kwargs else {}
 
-    def seek(self, position):
-        """Seek to the specified position (byte offset) in the S3 key.
+    def seek(self, offset, whence=constants.WHENCE_START):
+        """Seek to the specified position.
```
jcushman, line 291: The `whence` logic and `range_string` logic both have to move into this seek() so we can handle the case of reading bytes from the end of a file:

```
with open("s3://my-bucket/my-object") as f:
    f.seek(-1000, constants.WHENCE_END)
    f.read()
```

The above only requires one `GetObject` if we call it with `Range: bytes=-1000`.

smart_open/s3.py
```diff
@@ -279,47 +279,92 @@ class _SeekableRawReader(object):
...
+            # If we didn't know _content_length yet in .seek(), we may have requested an
+            # invalid range. In that case, set _position and _content_length from the actual
+            # object size returned by S3.
+            orig_error = error.__context__
+            aws_error = orig_error and hasattr(orig_error, 'response') and orig_error.response.get('Error', {})
+            if aws_error and aws_error.get('Message') == 'Requested Range Not Satisfiable':
+                if 'ActualObjectSize' in aws_error:
+                    self._position = self._content_length = int(aws_error['ActualObjectSize'])
+                else:
+                    # this shouldn't happen with real S3, but does happen with moto which seems to lack ActualObjectSize
```
jcushman, line 353: This `else` is a little odd -- real AWS returns `'ActualObjectSize'` in the response, so we only need one API call for seeking past the end of the file:

```
with open("s3://my-bucket/my-object") as f:
    f.seek(1_000_000_000)
```

`moto` doesn't return `'ActualObjectSize'`, so we have to go back and fetch it, which makes the tests a little different from real S3.

jcushman:
Additional API thoughts:

`_content_length` could raise an `AttributeError` if it is accessed before length is available, which might avoid some subtle bugs when its value is None.

`open()` could have new `seek=0, seek_whence=constants.WHENCE_START` parameters, and then `__init__` could do

```
    if seek is not None:
        self.seek(seek, seek_whence)
```

This way the current behavior is preserved by default, without the redundant API calls, but you can also efficiently handle a seek-before-read. Probably it would be a good idea to implement that for all backends if implementing it at all, though.

smart_open/s3.py
```diff
@@ -279,47 +279,98 @@ class _SeekableRawReader(object):
...
+                orig_error
+                and hasattr(orig_error, 'response')
+                and orig_error.response.get('Error', {})
+            )
+            if aws_error and aws_error.get('Message') == 'Requested Range Not Satisfiable':
+                if 'ActualObjectSize' in aws_error:
+                    self._position = self._content_length = int(aws_error['ActualObjectSize'])
+                else:
+                    # this shouldn't happen with real S3,
+                    # but does happen with moto which seems to lack ActualObjectSize
```
mpenkov, line 359: Are you sure this is the case?

If yes, then please create an issue in the moto repository, describe the problem there, and link to the new issue in your code. We can deal with moto's quirks, but we don't want them in our code indefinitely, so if we include this in our code base, we want to connect it to a future resolution.

smart_open/s3.py
```diff
@@ -279,47 +279,98 @@ class _SeekableRawReader(object):
...
+        try:
+            # Optimistically try to fetch the requested content range.
+            response = _get(
                 self._object,
                 version=self._version_id,
                 Range=range_string,
                 **self._object_kwargs
-            )['Body']
+            )
+        except IOError as error:
```
mpenkov, line 344: This exception-handling block is a bit too busy. I think it's worth simplifying it. Make it obvious what you're doing and why.

To me, it looks like there are two kinds of IOError:

1. An "expected" one. In this case, we indicate that we have encountered the EOF.
2. An "unexpected" one. In this case, we re-raise.

So, I would do something like:

```python
_OUT_OF_RANGE = 'Requested Range Not Satisfiable'
"""Returned by AWS when we try to seek beyond EOF."""

def _unwrap_ioerror(ioe):
    if ioe.__context__ and hasattr(ioe.__context__, 'response'):
        inner_error = ioe.__context__.response.get('Error')
        if inner_error.get('Message') == _OUT_OF_RANGE:
            return inner_error

try:
    ...
except IOError as ioe:
    inner_error = _unwrap_ioerror(ioe)
    if not inner_error:
        raise

    try:
        self._position = self._content_length = inner_error['ActualObjectSize']
    except KeyError:
        self._position = self._content_length = ... # handle moto quirk here
else:
    ...
```

smart_open/s3.py
```diff
@@ -279,47 +279,98 @@ class _SeekableRawReader(object):
...
+            # If we didn't know _content_length yet in .seek(), we may have requested an
+            # invalid range. In that case, set _position and _content_length from the actual
+            # object size returned by S3.
+            orig_error = error.__context__
```
mpenkov, line 348: Can we avoid accessing a dunder here? It's a bit fragile.

smart_open/s3.py
```diff
@@ -279,47 +279,98 @@ class _SeekableRawReader(object):
...
+                    self._position = self._content_length = _get(
+                        self._object,
+                        version=self._version_id,
+                        **self._object_kwargs
+                    )['ContentLength']
+                self._body = io.BytesIO()
+                return
+            raise
+        else:
+            # Set _position and _content_length from the content-range header returned by S3.
```
mpenkov, line 369: This comment adds little value to the source, because it describes _what_ the source does, and that is already obvious from the source itself. Please remove.

smart_open/s3.py
```diff
@@ -527,6 +565,10 @@ def to_boto3(self):
     #
     # Internal methods.
     #
+    @property
```
mpenkov, line 568: Do we actually need this property?

It's internal, so not subject to any contracts. Our internal code can use `self._raw_reader._content_length` directly, right?

smart_open/s3.py
```diff
@@ -478,18 +525,9 @@ def seek(self, offset, whence=constants.WHENCE_START):
...
-        if whence == constants.WHENCE_START:
-            new_position = offset
-        elif whence == constants.WHENCE_CURRENT:
-            new_position = self._current_pos + offset
-        else:
-            new_position = self._content_length + offset
-        new_position = smart_open.utils.clamp(new_position, 0, self._content_length)
-        self._current_pos = new_position
-        self._raw_reader.seek(new_position)
+        self._raw_reader._position = self._current_pos
```
mpenkov, line 529: Why are you tweaking an internal attribute of the raw reader here prior to seeking? It seems like you're breaking encapsulation.

smart_open/s3.py
```diff
@@ -412,8 +457,10 @@ def read(self, size=-1):
             return b''
         elif size < 0:
             from_buf = self._read_from_buffer()
+            # call read() before setting _current_pos to make sure _content_length is set
+            out = from_buf + self._raw_reader.read()
```
mpenkov, line 461: Why bother defining `from_buf` above, when you can do this:

```suggestion
            out = self._read_from_buffer() + self._raw_reader.read()
```

smart_open/s3.py
```diff
@@ -330,11 +381,11 @@ def _read_from_body(self, size=-1):
 
     def read(self, size=-1):
         """Read from the continuous connection with the remote peer."""
-        if self._position >= self._content_length:
-            return b''
         if self._body is None:
-            # When the first read() after __init__() or seek(), self._body is not exist.
+            # In the first read() after __init__(), self._body does not exist.
```
mpenkov, line 385: Well, technically it does exist: it is explicitly defined as `None`.

```suggestion
            # This is necessary for the very first read() after __init__().
```

smart_open/s3.py
```diff
@@ -279,47 +279,98 @@ class _SeekableRawReader(object):
...
+            self._position = self._content_length
         else:
-            self._body = _get(
+            range_string = smart_open.utils.make_range_string(start, stop)
+            self._load_body(range_string)  # sets self._position
+
+        return self._position
+
+    def _load_body(self, range_string=None):
+        """Build a continuous connection with the remote peer starts from the current postion.
```
mpenkov, line 330: This docstring doesn't really match the name of the method.

If it were me, I would write something like:

> Reads the specified range of bytes over the wire. If no range is specified, reads the entire content. Stores the result in `_body`. Updates `_position` and `_content_length` as a side effect.

smart_open/s3.py
jcushman: Filed issue and linked.

smart_open/s3.py
jcushman: I made the helper function a little more general -- it returns the `'Error'` dict from the s3 response, which seems like it might be useful elsewhere at some point.

smart_open/s3.py
jcushman: Attached the wrapped error as `backend_error`, a custom attribute to the IOError. I can see that being generally useful at some point.

smart_open/s3.py
jcushman: The issue is that _current_pos and _raw_reader._position don't always have the same value because of buffering, so `WHENCE_CURRENT` doesn't work. Latest code uses an alternate approach to fix that:

```
        if whence == constants.WHENCE_CURRENT:
            whence = constants.WHENCE_START
            offset += self._current_pos
```

 Not sure that this is cleaner, but 🤷‍♂️.

smart_open/s3.py
jcushman: True that the name and docstring don't match, but the docstring is more accurate -- it opens a connection but doesn't necessarily grab all the data yet. (Imagine calling `open(path).read(10)` on a gigabyte-large file.) I renamed the method to `_open_body` and expanded the docstring.

jcushman:
@mpenkov, this is ready for review.

There's also still an outstanding design question from my previous comment about whether to add a `seek(0)` to `Reader.__init__`, optionally controlled by new parameters to `open()`.

smart_open/utils.py
```diff
@@ -94,13 +94,13 @@ def clamp(value, minval, maxval):
...
-def make_range_string(start, stop=None):
+def make_range_string(start=None, stop=None):
     """Create a byte range specifier in accordance with RFC-2616.
 
     Parameters
     ----------
-    start: int
-        The start of the byte range
+    start: int, optional
+        The start of the byte range.  If unspecified, stop indicated offset from EOF.
```
mpenkov, line 103: Can you point to the part of the [RFC](https://tools.ietf.org/html/rfc2616) that describes this?

I couldn't find anything about this use case (start is None, stop is non-None).

```
   If the last-byte-pos value is present, it MUST be greater than or
   equal to the first-byte-pos in that byte-range-spec, or the byte-
   range-spec is syntactically invalid. The recipient of a byte-range-
   set that includes one or more syntactically invalid byte-range-spec
   values MUST ignore the header field that includes that byte-range-
   set.
```

jcushman: Sure, it's on the next page:

```
       suffix-byte-range-spec = "-" suffix-length
       suffix-length = 1*DIGIT

   A suffix-byte-range-spec is used to specify the suffix of the
   entity-body, of a length given by the suffix-length value. (That is,
   this form specifies the last N bytes of an entity-body.) If the
   entity is shorter than the specified suffix-length, the entire
   entity-body is used.
```

FWIW this behavior is confirmed by `test_seek_end`.

smart_open/tests/test_s3_version.py
```diff
@@ -91,7 +91,7 @@ def test_bad_id(self):
         """Does passing an invalid version_id exception into the s3 submodule get handled correctly?"""
         params = {'version_id': 'bad-version-does-not-exist'}
         with self.assertRaises(IOError):
-            open(self.url, 'rb', transport_params=params)
+            open(self.url, 'rb', transport_params=params).read()
```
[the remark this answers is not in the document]
jcushman: See my original message with the PR, and also [this comment](https://github.com/RaRe-Technologies/smart_open/pull/495#issuecomment-625804384). If we want to raise on open, we simply add a `this.seek(0)` call to `__init__`. But then the seek-before-read usecase will unavoidably result in redundant API calls:

```
with open('s3://bucket/key') as f:
    f.seek(1000000)
    f.read()
```

The redundant API calls could be avoided (by careful users) if we added `seek` and `seek_whence` parameters to `open()`, so a user could do:

```
with open('s3://bucket/key', seek=1000000) as f:
    f.read()
```

or

```
with open('s3://bucket/key', seek=None) as f:
    f.seek(1000000)
    f.read()
```

which would be handled in `__init__` by:

```
    if seek is not None:
        self.seek(seek, seek_whence)
```

I'm happy to do any of these, just looking for a decision on the design question.

smart_open/tests/test_s3.py
```diff
@@ -124,34 +143,41 @@ def test_iter(self):
...
         fin = smart_open.s3.SeekableBufferedInputBase(BUCKET_NAME, KEY_NAME)
         output = [line.rstrip(b'\n') for line in fin]
         self.assertEqual(output, expected.split(b'\n'))
+        self.assertDictEqual(self.api_calls, {'GetObject': 1})
 
     def test_iter_context_manager(self):
         # same thing but using a context manager
         expected = u"hello wořld\nhow are you?".encode('utf8')
         put_to_bucket(contents=expected)
+        self.api_calls.clear()
```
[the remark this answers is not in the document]
jcushman: Yes, it's there so test-case-specific setup calls don't pollute the count. I don't think it belongs in either `put_to_bucket` or `setUp`, though. Note that it sometimes comes after other test-specific setup calls, e.g. in `test_read_gzip`.

If you prefer, it might be a natural fit for a context manager, e.g.:

```
    def test_iter_context_manager(self):
        # same thing but using a context manager
        expected = u"hello wořld\nhow are you?".encode('utf8')
        put_to_bucket(contents=expected)
        with self.assertApiCalls(GetObject=1):
            with smart_open.s3.SeekableBufferedInputBase(BUCKET_NAME, KEY_NAME) as fin:
                output = [line.rstrip(b'\n') for line in fin]
                self.assertEqual(output, expected.split(b'\n'))
```

That's how it's handled in [django_assert_num_queries](https://pytest-django.readthedocs.io/en/latest/helpers.html#django-assert-num-queries), which is a similar test helper.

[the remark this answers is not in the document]
jcushman: I refactored to use the `with` approach.

smart_open/utils.py
jcushman: I added a `seek(0)` call to __init__, and a `defer_seek` parameter that allows that call to be skipped.

jcushman:
@mpenkov I believe this is ready for review. This 3rd version includes a proposed API where `seek(0)` is called by default during `open()`, but can be skipped with `defer_seek=True`. This means there is no user-visible change from the current behavior.

mpenkov:
Can you please fix the flake8 issue?

jcushman:
I fixed the flake8 issue specific to this PR, but left the flake8 issue that is [already failing on the develop branch](https://travis-ci.org/github/RaRe-Technologies/smart_open/jobs/687984676), as that's arising from a file not implicated by this PR.

petedannemann:
@jcushman can you rebase off of develop? We fixed the flake8 issue in another PR and merged it

mpenkov:
Thank you for your contribution and your patience @jcushman ! And congrats on your first PR to smart_open 🥇 !

PR #515, opened 2020-07-01
Lint README.rst on each commit

mpenkov:


mpenkov:
We already do this, but slightly differently now: https://github.com/RaRe-Technologies/smart_open/blob/develop/tox_helpers/doctest.py

Issue #468, opened 2020-04-08
Cannot use `smart_open.s3.iter_bucket` because of missing `boto` dependancy.


roamdam:
#### Problem description

It seems that `1.11` causes import problems with `boto`. `boto` does not seem to appear as a dependancy for the package anymore, while it is still imported in `smart_open.s3`. That does not throw an error when installing but does when I try to use `smart_open.s3.iter_bucket`.

#### Steps/code to reproduce the problem

I've made the following to be sure, on a brand new virtual environment

```bash
mkdir test_repo && cd test_repo
virtualenv venv
pip3 install smart-open
python3 test-import.py
```
With `test-import.py` being : 

```python
from smart_open.s3 import iter_bucket

iter_bucket("some bucket")
```
Here is the result from `python3 test-import.py` : 

```python
$ python3 test-import.py 
unable to import 'smart_open.gcs', disabling that module
unable to import 'smart_open.s3', disabling that module
Traceback (most recent call last):
  File "test-import.py", line 1, in <module>
    from smart_open.s3 import iter_bucket
  File "/Users/**/test-repo/venv/lib/python3.6/site-packages/smart_open/__init__.py", line 29, in <module>
    from .s3 import iter_bucket as s3_iter_bucket
  File "/Users/**/test-repo/venv/lib/python3.6/site-packages/smart_open/s3.py", line 15, in <module>
    import boto
ModuleNotFoundError: No module named 'boto'
```

#### Versions

Please provide the output of:

```python
import platform, sys, smart_open
print(platform.platform())
print("Python", sys.version)
print("smart_open", smart_open.__version__)
```
The above import of smart_open does not work either, but I'm using version `1.11.0`

```
Darwin-18.7.0-x86_64-i386-64bit
Python 3.6.7 (v3.6.7:6ec5cf24b7, Oct 20 2018, 03:02:14) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]
smart_open 1.11.0
```

#### Checklist

Before you create the issue, please make sure you have:

- [x] Described the problem clearly
- [x] Provided a minimal reproducible example, including any required data
- [x] Provided the version numbers of the relevant software

mpenkov:
Thank you for letting us know. We've released version 1.11.1 to deal with this problem. Please upgrade.

piskvorky:
@mpenkov how come the testing system did not catch this? Did the recent refactoring break the CI?

[...]

Issue #409, opened 2020-01-12
Remove boto entirely


mpenkov:
We don't really need `boto`, IMHO. It's entirely separate to `boto3`.

Let's remove it in a separate PR.

_Originally posted by @mpenkov in https://github.com/RaRe-Technologies/smart_open/pull/408/review_comment/create_

Issue #391, opened 2019-11-25
Is that posiible to remove boto dependency?


NumberVII:
Boto is quite large and no longer provided in AWS lambda env. Any plan to remove that?

piskvorky:
Possibly. Can you prepare a PR?

PR #523, opened 2020-07-20
Remove boto as a dependency

isobit:
Fixes:
- https://github.com/RaRe-Technologies/smart_open/issues/391
- https://github.com/RaRe-Technologies/smart_open/issues/409

Related to:
- https://github.com/RaRe-Technologies/smart_open/pull/413
- https://github.com/RaRe-Technologies/smart_open/issues/468
- https://github.com/RaRe-Technologies/smart_open/pull/408


setup.py
```diff
@@ -41,6 +41,7 @@ def read(fname):
     'moto[server]',
     'pathlib2',
     'responses',
+    'boto',
```
mpenkov, line 44: Why do our tests still need boto?

isobit: There's a `test_old` method which uses an old `boto` bucket: https://github.com/RaRe-Technologies/smart_open/blob/ecce29ebb3b0dee519f6939e924ee4bb5974ca33/smart_open/tests/test_s3.py#L576

I could at least leave a comment here with an explanation. I suppose alternatively the support for passing in old boto buckets could be deprecated and the test removed.

mpenkov: Yeah, I think it's best to remove it altogether.

Sorry for the late response.

README.rst
```diff
@@ -29,9 +29,9 @@ What?
 Why?
 ====
 
-Working with large remote files, for example using Amazon's `boto <http://docs.pythonboto.org/en/latest/>`_ and `boto3 <https://boto3.readthedocs.io/en/latest/>`_ Python library, is a pain.
-``boto``'s ``key.set_contents_from_string()`` and ``key.get_contents_as_string()`` methods only work for small files, because they're loaded fully into RAM, no streaming.
-There are nasty hidden gotchas when using ``boto``'s multipart upload functionality that is needed for large files, and a lot of boilerplate.
+Working with large remote files, for example using Amazon's `boto3 <https://boto3.readthedocs.io/en/latest/>`_ Python library, is a pain.
+``boto3``'s ``key.set_contents_from_string()`` and ``key.get_contents_as_string()`` methods only work for small files, because they're loaded fully into RAM, no streaming.
```
piskvorky, line 33: Does boto3 even have `set_contents_from_string` / `get_contents_as_string `?

isobit: Ah good point, I don't think it does. `boto3` has `upload_fileobj()` and `download_fileobj()` which do handle multipart streaming. The point about having a unified pythonic API still stands though obviously.

mpenkov:
Thank you for your contribution and your patience @isobit ! And congrats on your first PR to smart_open 🥇

PR #534, opened 2020-08-24
Support installing smart_open without AWS dependencies

### LINKED ISSUE #463: Make the submodule importing system robust to missing extras
I'm not sure about this but I think the imports in the code should now be in try except ? Like if someone installs the smart_open[gcs] the smart_open.s3 will throw an import error

_Originally posted by @Tyrannas in https://github.com/RaRe-Technologies/smart_open/pull/454#issuecomment-609804100_

### LINKED ISSUE #533: Installing without AWS dependencies?
First, thanks for building the awesome library. It works great! 👏  

I've built a [library](https://github.com/justindujardin/pathy) that subclasses pathlib.Path to work with bucket storage, and it uses smart_open to provide the streamable io interface.

When installing smart_open it pulls in the AWS dependencies even if I'm only using it for GCS. I see that you all have setup the package to allow installing only the dependencies for each service, but it seems incomplete?

https://github.com/RaRe-Technologies/smart_open/blob/b6cf769266b34ad2146e9ee67cc56b331ab8e91b/setup.py#L88-L91

I found this comment about the extra dependencies being temporary, and I wonder if you have a timeline for a release that removes this?

justindujardin:
#### Title

Support installing smart_open without AWS dependencies

#### Motivation

Using smart_open with GCS should not require AWS packages to be installed.

- Fixes #533 
- Fixes #463 

#### Changes

- [x] show friendly errors when using `smart_open.open` with an unsupported scheme because of missing dependencies
- [x] show deprecation warning once when using s3_iter_bucket from the top-level module
- [x] update tox.ini to run dependency tests without any extras packages installed
- [x] don't install AWS deps as part of the base package

#### Checklist


justindujardin:
@mpenkov @piskvorky could you provide review/feedback on this?

justindujardin:
I added code to throw the helpful import errors when you call `smart_open.open("scheme://bar")` with a scheme that is unsupported because it's missing imports. The code isn't quite as nice as I'd like it to be (I had to duplicate the schemes outside of the transport file) but it yields a nice result when you use the open function:

```bash
(.env) smart_open > python
>>> open("gs://foo")
  File "smart_open/smart_open/smart_open_lib.py", line 223, in open
    binary = _open_binary_stream(uri, binary_mode, transport_params)
  File "smart_open/smart_open/smart_open_lib.py", line 401, in _open_binary_stream
    submodule = transport.get_transport(scheme)
  File "smart_open/smart_open/transport.py", line 86, in get_transport
    raise ImportError(_ERRORS[scheme])
ImportError: You are trying to use the GCS functionality of smart_open
but you do not have the correct GCS dependencies installed. Try:

    pip install smart_open[gcs]


>>> open("s3://foo")
  File "smart_open/smart_open/smart_open_lib.py", line 223, in open
    binary = _open_binary_stream(uri, binary_mode, transport_params)
  File "smart_open/smart_open/smart_open_lib.py", line 401, in _open_binary_stream
    submodule = transport.get_transport(scheme)
  File "smart_open/smart_open/transport.py", line 86, in get_transport
    raise ImportError(_ERRORS[scheme])
ImportError: You are trying to use the S3 functionality of smart_open
but you do not have the correct AWS dependencies installed. Try:

    pip install smart_open[aws]


>>> open("azure://foo")
  File "smart_open/smart_open/smart_open_lib.py", line 223, in open
    binary = _open_binary_stream(uri, binary_mode, transport_params)
  File "smart_open/smart_open/smart_open_lib.py", line 401, in _open_binary_stream
    submodule = transport.get_transport(scheme)
  File "smart_open/smart_open/transport.py", line 86, in get_transport
    raise ImportError(_ERRORS[scheme])
ImportError: You are trying to use the Azure functionality of smart_open
but you do not have the correct Azure dependencies installed. Try:

    pip install smart_open[azure]

```

setup.py
```diff
@@ -42,6 +42,7 @@ def read(fname):
     'pathlib2',
     'responses',
     'boto3',
+    'python-jose',
```
mpenkov, line 45: What is this and why do we need it?

smart_open/__init__.py
```diff
@@ -27,11 +27,44 @@
...
 #
-logging.getLogger(__name__).addHandler(logging.NullHandler())
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
 
 from smart_open import version  # noqa: E402
 from .smart_open_lib import open, parse_uri, smart_open, register_compressor  # noqa: E402
-from .s3 import iter_bucket as s3_iter_bucket  # noqa: E402
+
+deprecated_warning = """smart_open.s3_iter_bucket is deprecated and will stop functioning
```
mpenkov, line 36: This looks like an internal constant, so it's better to explicitly name it like one.

```suggestion
_WARNING = """smart_open.s3_iter_bucket is deprecated and will stop functioning
```

smart_open/__init__.py
```diff
@@ -27,11 +27,44 @@
...
+in a future version. Please import iter_bucket from the smart_open.s3 module instead:
+
+    from smart_open.s3 import iter_bucket as s3_iter_bucket
+
+"""
+_warned = False
+
+
+def s3_iter_bucket(
+    bucket_name,
```
mpenkov, line 46: Please indent the parameter list more, e.g.

```suggestion
        bucket_name,
```

to separate it from the function body visually.

smart_open/__init__.py
```diff
@@ -27,11 +27,44 @@
...
 from .smart_open_lib import open, parse_uri, smart_open, register_compressor  # noqa: E402
-from .s3 import iter_bucket as s3_iter_bucket  # noqa: E402
+
+deprecated_warning = """smart_open.s3_iter_bucket is deprecated and will stop functioning
+in a future version. Please import iter_bucket from the smart_open.s3 module instead:
+
+    from smart_open.s3 import iter_bucket as s3_iter_bucket
+
+"""
+_warned = False
```
mpenkov, line 42:
```suggestion
_WARNED = False
```

smart_open/azure.py
```diff
@@ -15,8 +15,17 @@
 import smart_open.bytebuffer
 import smart_open.constants
 
-import azure.storage.blob
-import azure.core.exceptions
+import_error = """You are trying to use the Azure functionality of smart_open
```
mpenkov, line 18: I don't like the idea of duplicating this across all of the modules.

The pattern is predictable:

- `azure.py` -> `pip install smart_open[azure]`
- `gcs.py` -> `pip install smart_open[gcs]`
- `s3.py` -> `pip install smart_open[aws]` (this one needs to change)
- ... and so on

So all you need to do is provide an informative message when handling the ImportError in transport.py.  Something like:

```python
try:
    ... # code that tries to import backend
except ImportError as error:
    logger.exception(error)
    logger.info('try `pip install smart_open[%s]` to install the missing dependencies' % submodule)
```

For `s3.py`, we should either rename the module to `aws.py` or rename the extras to `s3` to make things more predictable. I think the latter is easier than the former.

smart_open/transport.py
```diff
@@ -63,29 +72,29 @@ def get_transport(scheme):
...
-register_transport('smart_open.ssh')
-register_transport('smart_open.webhdfs')
+    if scheme in _REGISTRY:
+        return _REGISTRY[scheme]
+    if scheme in _ERRORS:
+        raise ImportError(_ERRORS[scheme])
+    raise NotImplementedError(message)
+
+
+register_transport(smart_open.local_file, ["file"])
```
mpenkov, line 90: Why is it necessary to specify the second parameter here? It sounds like the sort of thing the submodules should know how to handle themselves.

mpenkov (changes requested):
Thank you for your patience. Left you a review, please have a look.

smart_open/azure.py
justindujardin: Sure, I can do that. One thing that will change by moving this error handling to transport is that you won't get nice errors if you're using smart_open's service-specific modules from python code, e.g.

```python
from smart_open import gcs
```

As-is that code would show the suggested pip install error, but if moved it will revert to the normal GCS import error. Is that acceptable?

smart_open/transport.py
justindujardin: Yeah, that would normally be the case if you imported the submodule without error. But consider if you tried to open a blob (`gs://foo/bar`) without the GCS libraries installed. When the import error is raised from `gcs.py`, the `SCHEME/S` cannot be referenced in `transport.py`.

So without specifying the schemes here, when the open function is called for a GCS blob without the GCS libraries installed, the gs scheme isn't in the known list and transport has to show the generic "please look at this readme file" error.

setup.py
justindujardin: pip threw some error about resolving dependencies that when I ran the test suite on MacOS Catalina. I'll try removing it and report back with the specific error.

justindujardin: When running the tests locally on MacOS Catalina, I get this error:

```
ImportError while importing test module 'smart_open/tests/test_smart_open.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
smart_open/tests/test_smart_open.py:21: in <module>
    from moto import mock_s3
.env/lib/python3.7/site-packages/moto/__init__.py:14: in <module>
    from .cognitoidp import mock_cognitoidp, mock_cognitoidp_deprecated  # noqa
.env/lib/python3.7/site-packages/moto/cognitoidp/__init__.py:2: in <module>
    from .models import cognitoidp_backends
.env/lib/python3.7/site-packages/moto/cognitoidp/models.py:13: in <module>
    from jose import jws
E   ModuleNotFoundError: No module named 'jose'
```

installing python-jose resolves it. The error looks like it tracks back to the `mock_s3` package

smart_open/azure.py
justindujardin: I refactored the code to work as you suggest

smart_open/transport.py
justindujardin: I removed this code by handling the ImportError in the various transports and setting a module attribute `MISSING_DEPS=True`. Transport is then able to inspect the schemes even if they have missing packages.

smart_open/__init__.py
```diff
@@ -33,17 +33,17 @@
...
     from smart_open.s3 import iter_bucket as s3_iter_bucket
 
 """
-_warned = False
+_WARNED = False
 
 
 def s3_iter_bucket(
-    bucket_name,
+        bucket_name,
```
mpenkov, line 46: Please do this for all the other parameters as well. Sorry, I thought it was obvious.

justindujardin: Were you viewing an old diff maybe? They're all indented the same as far as I can see

piskvorky: Please put the final `):` on a separate line. It helps readability.

justindujardin: I'm happy to put the code in a style you prefer. Have y'all considered using something like Black to auto-format your code? It would certainly cut down on this kind of back and forth, and you could even configure a check to fail a PR so the style matches what you want before you even leave a review. 

I'd be happy to set that up in a separate PR if you're interested. For now I've matched the style that I found for other multi-line functions in the codebase. If you want it different, perhaps suggest an edit?

piskvorky: Yes, we did consider black. Just didn't get to it yet :)

setup.py
mpenkov: I think that's a problem with `moto`. Please open a ticket with them and let them know they're missing a dependency.

We shouldn't have to deal with it in our code.

smart_open/__init__.py
mpenkov: Yeah, I still have a strong preference against automatic code formatting.

smart_open/transport.py
```diff
@@ -54,7 +66,10 @@ def register_transport(submodule):
 
     for scheme in schemes:
         assert scheme not in _REGISTRY
-        _REGISTRY[scheme] = submodule
+        if hasattr(submodule, "MISSING_DEPS") and getattr(submodule, "MISSING_DEPS") is True:
```
mpenkov, line 69: Is this not sufficient?

```suggestion
        if getattr(submodule, "MISSING_DEPS") is True:
```

If there is no such attribute, getattr should return the default (None), right?

justindujardin: It's not, because `getattr` raises `AttributeError` when you use it like that. However, I looked closer at the API after your comment and it turns out if you explicitly specify the default, it does not raise. Thanks for that tip!

setup.py
justindujardin: Removed it

smart_open/__init__.py
justindujardin: @mpenkov It's interesting that you have a strong preference against it. I looked at some of your PRs and you're clearly a talented engineer. Why would you want to spend your valuable time nitpicking my code style in addition to the substance? When you auto-format code you don't always get the _exact_ style you want, but with OSS where consistency is hard to maintain, it eliminates an entire class of thing that you have to code-review for.

justindujardin:
Thanks for the thorough review! I think that I've addressed all your comments now. Let me know if you want any more changes.

smart_open/__init__.py
mpenkov: I've addressed this [elsewhere](https://github.com/RaRe-Technologies/gensim/pull/2712#discussion_r364076359).

Under certain project or team dynamics, I can see how something like Black may be worth it. For example, if most PRs were from first-time contributors who are completely unfamiliar with the team conventions, or if every PR descended into a flame war about code style. Luckily, neither smart_open nor gensim seem to suffer from that problem.

Thanks for the compliment :)

mpenkov (approved):
Looks good! Thank you for your contribution @justindujardin, and congrats on your first PR to smart_open 🥇

justindujardin:
@mpenkov any ballpark for when this might ship in a new pypi version? 🙏

mpenkov:
I need to fix https://github.com/RaRe-Technologies/smart_open/issues/538 first, after that nothing is blocking the release.

mpenkov:
https://pypi.org/project/smart-open/2.2.0/

PR #541, opened 2020-10-01
install s3 dependencies by default

### LINKED ISSUE #540: 2.2.0 minor release introduces breaking changes
Hi,

I am generally curious as to why `2.2.0` was released with such a breaking change to installation / usage for a minor version vs a major version? Also the release notes are not accurate as `aws` is not a valid extra for installation.

mpenkov:
Fix #540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,14 @@
 # Unreleased
 
+- Include S3 dependencies by default, because removing them in the 2.2.0 minor release was a mistake.
+
 # 2.2.0, 25 Sep 2020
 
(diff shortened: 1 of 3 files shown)


piskvorky (approved):
Ah, clever solution! Install by default, instead of reverting 👍 

How did you test this?

mpenkov:
> How did you test this?

I tested in a clean virtual environment.

```
virtualenv env
source env/bin/activate
pip install -e .
pip show boto3
```

where the last line outputs the version of `boto3` that got installed.

PR #542, opened 2020-10-01
Remove TODO in file opener logic

### LINKED ISSUE #519: Obsolete log in smart_lib_opener
Using the example from the Readme, the following code works as expected.

```python
from smart_open import open as smart_open
import pandas as pd

def cloud_file(azure_storage_connection_string):
    client = BlobServiceClient.from_connection_string(azure_storage_connection_string)
    return smart_open(uri, 'rb', transport_params=dict(client=client))

df = pd.read_csv(cloud_file(azure_storage_connection_string))
```

However, also getting the following log: `CRITICAL:smart_open.smart_open_lib:TODO`  (Found on v2.0.0, v2.1.0)
Is the code above missing anything or can [this](https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/smart_open_lib.py#L403) log line be removed?

Thanks

mpenkov:
At this stage, we're sure that `uri` is a string, so we can use it to
name any unnamed object.

Fix #519

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -400,8 +400,7 @@ def _open_binary_stream(uri, mode, transport_params):
     submodule = transport.get_transport(scheme)
     fobj = submodule.open_uri(uri, mode, transport_params)
     if not hasattr(fobj, 'name'):
-        logger.critical('TODO')
-        fobj.name = 'unknown'
+        fobj.name = uri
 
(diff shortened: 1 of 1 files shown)


piskvorky (approved):
Not sure about the implications of this… but it cannot be worse than `unknown` + logging a CRITICAL TODO to the user.

PR #543, opened 2020-10-01
Revert "install s3 dependencies by default (#541)"

mpenkov:
With 2.2.1 released, we can revert the temporary fix and go back to the recently introduced dependency behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This release modifies the behavior of setup.py with respect to dependencies.
 Previously, `boto3` and other AWS-related packages were installed by default.
 Now, in order to install them, you need to run either:
 
-    pip install smart_open[aws]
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ This release modifies the behavior of setup.py with respect to dependencies.
 Previously, `boto3` and other AWS-related packages were installed by default.
 Now, in order to install them, you need to run either:
 
-    pip install smart_open[s3]
+    pip install smart_open[aws]
```
piskvorky, line 13: Is this right?

mpenkov: Ah, yes, this particular part we should _not_ revert. Good catch.

mpenkov, line 13:
```suggestion
    pip install smart_open[s3]
```

PR #545, opened 2020-10-03
Add type annotations to smart_open

mpenkov:


piskvorky:
Check out [PEP 585](https://www.python.org/dev/peps/pep-0585/): starting with python3.7, annotations become infinitely more readable.

mpenkov:
Will that much change? From looking at that PEP, we'll go from using e.g. `Dict[str, int]` to `dict[str, int]` (and get rid of the `from typing import (...)`.

piskvorky:
To me – yes. I don't have anything against annotation but rarely use them, because I find they obfuscate more than clarify. Cons>pros.

A nicer, better integrated annotation syntax could tip the scales. But only a nice-to-have either way, nothing critical.

PR #546, opened 2020-10-05
Fix tests that were broken under Windows

### LINKED ISSUE #330: error while running tests in win10
When executing a command `python -m unittest discover` in the project directory:
- for Ubuntu -  the result is ok
- for Win10 - FAILED (failures = 4, errors = 18, skipped = 4)
image

interpolatio:
# Motivation

Tests work correctly in Linux, but they were executed with errors in Windows.

- Fixes #330 

#### Checklist


smart_open/tests/test_hdfs.py
```diff
@@ -31,23 +32,29 @@
 # such as Windows.
 #
 class CliRawInputBaseTest(unittest.TestCase):
+    def _setUp(self, test_file):
+        if platform == "linux" or platform == "linux2":
+            path = P.join(CURR_DIR, 'test_data/' + test_file)
+            self.cat = subprocess.Popen(['cat', path], stdout=subprocess.PIPE)
+        elif platform == "win32":
```
mpenkov, line 39: Other platforms (e.g. MacOS aka darwin) have cat. I suggest handling Windows as a special case, and expecting cat to be there otherwise.

smart_open/tests/test_hdfs.py
```diff
@@ -31,23 +32,29 @@
 # such as Windows.
 #
 class CliRawInputBaseTest(unittest.TestCase):
+    def _setUp(self, test_file):
+        if platform == "linux" or platform == "linux2":
+            path = P.join(CURR_DIR, 'test_data/' + test_file)
```
mpenkov, line 37: The join function will handle slashes for you (also same thing below).

```suggestion
            path = P.join(CURR_DIR, 'test_data' + test_file)
```

smart_open/tests/test_hdfs.py
```diff
@@ -17,6 +17,7 @@
 
 import smart_open.hdfs
 
+from sys import platform
```
mpenkov, line 20: Avoid importing stuff directly into the module namespace (namespaces are a good thing, let's use them).

```suggestion
import sys
```

smart_open/tests/test_hdfs.py
```diff
@@ -31,23 +32,29 @@
...
+            path = P.join(CURR_DIR, 'test_data/' + test_file)
+            self.cat = subprocess.Popen(['cat', path], stdout=subprocess.PIPE)
+        elif platform == "win32":
+            path = P.join(CURR_DIR, 'test_data\\' + test_file)
+            self.cat = subprocess.Popen(['type', path], stdout=subprocess.PIPE, shell=True)
+
     def test_read(self):
-        path = P.join(CURR_DIR, 'test_data/crime-and-punishment.txt')
-        cat = subprocess.Popen(['cat', path], stdout=subprocess.PIPE)
+        self._setUp(test_file='crime-and-punishment.txt')
```
mpenkov, line 44: This is over-engineered. You don't need to use inheritance and OO for this stuff. Do this instead:

At module level:

```python
def cat(path):
    executable = 'cat'
    shell = False
    if sys.platform == 'win32':
        executable = 'type'
        shell = 'True'

    return subprocess.Popen([executable, path, stdout=subprocess.PIPE, shell=shell)
```

then in your tests:

```python
with mock.patch('subprocess.Popen', return_value=cat(path)):
```

smart_open/tests/test_smart_open.py
```diff
@@ -679,15 +679,15 @@ def test_open_with_keywords(self):
         """This test captures Issue #142."""
         fpath = os.path.join(CURR_DIR, 'test_data/cp852.tsv.txt')
         with open(fpath, 'rb') as fin:
-            expected = fin.read().decode('cp852')
+            expected = fin.read().decode('cp852').replace('\r\n', '\n')
```
mpenkov, line 682: This is hacky. We shouldn't have to tweak the expected value like this. Remember, smart_open needs to behave exactly the same as the built-in open function in such conditions.

I think either:

1. We need to specify the line terminator explicitly when reading under Windows, or
2. There is a bug with how we handle line terminators under Windows

Do you have time to investigate?

mpenkov (changes requested):

smart_open/tests/test_smart_open.py
interpolatio: Yes, I have researched this question.

1. Cannot specify line terminator when opening files in binary mode.
2. When opened in text mode, Windows automatically replaces \p \n with \n. This conversion does not occur when the string is decoded. So I applied an explicit replacement.

I used this documentation:
https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files
In text mode, the default when reading is to convert platform-specific line endings (\ n on Unix, \ r \ n on Windows) to just \ n. When writing in text mode, the default is to convert occurrences of \ n back to platform-specific line endings.

If this option is incorrect, I am ready to think more.

mpenkov: OK then, so why not open both files in text mode? It seems like the binary flag is irrelevant here.

smart_open/tests/test_hdfs.py
```diff
@@ -30,24 +31,30 @@
 # Since these tests use cat, they will not work in an environment without cat,
 # such as Windows.
 #
+def cat(path):
+    executable = 'cat'
+    shell = False
+    if sys.platform == 'win32':
+        executable = 'type'
+        shell = 'True'
```
mpenkov, line 39: This should be a bool.

```suggestion
        shell = True
```

mpenkov:
OK, finally merged. Thanks @interpolatio!

PR #547, opened 2020-10-06
Implement option to write to a custom buffer during S3 writes

mpenkov:
Currently, we buffer in-memory. This works well in most cases, but can
be a problem when performing many uploads in parallel.  Each upload
needs to buffer at least 5MB of data (that is the minimum part size for
AWS S3). When performing hundreds or thousands of uploads, memory can
become a bottleneck.

This PR relieves that bottleneck by buffering to temporary files
instead. Each part gets saved to a separate temporary file and discarded
after upload. This reduces memory usage at the cost of increased IO with
the local disk.


smart_open/s3.py
```diff
@@ -184,6 +185,7 @@ def open(
         singlepart_upload_kwargs=None,
         object_kwargs=None,
         defer_seek=False,
+        diskbuffer=False,
         ):
```
piskvorky, line 189: De-dent the closing bracket (here and everywhere).

smart_open/s3.py
```diff
@@ -807,7 +833,8 @@ def _upload_next_part(self):
         logger.debug("upload of part #%i finished" % part_num)
 
         self._total_parts += 1
-        self._buf = io.BytesIO()
+        self._buf.close()
```
piskvorky, line 836: Is the temp file removal safe even if something goes wrong? We don't want to pollute the disk on errors.

piskvorky (commented):
What's the motivation? Who needs thousands of parallel uploads and why?

Seems like a strange obfuscation so -1 on including, unless clearly motivated.

menshikh-iv (approved):

help.txt
```diff
@@ -324,9 +328,9 @@ DATA
     __all__ = ['open', 'parse_uri', 'register_compressor', 's3_iter_bucket...
 
 VERSION
-    2.1.1
+    2.2.1
 
 FILE
-    /home/misha/git/smart_open/smart_open/__init__.py
+    /Users/misha/git/smart_open/smart_open/__init__.py
```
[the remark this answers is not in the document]
mpenkov: Migration between home and office :)

mpenkov:
> What's the motivation? Who needs thousands of parallel uploads and why?
> 
> Seems like a strange obfuscation so -1 on including, unless clearly motivated.

I'm partitioning a large (100GB, hundreds of millions of lines) dataset in preparation for some data processing. I need each partition to be relatively small, so I'm using a large number of partitions (hundreds/thousands). If I have all the partitions open for writing simultaneously, the partitioning is very simple:

```
for record in dataset:
    partition_index = calculate_partition(record)
    partitions[partition_index].write(record)
```

smart_open/s3.py
mpenkov: I refactored my solution. The user now passes in their own buffer (if they want to, otherwise nothing changes).

piskvorky: Ah, you shifted the problem on to the user. Clever :)

smart_open/s3.py
```diff
@@ -807,7 +828,9 @@ def _upload_next_part(self):
         logger.debug("upload of part #%i finished" % part_num)
 
         self._total_parts += 1
-        self._buf = io.BytesIO()
+
+        self._buf.seek(0)
+        self._buf.truncate(0)
```
piskvorky, line 833: This is interesting. Hypothetically, users could pass in another smart_open object as the buffer, right? As long as it supports `truncate(0)`.

piskvorky (approved):

smart_open/s3.py
mpenkov: In theory, yes. I haven't gone done that particular inception-like rabbit hole yet, though.

howto.md
```diff
@@ -197,6 +197,35 @@ It is possible to save both CPU time and memory by sharing the same resource acr
...
+```python
+import boto3
+from smart_open import open
+with tempfile.NamedTemporaryFile() as tmp:
+    tp = {'writebuffer': tmp}
+    with open('s3://bucket/key', 'w', transport_params=tp) as fout:
+        fout.write(lots_of_data)
+```
+
+This option reduces memory usage at the expense of additional disk I/O (writing to a reading from a hard disk is slower).
```
piskvorky, line 227: `This option reduces memory usage at the expense of additional disk I/O (writing to and reading from a hard disk is slower).`

mpenkov: I'll fix this on develop.

piskvorky: Thanks. Super minor, sorry I noticed it only now.

PR #554, opened 2020-10-30
Updated gcs.py to work with GCS generation id

creativesands:
example of GCS URI with generation id:
gs://bucket/file.txt#1603446325157502
with smart_open:
open('gs://bucket/file.txt#1603446325157502', transport_params={'client': client})

#### Motivation

Had to make smart_open be able to open specific versions of a GCS blob based on its generation id. Reads will use the generation id while writes will ignore it in the URI (a new id will be generated by GCS on write).

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -214,7 +214,12 @@ class Reader(io.BufferedIOBase):
         if client is None:
             client = google.cloud.storage.Client()
 
-        self._blob = client.bucket(bucket).get_blob(key)  # type: google.cloud.storage.Blob
+        parts = key.split('#')
+        if len(parts) == 2:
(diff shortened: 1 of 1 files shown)


petedannemann:
I think I'd prefer to add a transport parameter instead of tacking it onto the filename. This looks hacky to me. Look at `version_id` in [s3.py](https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/s3.py) for a similar example.

mpenkov:
@creativesands How's it going? Were you able to check out the suggestion from @petedannemann?

ddelange:
superceded by https://github.com/piskvorky/smart_open/pull/817

PR #559, opened 2020-11-17
Improve file mode handling

### LINKED ISSUE #556: ValueError when attempting to read/write gzip files using rt/wt mode
```
  File "/tmp/commudata-bas6ysce.venv/script.py", line 347, in main
    do_work()
  File "/tmp/commudata-bas6ysce.venv/script.py", line 253, in do_work
    with smart_open.open(input_url, 'rt') as fin, smart_open.open(with_key_url, 'wt') as fout:
  File "/home/misha/envs/databrewery/lib/python3.7/site-packages/smart_open-3.0.0-py3.7.egg/smart_open/smart_open_lib.py", line 226, in open
    decompressed = compression.compression_wrapper(binary, mode)
  File "/home/misha/envs/databrewery/lib/python3.7/site-packages/smart_open-3.0.0-py3.7.egg/smart_open/compression.py", line 96, in compression_wrapper
    return callback(file_obj, mode)
  File "/home/misha/envs/databrewery/lib/python3.7/site-packages/smart_open-3.0.0-py3.7.egg/smart_open/compression.py", line 62, in _handle_gzip
  File "/usr/lib/python3.7/gzip.py", line 164, in __init__
    raise ValueError("Invalid mode: {!r}".format(mode))
ValueError: Invalid mode: 'rt'
```

The workaround is to us r/w instead of rt/wt, but it's not convenient. We should handle this edge case in our code.

mpenkov:
Make it more robust to the arbitrary character ordering.

Fix #556

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -213,7 +213,11 @@ def open(
     #                          binary             decompressed       decode
     #
 
-    binary_mode = _get_binary_mode(mode)
+    try:
+        binary_mode = _get_binary_mode(mode)
+    except ValueError as ve:
+        raise NotImplementedError(ve.args[0])
+
     binary = _open_binary_stream(uri, binary_mode, transport_params)
     if ignore_ext:
         decompressed = binary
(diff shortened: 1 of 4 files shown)


mpenkov:
I disabled the HTTP tests because they seem to be failing on Travis now. Will investigate separately.

smart_open/smart_open_lib.py
```diff
@@ -317,7 +370,7 @@ def _open_binary_stream(uri, mode, transport_params):
         return uri
 
     if not isinstance(uri, str):
-        raise TypeError("don't know how to handle uri %r" % uri)
+        raise TypeError("don't know how to handle uri %s" % repr(uri))
```
piskvorky, line 373: Why this change?

smart_open/tests/test_smart_open.py
```diff
@@ -1621,6 +1635,46 @@ def test(self):
...
+            ('at+', 'ab+'),
+            ('a+t', 'ab+'),
+        ]
+    )
+    def test(self, mode, expected):
+        actual = smart_open.smart_open_lib._get_binary_mode(mode)
+        assert actual == expected
+
+    @parameterizedtestcase.ParameterizedTestCase.parameterize(
+        ('mode', 'expected'),
```
piskvorky, line 1664: I'm not familiar with `ParameterizedTestCase `, but why bother declaring two fields and then supply only one?

smart_open/smart_open_lib.py
```diff
@@ -233,6 +232,60 @@ def open(
...
+    if 'a' in mode:
+        transfer('a')
+    elif 'w' in mode:
+        transfer('w')
+    elif 'r' in mode:
+        transfer('r')
+    else:
+        raise ValueError(
+            "Must have exactly one of create/read/write/append "
+            "mode and at most one plus"
```
piskvorky, line 264: This doesn't match built-in `open`: `open(fname)` works, with a default `mode='r'`.

smart_open/smart_open_lib.py
mpenkov: uri is a namedtuple, so using it with `%` is a type error

```
>>> '%r' % (1, 2, 3)
TypeError: not all arguments converted during string formatting
```

smart_open/smart_open_lib.py
mpenkov: By this time, we've already applied the default, so that is covered.

The error handling here is to cover `open(filename, '+')`, which will trigger that error with the built-in open.

smart_open/tests/test_smart_open.py
mpenkov: You're right, that's copypasted from the test above. Will fix.

piskvorky (approved):

smart_open/tests/test_smart_open.py
mpenkov, line 1664:
```suggestion
        ('mode', ),
```

PR #562, opened 2020-11-21
Bump minimum Python version to 3.6

mpenkov:
Python 3.5 has reached end of life and we no longer need to support it.

Also made `requests` an optional dependency. It occupies over a MB of storage (via its chardet dependency) and that matters when deploying in tight environments, like AWS Lambda.

### DIFF
--- a/smart_open/webhdfs.py
+++ b/smart_open/webhdfs.py
@@ -16,7 +16,10 @@ import io
 import logging
 import urllib.parse
 
-import requests
+try:
+    import requests
+except ImportError:
+    MISSING_DEPS = True
 
 from smart_open import utils, constants
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,6 @@ os: linux
 
 jobs:
   include:
-    - python: '3.7'
-      name: "flake8"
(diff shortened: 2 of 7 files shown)


smart_open/webhdfs.py
```diff
@@ -16,7 +16,10 @@
 import logging
 import urllib.parse
 
-import requests
+try:
+    import requests
+except ImportError:
+    MISSING_DEPS = True
```
piskvorky, line 22: What are the implications of `MISSING_DEPS`?

mpenkov: We import modules like smart_open.http and smart_open.hdfs dynamically. Trapping the ImportError and setting MISSING_DEPS allows us to import these modules fully (for discovering their functionality) but warn the user if they try to actually invoke the functionality with the missing dependencies.

The code that does all that is here: https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/transport.py

piskvorky (approved):

piskvorky:
@mpenkov maybe mentioned these "optional installs" in README too?

Right now, the README contains examples for HTTP and WEBHDFS that will fail.

PR #567, opened 2020-12-09
add localstack to howto

mpenkov:
I had to use localstack for testing one of my smart_open-enabled apps, so I thought it'd be useful to mention it in the docs.


piskvorky (approved):
Looks good. Too niche for README, but Howto seems the right place for it.

PR #569, opened 2020-12-18
refactor s3 submodule to minimize resource usage

mpenkov:
Creating sessions and resources costs time and memory. If possible, smart_open should avoid creating resources by itself, and allow the user to specify them up front.

Here are some CPU-time benchmark results:

```
$ time python benchmark/read_s3.py < benchmark/urls.txt
real    1m20.786s
user    0m8.619s
sys     0m0.894s

$ time python benchmark/read_s3.py create_session < benchmark/urls.txt
real    1m45.826s
user    0m4.554s
sys     0m0.149s

$ time python benchmark/read_s3.py create_resource < benchmark/urls.txt
real    0m22.046s
user    0m1.474s
sys     0m0.065s

$ time python benchmark/read_s3.py create_session_and_resource < benchmark/urls.txt
real    0m21.086s
user    0m1.496s
```

There are memory benefits as well, but I didn't benchmark them, because the CPU benchmarks were compelling enough.

### DIFF
--- a/benchmark/read_s3.py
+++ b/benchmark/read_s3.py
@@ -1,0 +1,19 @@
+import sys
+
+import boto3
+import smart_open
+
+urls = [line.strip() for line in sys.stdin]
+
+tp = {}
+if 'create_session_and_resource':
+    tp['session'] = boto3.Session()
+    tp['resource'] = tp['session'].resource('s3')
+elif 'create_resource' in sys.argv:
+    tp['resource'] = boto3.resource('s3')
+elif 'create_session' in sys.argv:
+    tp['session'] = boto3.Session()
+
+for i, url in enumerate(urls, 1):
+    # print('%d/%d %s' % (i, len(urls), url))
+    smart_open.open(url, transport_params=tp).read()
--- a/benchmark/read_s3.py
+++ b/benchmark/read_s3.py
@@ -14,6 +14,5 @@ elif 'create_resource' in sys.argv:
 elif 'create_session' in sys.argv:
     tp['session'] = boto3.Session()
 
-for i, url in enumerate(urls, 1):
-    # print('%d/%d %s' % (i, len(urls), url))
+for url in urls:
     smart_open.open(url, transport_params=tp).read()
--- a/benchmark/read_s3.py
+++ b/benchmark/read_s3.py
@@ -6,7 +6,7 @@ import smart_open
 urls = [line.strip() for line in sys.stdin]
 
 tp = {}
-if 'create_session_and_resource':
+if 'create_session_and_resource' in sys.argv:
     tp['session'] = boto3.Session()
     tp['resource'] = tp['session'].resource('s3')
 elif 'create_resource' in sys.argv:
--- a/howto.md
+++ b/howto.md
@@ -182,22 +182,28 @@ It is possible to save both CPU time and memory by sharing the same resource acr
 
 ```
 >>> import boto3
->>> tp = {'resource': boto3.resoure('s3')}
->>> urls = ['s3://commoncrawl/robots.txt'] * 3  # These URLs can be unique
->>> for url in urls:
-...     with smart_open.open(url) as fin:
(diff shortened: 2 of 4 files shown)


smart_open/s3.py
```diff
@@ -201,8 +202,13 @@ def open(
         The minimum part size for multipart uploads.  For writing only.
     session: object, optional
         The S3 session to use when working with boto3.
+        If you don't specify this, then smart_open will create a new session for you.
+    resource: object, optonal
```
piskvorky, line 206:
```suggestion
    resource: object, optional
```

benchmark/read_s3.py
```diff
@@ -0,0 +1,19 @@
...
+if 'create_session_and_resource' in sys.argv:
+    tp['session'] = boto3.Session()
+    tp['resource'] = tp['session'].resource('s3')
+elif 'create_resource' in sys.argv:
+    tp['resource'] = boto3.resource('s3')
+elif 'create_session' in sys.argv:
+    tp['session'] = boto3.Session()
+
+for i, url in enumerate(urls, 1):
+    # print('%d/%d %s' % (i, len(urls), url))
```
piskvorky, line 18: Either use logging (at DEBUG level?) or remove please.

smart_open/s3.py
```diff
@@ -420,30 +427,68 @@ def read(self, size=-1):
...
+            logger.warning('ignoring resource_kwargs because resource was passed explicitly')
+        rw._session = None
+        rw._resource = resource
+    elif session:
+        rw._session = session
+        rw._resource = rw._session.resource('s3', **resource_kwargs)
+        rw._resource_kwargs = resource_kwargs
+    else:
+        rw._session = boto3.Session()
+        rw._resource = rw._session.resource('s3', **resource_kwargs)
```
piskvorky, line 459: How will such "caching" work in the presence of multi-threading and multi-processing?

Sounds risky, unless boto3 is safe / clever enough.

smart_open/s3.py
```diff
@@ -1123,7 +1160,7 @@ def _download_key(key_name, bucket_name=None, retries=3, **session_kwargs):
         raise ValueError('bucket_name may not be None')
 
     #
-    # https://geekpete.com/blog/multithreading-boto3/
+    # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-and-multiprocessing
```
piskvorky, line 1163: Oh, there we go. Sounds like this implicit caching is not a good idea.

piskvorky (changes requested):

smart_open/s3.py
mpenkov: It's not safe - boto3 explicitly mentions that resource objects aren't threadsafe/process-safe.

We could do the same thing in our docs, e.g.

> If you re-use boto3 session or resource objects with smart_open, take care not to share them across multiple threads or subprocesses, because that is unsafe (see boto3 docs).

howto.md
```diff
@@ -173,6 +173,32 @@ s3.ObjectVersion(bucket_name='smart-open-versioned', object_key='demo.txt', id='
...
+By default, calling `smart_open.open` with an S3 URL will create its own boto3 session and resource.
+These are expensive operations: they require both CPU time to construct the objects from a low-level API definition, and memory to store the objects once they have been created.
+It is possible to save both CPU time and memory by sharing the same resource across multiple `smart_open.open` calls, for example:
+
+```
+>>> import boto3
+>>> tp = {'resource': boto3.resoure('s3')}
+>>> urls = ['s3://commoncrawl/robots.txt'] * 3  # These URLs can be unique
+>>> for url in urls:
+...     with smart_open.open(url) as fin:
```
piskvorky, line 188: I don't get it. Where is the "same resource" shared?

A typo or some magic?

piskvorky (changes requested):

piskvorky (approved):

PR #582, opened 2021-02-09
Support tell() for text mode write on s3/gcs/azure

### LINKED ISSUE #581: io.UnsupportedOperation when calling tell()
#### Problem description

Performing a `file_obj.tell()` on a file object returned by `smart_open.open` results in an error in the 4.1.2 release. However, when I switch to the previous 4.1.0 release, this does not occur. A quick search seems to point to #578 which introduces use of `io.TextIOWrapper`. Calling `tell()` is not using the underlying `tell()` from the cloud specific writers. The original codecs implementation inherits functions like `tell()` from the file object that is passed in.

```python
cloud_location = 's3://examplebucket/foobar'
import smart_open
with smart_open.open(cloud_location, 'w') as f:
    print(f.tell())
```

Above code with a valid `cloud_location` throws the following:
```python
  File "<stdin>", line 2, in <module>
io.UnsupportedOperation: underlying stream is not seekable
```

#### Versions

Linux-5.4.0-1029-aws-x86_64-with-glibc2.29
Python 3.8.5 (default, Jul 28 2020, 12:59:40)
[GCC 9.3.0]
smart_open 4.1.2

#### Checklist

markopy:
This fixes #581

`TextIOWrapper` requires the underlying stream to indicate support for tell via `seekable()`. This was already implemented for S3/Azure/GCS when opening files in read mode. This PR does the same for write mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Support tell() for text mode write on s3/gcs/azure (PR [#582](https://github.com/RaRe-Technologies/smart_open/pull/582), [@markopy](https://github.com/markopy))
+
 # 4.1.2, 18 Jan 2021
 
 - Correctly pass boto3 resource to writers (PR [#576](https://github.com/RaRe-Technologies/smart_open/pull/576), [@jackluo923](https://github.com/jackluo923))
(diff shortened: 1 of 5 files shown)


mpenkov:
Merged. Thank you @markopy !

PR #583, opened 2021-02-10
Refactor S3, replace high-level resource/session API with low-level client API

mpenkov:
While functionally they are the same, the session/resource stuff is not safe to use across multiple threads and subprocesses, and the client stuff is. The former _is_ a bit easier to use directly than the latter, but this does not impact the user, because smart_open is doing all the work.

I also refactored the way we pass keyword parameters to boto3. Previously, we had a separate parameter for each boto3 function. This was a pain, because this made parameter lists longer for each new function, e.g.

- resource_kwargs
- multipart_upload_kwargs
- singlepart_upload_kwargs
- object_kwargs

This PR moves all of the above parameters into a single nested dict and introduces a wrapper that transparently injects the parameters into the required function call.

This does break backwards compatibility, so will need a major version bump when releasing.

### DIFF
--- a/MIGRATING_FROM_OLDER_VERSIONS.rst
+++ b/MIGRATING_FROM_OLDER_VERSIONS.rst
@@ -71,6 +71,24 @@ The `client_kwargs` dict can thus contain the following members:
 - `s3.Client.get_object`
 - `s3.Client.put_object`
 
+Here's a before-and-after example for connecting to a custom endpoint.  Before:
+
+.. code-block:: python
+
+    session = boto3.Session(profile_name='digitalocean')
+    resource_kwargs = {'endpoint_url': 'https://ams3.digitaloceanspaces.com'}
+    with open('s3://bucket/key.txt', 'wb', transport_params={'resource_kwarg': resource_kwargs}) as fout:
+        fout.write(b'here we stand')
+
+After:
+
+.. code-block:: python
+
+    session = boto3.Session(profile_name='digitalocean')
+    client = session.client('s3', endpoint_url='https://ams3.digitaloceanspaces.com')
+    with open('s3://bucket/key.txt', 'wb', transport_params={'client': client}) as fout:
+        fout.write(b'here we stand')
+
 See `README <README.rst>`_ and `HOWTO <howto.md>`_ for more examples.
 
 .. _resource API: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#service-resource
--- a/MIGRATING_FROM_OLDER_VERSIONS.rst
+++ b/MIGRATING_FROM_OLDER_VERSIONS.rst
@@ -4,6 +4,7 @@ Migrating to the new client-based S3 API
 Version of smart_open prior to 5.0.0 used the boto3 `resource API`_ for communicating with S3.
 This API was easy to integrate for smart_open developers, but this came at a cost: it was not thread- or multiprocess-safe.
 Furthermore, as smart_open supported more and more options, the transport parameter list grew, making it less maintainable.
+
 Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
 Functionally, very little changes for the smart_open user. 
 The only difference is in passing transport parameters to the S3 backend.
--- a/MIGRATING_FROM_OLDER_VERSIONS.rst
+++ b/MIGRATING_FROM_OLDER_VERSIONS.rst
@@ -1,7 +1,7 @@
 Migrating to the new client-based S3 API
 ========================================
 
-Version of smart_open prior to 5.0.0 used the boto3 `resource API_` for communicating with S3.
+Version of smart_open prior to 5.0.0 used the boto3 `resource API`_ for communicating with S3.
 This API was easy to integrate for smart_open developers, but this came at a cost: it was not thread- or multiprocess-safe.
 Furthermore, as smart_open supported more and more options, the transport parameter list grew, making it less maintainable.
 Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
--- a/MIGRATING_FROM_OLDER_VERSIONS.rst
+++ b/MIGRATING_FROM_OLDER_VERSIONS.rst
@@ -5,7 +5,7 @@ Version of smart_open prior to 5.0.0 used the boto3 `resource API`_ for communic
 This API was easy to integrate for smart_open developers, but this came at a cost: it was not thread- or multiprocess-safe.
 Furthermore, as smart_open supported more and more options, the transport parameter list grew, making it less maintainable.
 Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
-Functionally, the little changes for the smart_open user. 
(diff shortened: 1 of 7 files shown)


smart_open/s3.py
```diff
@@ -52,6 +52,26 @@
 _OUT_OF_RANGE = 'InvalidRange'
 
 
+class _ClientWrapper:
+    """Wraps a client to inject the appropriate keyword args into each method call.
+
+    The keyword args are a dictionary keyed by the fully qualified method name.
+    For example, S3.Client.create_multipart_upload.
+
+    See https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#client
```
menshikh-iv, line 61: Deserve moar examples & explanations in README

menshikh-iv (commented):
Awesome work @mpenkov :fire:

Just one note

smart_open/s3.py
mpenkov: I added more info to the howto and help.txt - it's a bit too low-level for the README.

menshikh-iv (approved):

mpenkov:
@piskvorky Can you please have a look at this? I'd like to merge and release so I can use this in one of my projects.

mpenkov:
@piskvorky Ping

smart_open/s3.py
```diff
@@ -687,13 +666,13 @@ def to_boto3(self):
         `boto3.s3.Object` may not necessarily affect the current instance.
 
         """
+        if not resource:
```
piskvorky, line 669: Safer & more idiomatic:

```suggestion
        if resource is None:
```

piskvorky (commented):
I did a cursory review, code seems good. But what about compatibility? What does this do to existing users?

mpenkov:
This is a breaking change. People doing stuff with kwargs will have to switch from using resource paramaters to using client ones. This will mean a major version bump.

piskvorky:
OK. And is this documented somewhere? A migration guide?

mpenkov:
Good point. I added a section to the [migration guide](https://github.com/RaRe-Technologies/smart_open/blob/client/MIGRATING_FROM_OLDER_VERSIONS.rst). Please have a look.

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
+Migrating to the new client-based S3 API
+========================================
+
+Version of smart_open prior to 5.0.0 used the boto3 `resource API_` for communicating with S3.
```
piskvorky:
```suggestion
Version of smart_open prior to 5.0.0 used the boto3 `resource API`_ for communicating with S3.
```

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
...
+This API was easy to integrate for smart_open developers, but this came at a cost: it was not thread- or multiprocess-safe.
+Furthermore, as smart_open supported more and more options, the transport parameter list grew, making it less maintainable.
+Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
```
piskvorky, line 7: Missing link target.

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
...
+Functionally, the little changes for the smart_open user.
```
piskvorky:
```suggestion
Functionally, little changes for the smart_open user. 
```

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
...
+    smart_open.open('s3://bucket/key', transport_params={'client': client})
+
+If you were passing any of the `*_kwargs` parameters, you will need to include them in `client_kwargs`, keeping in mind the following transformations.
+
+========================== ====================================== ==========================
+Parameter name             Resource API method                    Client API function
+========================== ====================================== ==========================
+`multipart_upload_kwargs`  `s3.Object.initiate_multipart_upload`_ `create_multipart_upload`_
+`object_kwargs`            `s3.Object.get`_                       `get_object`_
+`resource_kwargs`          ???                                    ???
```
piskvorky: What's this?

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
...
+More specifically, the following S3 transport parameters are no longer supported:
+
+- `multipart_upload_kwargs`
+- `object_kwargs`
+- `resource`
+- `resource_kwargs`
+- `session`
+- `singlepart_upload_kwargs`
+
+If you weren't using the above parameters, nothing changes for you.
```
piskvorky, line 20:
```suggestion
**If you weren't using the above parameters, nothing changes for you.**

```

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,75 @@
...
+========================== ====================================== ==========================
+Parameter name             Resource API method                    Client API function
+========================== ====================================== ==========================
+`multipart_upload_kwargs`  `s3.Object.initiate_multipart_upload`_ `create_multipart_upload`_
+`object_kwargs`            `s3.Object.get`_                       `get_object`_
+`resource_kwargs`          ???                                    ???
+`singlepart_upload_kwargs` `s3.Object.put`_                       `put_object`_
+========================== ====================================== ==========================
+
+The `client_kwargs` dict can thus contain the following members:
```
piskvorky, line 66: OK I didn't really understand this section. But I never used these extra params, so not the target audience.

Any way to be clearer, maybe give an example?

piskvorky (changes requested):

MIGRATING_FROM_OLDER_VERSIONS.rst
mpenkov: Still a work in progress, it has been fixed since.

mpenkov:
Updated https://github.com/RaRe-Technologies/smart_open/blob/client/MIGRATING_FROM_OLDER_VERSIONS.rst, please have a look.

MIGRATING_FROM_OLDER_VERSIONS.rst
```diff
@@ -1,3 +1,107 @@
+Migrating to the new client-based S3 API
+========================================
+
+Version of smart_open prior to 5.0.0 used the boto3 `resource API`_ for communicating with S3.
+This API was easy to integrate for smart_open developers, but this came at a cost: it was not thread- or multiprocess-safe.
+Furthermore, as smart_open supported more and more options, the transport parameter list grew, making it less maintainable.
+Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
```
piskvorky, line 7:
```suggestion

Starting with version 5.0.0, smart_open uses the `client API`_ instead of the resource API.
```

piskvorky (approved):

PR #589, opened 2021-03-04
Remove `tests` directory from package

e-nalepa:
Fix #588

This PR removes `tests` directory from package, and so reduces the size of the package by 60%.
(Keeping a small package size is important when using a package on AWS Lambda).

No new test is added, I checked the following points:
- Current tests are still passing
- `tests` directory is not any more present after a `$ pip install`

Before you create the PR, please make sure you have:

- [X] Picked a concise, informative and complete title
- [X] Clearly explained the motivation behind the PR
- [X] Linked to any existing issues that your PR will be solving
- [X] Included tests for any new functionality
- [X] Checked that all unit tests pass

PR #591, opened 2021-03-05
following up on #589, adjust manifest

mpenkov:
- Adjust MANIFEST.in as suggested by @piskvorky 
- Remove test package data from setup.py


mpenkov:
@e-nalepa Please have a look in case we missed anything else

piskvorky:
Thanks.

What's the workflow for testing locally now?

mpenkov:
Same as before, "python setup.py develop; pytest smart_open".

piskvorky (approved):

PR #593, opened 2021-03-17
Fix error when using GCP urls without deps

polm:
Currently if you run code like this:

    from smart_open import open
    open("gs://asdf")

This is a fake URL but it works for this problem as is. Without the
Google Cloud deps, you get an error that says to do this:

    pip install smart_open[gcs]

But that doesn't work. You have to do this:

    pip install smart_open[gcp]

It seemed easier to change setup.py to match the error message so that's what I've done here.

#### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
     extras_require={
         'test': tests_require,
         's3': aws_deps,
-        'gcp': gcp_deps,
+        'gcs': gcp_deps,
         'azure': azure_deps,
         'all': all_deps,
         'http': http_deps,
--- a/setup.py
+++ b/setup.py
@@ -36,11 +36,11 @@ def read(fname):
     return io.open(os.path.join(os.path.dirname(__file__), fname), encoding='utf-8').read()
 
 aws_deps = ['boto3']
-gcp_deps = ['google-cloud-storage']
+gcs_deps = ['google-cloud-storage']
 azure_deps = ['azure-storage-blob', 'azure-common', 'azure-core']
 http_deps = ['requests']
 
-all_deps = aws_deps + gcp_deps + azure_deps + http_deps
+all_deps = aws_deps + gcs_deps + azure_deps + http_deps
 tests_require = all_deps + [
     'mock',
     'moto[server]==1.3.14',  # Older versions of moto appear broken
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
     extras_require={
         'test': tests_require,
         's3': aws_deps,
-        'gcs': gcp_deps,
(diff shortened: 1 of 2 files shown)


piskvorky:
Thanks! Yes, we don't want schizophrenic terminology. We should decide on one acronym and use that consistently everywhere in docs & code.

setup.py
```diff
@@ -76,7 +76,7 @@ def read(fname):
     extras_require={
         'test': tests_require,
         's3': aws_deps,
-        'gcp': gcp_deps,
+        'gcs': gcp_deps,
```
mpenkov, line 79: Please use GCS everywhere.

```suggestion
        'gcs': gcs_deps,
```

setup.py
```diff
@@ -76,7 +76,7 @@ def read(fname):
     extras_require={
         'test': tests_require,
         's3': aws_deps,
-        'gcs': gcp_deps,
+        'gcs': gcs_deps,
```
mpenkov, line 79: You also need to change the definition of gcp_deps to dcs_deps above ;)

piskvorky: Please run a grep on `gcp` in the entire codebase + docs. Let's fix this once and for all! Thanks.

polm:
OK, just pushed a commit that I think should take care of the remaining gcp references. The only grep hit for `gcp` now is in some of the example URLs, which seem to point to real data.

mpenkov:
Thank you for your work and your patience @polm.

@piskvorky @petedannemann Is there a reason why we're using "gs://..." as the URL scheme, and "gcs" / "gcp" in the dependency list? Wouldn't it make sense to make sure that everything, meaning:

- the protocol schema (gs)
- the module name (smart_open.gs)
- the extra dependency (e.g. pip install smart_open[gs])

match up? I realize that a foolhardy consistency is the hobgoblin of little minds etc., but this seems to be one of the places where it's worth enforcing the consistency.

piskvorky:
I'm still of the same opinion:

> Yes, we don't want schizophrenic terminology. We should decide on one acronym and use that consistently everywhere: in all docs & code.

petedannemann:
I think this is my fault. I believe I tried to follow the same pattern as S3 in [`setup.py`](https://github.com/RaRe-Technologies/smart_open/blob/develop/setup.py#L38) with using the cloud provider name as the deps list. I think I overlooked that in the installation we use the object storage name. I am in favor of using `gs` for everything.

mpenkov:
OK. I'll merge these changes and take care of the module renaming in a separate PR. Thank you @polm for fixing the bug and drawing our attention to the problem! And congrats on your first PR to the smart_open repo!

PR #597, opened 2021-03-22
Fix potential infinite loop when reading from webhdfs

### LINKED ISSUE #592: readlines on files from hdfs using webhdfs hangs
#### Problem description

readlines on files from hdfs using webhdfs hangs

```python
with open("webhdfs://namenode/myfile.txt", "r") as f:
    lines = f.readlines()
```

The issue seems to be in the following line 
https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/webhdfs.py#L147

Sometimes ```self._response.raw.read(io.DEFAULT_BUFFER_SIZE)``` returns 0 bytes and StopIteration is never thrown resulting in the call hanging for ever. I fixed it locally with the following change which seems to work.

```python
while len(self._buf) < size:
    raw_data = self._response.raw.read(io.DEFAULT_BUFFER_SIZE)
    if len(raw_data) == 0:
        break

    self._buf += raw_data
```

#### Versions

Please provide the output of:

```python
import platform, sys, smart_open
print(platform.platform())
print("Python", sys.version)
print("smart_open", smart_open.__version__)
```
Linux-5.8.0-44-generic-x86_64-with-debian-bullseye-sid
Python 3.7.3 (default, Sep  3 2020, 16:41:28)
[GCC 9.3.0]
smart_open 4.2.0

#### Checklist

traboukos:
when data read would return 0 length. The loop was expecting a StopIteration
exception which was never thrown. The 0 bytes returned case is explicitly
handled.

- Fixes #592


mpenkov (approved):
Looks good to me!

mpenkov:
Merged. Thank you for your work @traboukos ! Congrats on your first PR to the smart_open project.

traboukos:
Thanks @mpenkov glad I could help

Issue #607, opened 2021-04-19
Add a feature to override file extension defining compression


dmcguire81:
#### Problem description

Transparent compression/decompression is a killer feature that could be made more broadly applicable, if decoupled from the extension of the underlying "file". Tying application-handling behavior to "file" extension is not a universally portable idea (for example, on classic UNIX, a "shebang" line encodes the same information), and some decoupling already exists in the current `ignore_ext` flag to `smart_open.open`. This feature would complete the decoupling by allowing the extension to be effectively overridden to *force* a compression-handling behavior, by supplying a `override_ext` string to `smart_open.open`. At that point, users would be free to name "files" without extensions, and still have the ability to take advantage of the great feature.

#### Steps/code to reproduce the problem

N/A (not a defect), but any file path without an extension is *not* eligible for transparent compression/decompression, currently.

#### Versions

```
>>> import platform, sys, smart_open
>>> print(platform.platform())
Darwin-19.6.0-x86_64-i386-64bit
>>> print("Python", sys.version)
Python 3.7.8 (default, Jul 27 2020, 17:21:35) 
[Clang 11.0.3 (clang-1103.0.32.59)]
>>> print("smart_open", smart_open.__version__)
smart_open 5.0.0.dev0
```

#### Checklist

Before you create the issue, please make sure you have:

- [X] Described the problem clearly
- [X] Provided a minimal reproducible example, including any required data
- [X] Provided the version numbers of the relevant software

mpenkov:
I think the current way we're handling compression is not ideal. It isn't obvious what the ignore_ext flag is doing, and it has no effect on files without an extension, as you've mentioned in the ticket and the associated PR.

This has been bothering me for a while, and I've come up with two solutions:

1. An additional parameter. As I've mentioned in the PR, it's not a particularly good idea. We already have tons of parameters. Plus, explaining how this new parameter will work in tandem with the existing ignore_ext flag will also confuse people.
2. Replace ignore_ext parameter with something that's more flexible and easier to explain to people. Of course, this would be a backwards-incompatible change, so we'd have to make a new major release for it.

Unless we find other alternatives, 2 is the way to go, but it will take some time to get it right. OTOH, we'd replace ignore_ext with a parameter like "compression", which will be an enum-like variable with the following values:

- none: no compression at all, equivalent to the current ignore_ext=True
- extension: default behavior, infer compression from extension, equievalent to current ignore_ext=False
[...]

PR #608, opened 2021-04-19
Add a feature to override file extension defining compression

dmcguire81:
#### Title

Add a feature to override file extension defining compression.

#### Motivation

```
- Implements #607
```

#### Tests

Added `test_smart_open.S3OpenTest.test_rw_gzip_override` and verified it fails as follows without the feature:

```
FAILED smart_open/tests/test_smart_open.py::S3OpenTest::test_rw_gzip_override  - OSError: Not a gzipped file...
```

Full test suite run output:

```
...
=============================================== short test summary info ===============================================
SKIPPED [1] smart_open/tests/test_package.py:18: requires missing dependencies
SKIPPED [1] smart_open/tests/test_package.py:13: requires missing dependencies
SKIPPED [1] smart_open/tests/test_package.py:24: requires missing dependencies
============================ 369 passed, 3 skipped, 7 warnings in 41.56s ==============================================
```

#### Checklist

### DIFF
--- a/smart_open/smart_open_lib.py
+++ b/smart_open/smart_open_lib.py
@@ -107,6 +107,7 @@ def open(
         closefd=True,
         opener=None,
         ignore_ext=False,
+        override_ext=None,
         transport_params=None,
         ):
     r"""Open the URI object, returning a file-like object.
@@ -139,6 +140,8 @@ def open(
         Mimicks built-in open parameter of the same name.  Ignored.
     ignore_ext: boolean, optional
         Disable transparent compression/decompression based on the file extension.
+    override_ext: str, optional
+        Force transparent compression/decompression on files based on supplied extension.
     transport_params: dict, optional
         Additional parameters for the transport layer (see notes below).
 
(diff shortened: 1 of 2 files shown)


mpenkov:
Thank you for your interest in smart_open. I understand your concern about handling compression behavior.

I don't think adding yet another parameter is the way to go here, because there are already a lot of parameters to the open function. So, in my opinion, this PR is premature. Let's keep the ticket open and discuss alternative solutions there.

dmcguire81:
As you mentioned, something that's backwards compatible will require a major release. If that's not on the horizon, what is the objection to "yet another parameter", if *is* backwards-compatible?

mpenkov:
It's more about cost-benefit than protection. For this PR, the cost is effort in reviewing, merging, documenting, explaining it to users, maintaining it and then eventually deprecating it once a proper solution is implemented. The benefit is immediately solving a problem that affects few people and does not strictly require immediate attention. It's simply not worth it.

I'd much rather invest that effort elsewhere, for example, into thinking of a proper way to solve the problem in #607.

PR #633, opened 2021-08-03
Work around changes to `urllib.parse.urlsplit`

judahrand:
#### Motivation

This attempts to fix the behaviour around question marks in URIs for S3 and GCS. `urllib.parse.urlsplit` is used and a hack was put in place to replace `?` with `\n` to avoid issues withs params splitting. However, the behaviour of `urllib.parse.urlsplit` has changed and now `\n`, `\t` and `\r` are all stripped from the URL before splitting.

https://bugs.python.org/issue43882

My solution is to instead use the 'null character'. This could technically appear in a GCS Blob name... but should it? To me this feels fairrrrrllllly safe?

#### Checklist

- [ x] Picked a concise, informative and complete title
- [ x] Clearly explained the motivation behind the PR
- [ x] Linked to any existing issues that your PR will be solving
- [ x] Included tests for any new functionality
- [ x] Checked that all unit tests pass

#### Workflow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ You may continue to use `ignore_ext` parameter for now, but it will be deprecate
 - Add new top-level compression parameter (PR [#609](https://github.com/RaRe-Technologies/smart_open/pull/609), [@dmcguire81](https://github.com/dmcguire81))
 - Drop mock dependency; standardize on unittest.mock (PR [#621](https://github.com/RaRe-Technologies/smart_open/pull/621), [@musicinmybrain](https://github.com/musicinmybrain))
 - Fix to_boto3 method (PR [#619](https://github.com/RaRe-Technologies/smart_open/pull/619), [@mpenkov](https://github.com/mpenkov))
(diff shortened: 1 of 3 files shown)


mpenkov:
Thank you for this contribution. I was wondering what's going on with those question marks... this PR makes it obvious where the problem was.

However, I'm not sure if using the null character is the best way around it. It's risky, because the same thing that happened to the newline can happen to the null: the library implementation can change (I wonder why they did this, but it's out of our control) and then we'll be back to where we started.

I think the most robust way to solve this is:

1. Check if the S3 URL contains the notorious question mark character
2. If it does, pick a random string of readable characters that **does not already occur in the URL**
3. Replace the question mark with that random string
4. Perform the split
5. Replace that random string with the question mark

I think that would be safe. What are your thoughts?

mpenkov:
@judahrand I improved your PR slightly by replacing null with a configurable placeholder.

@piskvorky Can you please have a look at this?

mpenkov:
Merged! Thank you @judahrand for taking care of this.

PR #635, opened 2021-08-03
Mark concurrency-related tests as "expected to fail" on MacOS

judahrand:
#### Motivation

Some tests fail to run on MacOS. I think this is probably due the use of `spawn` rather than `fork` for multiprocessing. This means that the keys held in memory by `moto` are lost in the subprocesses. My reasoning started with this comment: https://github.com/spulec/moto/issues/2111#issuecomment-480000481

And then looking here: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

I discovered that Windows and MacOS use `spawn` and Linux uses `fork`. It is quite possible (likely even?) that the `spawn` mechanism fails to identify the `moto` parts of the process memory as needed.

### DIFF
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -22,6 +22,7 @@ import boto3
 import botocore.client
 import botocore.endpoint
 import moto
+import pytest
 
 import smart_open
 import smart_open.s3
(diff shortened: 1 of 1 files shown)


judahrand:
These test failures should be fixed by #633

piskvorky:
Thanks.

>  It is quite possible (likely even?) that the spawn mechanism fails to identify the moto parts of the process memory as needed.

If that's the case, then that makes `moto` pretty useless. We don't want to run tests on Linux only, when smart_open works fine on any system (incl. Windows and Mac OSX).

judahrand:
> Thanks.
> 
> > It is quite possible (likely even?) that the spawn mechanism fails to identify the moto parts of the process memory as needed.
> 
> If that's the case, then that makes `moto` pretty useless. We don't want to run tests on Linux only, when smart_open works fine on any system (incl. Windows and Mac OSX).

Yup, I agree! It might be the case that it is safest to make all the tests which test the parallelisation require the `moto_server` like a few other tests do currently. This should avoid the issue if it is indeed an issue. Just makes the tests very slightly more painful to run locally.

mpenkov:
> If that's the case, then that makes `moto` pretty useless. We don't want to run tests on Linux only, when smart_open works fine on any system (incl. Windows and Mac OSX).

moto has it's quirks, like the problems with multiprocessing addressed in this PR, but it's still very useful for us, because it makes tests that don't touch multiprocessing (the vast majority of them) significantly easier. The alternative would be to run our own server (e.g. localstack) just for unit testing (like we do for some integration tests), which would be a huge PITA.

I think the general direction of this PR is correct: if we expect certain tests to fail on MacOS, then it's OK to disable them, or ignore their failure on that platform only. It would also be good to beef up the test suite with non-multiprocessing tests, but that's out of scope here.

mpenkov (approved):

mpenkov:
Merged. Thank you for your work @judahrand !!

PR #636, opened 2021-08-04
Don't leak compressed stream (#630)

### LINKED ISSUE #630: Leaked file object with gzip compression
#### Problem description

I am getting a warning that a file object is not closed when doing something like
```python3
from smart_open import open


def fun():
    with open("my_file.gz", encoding="utf-8") as f:
        print(f.read())


fun()
gc.collect()
```

This is because closing a GzipFile instance does not close the underlying stream. I didn't check whether the same applies to bz2.BZ2File and the support for .bz2 in smart_open.

https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/compression.py

```python3
def _handle_gzip(file_obj, mode):
    import gzip
```

https://docs.python.org/3/library/gzip.html#gzip.GzipFile

> Calling a GzipFile object’s close() method does not close fileobj, since you might wish to append more material after the compressed data. This also allows you to pass an io.BytesIO object opened for writing as fileobj, and retrieve the resulting memory buffer using the io.BytesIO object’s getvalue() method.

ampanasiuk:
* Don't leak compressed stream

Transparent compression works by adapting the binary stream with the
GzipFile or BZ2File class. These, however, by design don't close the
adapted stream. Add a matching
CompressionFormatTest.test_closes_compressed_stream test.
- Fixes #630


mpenkov:
Thank you for taking care of this! Merged.

PR #652, opened 2021-09-09
support container client and blob client for azure blob storage

### LINKED ISSUE #651: Supporting container SAS URLs for Azure Blob Storage
#### Supporting container SAS URLs for Azure Blob Storage

Thanks for making a very handy library!

I'm trying to compress batches of files into large archives and stream that to Azure Blob Storage. Based on a quick test, this code seems to work.

```python
import smart_open
from glob import iglob

transport_params = {
    'client': BlobServiceClient.from_connection_string(blob_connect_string),
}

with smart_open.open("azure://data/ziptest/batch.zip", mode="wb", transport_params=transport_params) as blob_stream:
    with zipfile.ZipFile(blob_stream, "w") as zf:
        for source_file in iglob(source_path+"/*"):
            with open(source_file, "rb") as sf:
                data = sf.read()
            zf.writestr(os.path.basename(source_file), data)
```

Based on least-privilege, I'd like to only give write permission to a specific container. Maybe something like:

```python
transport_params = {
    'container_client': ContainerClient.from_container_url(container_sas_url),
}
```

I'm wondering if you think something like that would be worth supporting? How difficult would it be to add? Or is there already support and I just haven't found it, yet? Thanks!

#### Versions

Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.10
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0]
smart_open 5.2.1

cbare:
#### Support ContainerClient and BlobClient in Azure storage

Support use of ContainerClient and BlobClient to access blobs in Azure storage in addition to existing support for BlobServiceClient. Overloads transport_params['client'] to accept any of the three types of clients supported by Azure storage, for example:

```python
transport_params = {
    'client': ContainerClient.from_container_url(container_sas_url),
}

with smart_open.open('azure://path/to/blob', mode='wb', transport_parameters=transport_parameters) as fout:
   fout.write(data)
```

#### Motivation

This allows for a least-privilege approach. You can access azure storage with a SAS URL scoped down to a specific container or blob, rather having the full run of the storage account.

```
- Fixes #651 
```

#### Tests

in smart_open/tests/test_azure.py

- test_read_container_client
- test_read_blob_client
- test_write_container_client

### DIFF
--- a/smart_open/azure.py
+++ b/smart_open/azure.py
@@ -11,6 +11,7 @@
 import base64
 import io
 import logging
+from typing import Union
 
 import smart_open.bytebuffer
 import smart_open.constants
@@ -68,7 +69,7 @@ def open(
         container_id,
         blob_id,
         mode,
-        client=None,  # type: azure.storage.blob.BlobServiceClient
+        client=None,  # type: Union[azure.storage.blob.BlobServiceClient,azure.storage.blob.ContainerClient,azure.storage.blob.BlobClient]
         buffer_size=DEFAULT_BUFFER_SIZE,
         min_part_size=_DEFAULT_MIN_PART_SIZE,
         max_concurrency=DEFAULT_MAX_CONCURRENCY,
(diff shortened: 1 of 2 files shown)


cbare:
Ah flake8, my old friend. Apparently, flake8 takes offense at long lines caused by the type annotations. I added a type alias to resolve that. This introduces the first use of import from typing and modern type annotation syntax, so far as I could tell. Does that impact support for older versions of Python?

cbare:
@mpenkov mind approving the CICD? Let me know what you think of the way I did the type annotations. I can change if a different way is better. Thanks!

mpenkov:
Sorry, I've been busy with gensim and other projects this week. I'll try to look at this more closely next week. Thank you for your efforts and your patience!

cbare:
I don't understand why that test is failing with this NameError.

```txt
smart_open/azure.py:46: in <module>
    azure.storage.blob.BlobServiceClient,
E   NameError: name 'azure' is not defined
```

Going back to using the Union in "# type:" comments, plus tagging those lines with "# noqa" seems to appease flake8. The noqa bit might cause a type-checker some grief, if one where to be run, but Visual Code seems not to mind.

mpenkov:
fixes #651

mpenkov:
Looks like tests are failing for an unrelated reason. I'll try to investigate over the weekend.

mpenkov:
OK, I finally got around to investigating this. The parameterizedtestcase package was misbehaving. I got rid of it, because it wasn't strictly necessary. I'll wait for the CI to complete and then continue reviewing this PR. Thanks for your patience!

PR #661, opened 2021-10-11
drop deprecated ignore_ext parameter

mpenkov:
This is a follow up from https://github.com/RaRe-Technologies/smart_open/issues/607

We currently have two mutually exclusive parameters, ignore_ext (deprecated) and compression. We introduced the new one and marked the old one as deprecated about 6 months ago. I think sufficient time has passed for us to remove this parameter.

We still have a function for compatibility with ancient code (smart_open.smart_open). That's still there and remains the same.

This is a backwards-incompatible change, so we should probably put it in the same release as #660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Unreleased
 
+This release deprecates the old `ignore_ext` parameter.
+Use the `compression` parameter instead.
+
+```python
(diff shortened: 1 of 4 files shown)


mpenkov:
OK, added migration docs. Please have a look.

mpenkov:
@piskvorky I'm going to merge this now, and release 6.0.0 within the next several days. Please have a look and let me know if anything looks fishy.

PR #673, opened 2021-12-02
Fix a changed boto3 documentation URL

ronkorving:
The URL about multiprocessing is now at https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-or-multiprocessing-with-resources

#### Title

A URL to a paragraph in the boto3 docs changed. This PR fixes a mention of that URL in the code comments of `s3.py`.

#### Motivation

As I was debugging performance issues, I came across this, so I figured I would fix it.


mpenkov:
Thank you!

PR #688, opened 2022-02-19
Make Python 3.7 the required minimum

mpenkov:
Tried adding Py3.10 tests under CI, but ran into problems with moto.


.github/workflows/python-package.yml
```diff
@@ -26,6 +22,10 @@ jobs:
             toxenv: "check_keys,py39-doctest,test_coverage,py39-integration"
             coveralls: true
 
+          # - python-version: '3.10'
+          # toxenv: "check_keys,py310-doctest,py310-test,py310-benchmark,py310-integration"
+          # result_key: benchmark-results-py310
```
piskvorky, line 27: Why is this commented out?

mpenkov: I'm debugging a problem with moto and Py3.10. We may not be able to add Py3.10 testing after all.

PR #697, opened 2022-04-30
Read importlib.metadata to find transport / compressor extensions

### LINKED ISSUE #691: Improve extension discovery using custom setuptools entry points
# Problem description

It would be great if `smart_open` could be extended using [setuptools / entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#advertising-behavior) capabilities.
For now, if user install a package that provide  new `smart_open` transport mechanism, it cannot be automaticaly registered.

## Example use case

**Please note**: schema is not relevant here.

In package `smart_ext_transport` I have following file:

```python
# smart_ext_transport/custom_transport.py
SCHEMA = "custom"
...
```

### Current situation

The only way to load extension for now are listed bellow:

**Option 1**: using `register_transport` everywhere and everytime it might be required :cry:

```python
register_transport("smart_ext_transport.custom_transport")
```

**Option 2**: using `__init__.py` and `import` statement:

Adding:

```python
# smart_ext_transport/__init__.py
register_transport("smart_ext_transport.custom_transport")
```

Then every time `custom` schema might be needed:

```python
import smart_ext_transport
```

### Improved situation with setuptools

In `setup.py` from `smart_ext_transport` we may add:

```python
from setuptools import setup

setup(
    name="smart_ext_transport",
    packages=["smart_ext_transport"],
    entry_points={"smart_open_transport": [
        "custom = smart_ext_transport.custom_transport"],
    },
)
```

Once package is installed, `smart_open` can automatically load every plugin that is registered using setuptools entry points.
User will no longer have to import package to register extension manually.

This is how pytest / setuptools CLI / flake8 / jupyter nb convert works for example.

# Versions

```
Linux-4.15.0-175-generic-x86_64-with-glibc2.27
Python 3.8.0 (default, Dec  9 2021, 17:53:27) 
[GCC 8.4.0]
smart_open 5.2.1
```

arthurlm:
#### Motivation

For now, if user install a package that provide  new `smart_open` transport mechanism, it cannot be automatically registered.
It would be great if `smart_open` could be extended using [setuptools / entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#advertising-behavior) capabilities.

Fixes #691 

##### Example use case

**Please note**: schema is not relevant here.

In package `smart_ext_transport` I have following file:

```python
# smart_ext_transport/custom_transport.py
SCHEMA = "custom"
...
```

##### Current situation

The only way to load extension for now are listed bellow:

**Option 1**: using `register_transport` everywhere and every time it might be required :cry:

```python
register_transport("smart_ext_transport.custom_transport")
```

**Option 2**: using `__init__.py` and `import` statement:

Adding:

```python
# smart_ext_transport/__init__.py
register_transport("smart_ext_transport.custom_transport")
```

Then every time `custom` schema might be needed:

```python
import smart_ext_transport
```

##### Improved situation with setuptools

In `setup.py` from `smart_ext_transport` we may add:

```python
from setuptools import setup

setup(
    name="smart_ext_transport",
    packages=["smart_ext_transport"],
    entry_points={"smart_open_transport": [
        "custom = smart_ext_transport.custom_transport"],
    },
)
```

Once package is installed, `smart_open` can automatically load every plugin that is registered using setuptools entry points.
User will no longer have to import package to register extension manually.

This is how pytest / setuptools CLI / flake8 / jupyter nb convert works for example.

Handling correct version of `importlib.metadata` is mainly inspired from `pytest` codebase.

See my issue about this PR: #691 

##### What it may change for the future ?

Future extension to `smart_open` may be moved out of core library.

Example use case may be new extensions libraries like:
- `smart_open_zstd` to add `zstandard` compress
- `smart_open_xxx_cloud` to add new cloud provider without having to implementing it in core library (like AliCloud).

Core library will only have to maintain base / robust API and no longer have to support all possible URI provider / compression standard.

#### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
         'all': all_deps,
         'http': http_deps,
         'webhdfs': http_deps,
+        ':python_version<"3.8"': ['importlib-metadata'],
     },
     python_requires=">=3.6,<4.0",
 
--- a/smart_open/compression.py
+++ b/smart_open/compression.py
@@ -7,9 +7,9 @@
 #
 """Implements the compression layer of the ``smart_open`` library."""
 import logging
-import importlib
-import importlib.metadata
 import os.path
+
+from smart_open.utils import find_entry_points
 
 logger = logging.getLogger(__name__)
 
@@ -158,5 +158,5 @@ def _register_compressor_entry_point(ep):
         logger.warning("Fail to load smart_open compressor extension: %s (target: %s)", ep.name, ep.value)
 
 
-for ep in importlib.metadata.entry_points().select(group='smart_open_compressor'):
+for ep in find_entry_points('smart_open_compressor'):
     _register_compressor_entry_point(ep)
--- a/smart_open/compression.py
+++ b/smart_open/compression.py
@@ -7,6 +7,8 @@
 #
 """Implements the compression layer of the ``smart_open`` library."""
 import logging
+import importlib
+import importlib.metadata
 import os.path
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


ddelange:
Thanks for the detailed proposal and the tests, and sorry for the slow reply.

We are going to pass on this. Automatically scanning installed packages for entry points and then importing plus registering whatever they advertise means any package present in the environment can silently inject code paths into smart_open. For a library at ~67M downloads/month that is a real attack surface and supply chain concern, and the failure modes (import side effects at startup, scheme/extension collisions, silently shadowed transports) are hard to reason about for fairly little gain.

The explicit path already covers this and stays flexible: a single `register_transport(...)` or `register_compressor(...)` call in your own code, or in your extension package's `__init__.py`, registers exactly what you want, when you want it. It edits the same global registry, just without the implicit discovery. That is the behavior we want to keep.

Closing with that. Appreciate the effort here.

PR #703, opened 2022-07-02
Add cert parameter to http transport params

stev-0:
#### Motivation

Although it seems to be a simple case of adding http transport parameters and passing them  to `requests`, there is no support for client certificate authentication for http

#### Tests

Added `test_smart_open::SmartOpenHttpTest::test_http_cert`


mpenkov (approved):
Looks good, thank you!

PR #718, opened 2022-08-25
Improve instructions for testing & contributing

Kache:
* Add cfg in pyproject.toml to support `pytest` without args
 * Add CONTRIBUTING.md with quickstart guide
 * Add .venv to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ target/
 
 # env files
 .env
+.venv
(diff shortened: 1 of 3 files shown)


Kache:
fyi @mpenkov
Did this while working on #719

cadnce:
Honestly, the more important thing for me would be understanding the auto formatted/config settings for the project. (Which could also end up in the pyproject.toml)

The rest of the setup seemed to be “standard python workflow”, but happy if the maintainers think it would be helpful!

mpenkov (approved):

mpenkov:
Thank you!

Issue #721, opened 2022-09-14
Resolve build problems with Github Actions


mpenkov:
Failing: https://github.com/RaRe-Technologies/smart_open/actions/runs/3049675356/jobs/4916085156

Passing: https://github.com/RaRe-Technologies/smart_open/actions/runs/2898563500/jobs/4610848056

There has only been one small change since the passing build, and it's unrelated to the problem being reported (an apparent circular import). I'm unable to reproduce the problem locally under Python 3.7 and 3.8 (on Linux). Suspect it may be a github actions problem.

mpenkov:
@piskvorky Pinging you just so you know. I did a release, and the CI exploded. This is surprising, because the same code built under the same conditions a few weeks ago.

PR #722, opened 2022-09-15
attempt to work around circular import #721

mpenkov:


piskvorky:
Was that it, the cause of our CI trouble? Sneaky stuff!

mpenkov:
Yeah, it was the root cause.

Issue #33, opened 2015-09-16
Support for reading and writing files directly to/from ftp


ziky90:
Another feature that I would like to have in smart_open is support for direct reading/writing of the data to/from ftp.

Anyone else interested in this? Does this make sense?

val314159:
I thought about making a scp interface in the worst possible way, (make a shell command).

trentgerman:
I'd be interested in seeing the implementation for this. I've used pysftp in the past for such tasks and it is very slow for large files (you're better of with sftp)

saschalang32:
Is this issue still relevant today? I'd like to work on that if that would be the case :)

t4nujms:
+1
ftplib currently has function retrbinary() where I currently download from FTP into a BytesIO object and then use smart_open to write to s3. This is the fastest I can get it so far.
Ideally I want the stream reading from FTP to write into S3 at the same time.

Would be great if smart_open can read FTP in the same way as the following example:
` with open('ftp://user@host/smart_open/tests/test_data/1984.csv') as fin:
...    with open('s3://bucket-name/smart_open/tests/test_data/1984.csv', 'w') as fout:
...        for line in fin:
...           fout.write(line)`

danielloader:
> +1
> ftplib currently has function retrbinary() where I currently download from FTP into a BytesIO object and then use smart_open to write to s3. This is the fastest I can get it so far.
> Ideally I want the stream reading from FTP to write into S3 at the same time.
> 
> Would be great if smart_open can read FTP in the same way as the following example (apologies the correct syntax indent not working in this comment box):
> 
> ```python
> with open('ftp://user@host/smart_open/tests/test_data/1984.csv') as fin:
>     with open('s3://bucket-name/smart_open/tests/test_data/1984.csv', 'w') as fout:
>         for line in fin:
>             fout.write(line)
> ```

Sort of related, but https://docs.python.org/3/library/shutil.html#shutil.copyfileobj is fantastic (and considerably faster) for copying between two file like objects like you're describing.

RachitSharma2001:
Hi everybody, is this issue still open? I know that this thread is quite old, but I still don't see this feature and I think it would be useful to add. I would love to work on it if possible.

RachitSharma2001:
As an update, I currently have a PR [here](https://github.com/RaRe-Technologies/smart_open/pull/723) that allows for opening ftp files using smart_open.

PR #723, opened 2022-09-16
Support for reading and writing files directly to/from ftp

RachitSharma2001:
#### Title
Support for reading and writing files directly to/from ftp
#### Motivation

As elaborated in issue #33, there is currently no support for reading from ftp servers. With my change, developers will be able to read ftp files using smart open by doing the following:
```
from smart_open import open
with open('ftp://user:password@host:port/dir1/dir2/file') as fin:
     for line in fin:
          print(line)
```
#### Tests
Added `test_ftp::FTPOpen`

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -137,6 +137,8 @@ jobs:
 
       - run: bash ci_helpers/helpers.sh enable_moto_server
         if: ${{ matrix.moto_server }}
+      
+      - run: bash ci_helpers/helpers.sh create_ftp_server
 
       - name: Run integration tests
         run: python ci_helpers/run_integration_tests.py
@@ -146,6 +148,8 @@ jobs:
 
       - run: bash ci_helpers/helpers.sh disable_moto_server
         if: ${{ matrix.moto_server }}
+      
+      - run: bash ci_helpers/helpers.sh delete_ftp_server
 
   benchmarks:
     needs: [linters,unit_tests]
--- a/ci_helpers/helpers.sh
+++ b/ci_helpers/helpers.sh
@@ -7,8 +7,17 @@ enable_moto_server(){
   moto_server -p5000 2>/dev/null&
 }
 
+create_ftp_server(){
+  docker run -d -p 21:21 -p 21000-21010:21000-21010 -e USERS="user|123|/home/user/dir" -e ADDRESS=localhost --name my-ftp-server delfer/alpine-ftp-server 
+}
+
 disable_moto_server(){
   lsof -i tcp:5000 | tail -n1 | cut -f2 -d" " | xargs kill -9
+}
+
+delete_ftp_server(){
+  docker kill my-ftp-server
+  docker rm my-ftp-server
 }
 
 "$@"
--- a/integration-tests/test_ftp.py
+++ b/integration-tests/test_ftp.py
@@ -1,0 +1,71 @@
+from __future__ import unicode_literals
+
+from smart_open import open
+
+def test_nonbinary():
+    file_contents = "Test Test \n new test \n another tests"
+    appended_content1 = "Added \n to end"
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file", "w") as f:
+        f.write(file_contents)
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file", "r") as f:
+        read_contents = f.read()
+        assert read_contents == file_contents
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file", "a") as f:
+        f.write(appended_content1)
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file", "r") as f:
+        read_contents = f.read()
+        assert read_contents == file_contents + appended_content1
+
+def test_binary():
+    file_contents = b"Test Test \n new test \n another tests"
+    appended_content1 = b"Added \n to end"
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file2", "wb") as f:
+        f.write(file_contents)
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file2", "rb") as f:
+        read_contents = f.read()
+        assert read_contents == file_contents
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file2", "ab") as f:
+        f.write(appended_content1)
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file2", "rb") as f:
+        read_contents = f.read()
+        assert read_contents == file_contents + appended_content1
+
+def test_line_endings_non_binary():
+    B_CLRF = b'\r\n'
+    CLRF = '\r\n'
+    file_contents = f"Test Test {CLRF} new test {CLRF} another tests{CLRF}"
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file3", "w") as f:
+        f.write(file_contents)
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file3", "r") as f:    
+        for line in f:
+            assert not CLRF in line
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file3", "rb") as f:    
+        for line in f:
+            assert B_CLRF in line
+
+def test_line_endings_binary():
+    B_CLRF = b'\r\n'
+    CLRF = '\r\n'
+    file_contents = f"Test Test {CLRF} new test {CLRF} another tests{CLRF}".encode('utf-8')
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file4", "wb") as f:
+        f.write(file_contents)
+
+    with open("ftp://user:123@localhost:21/home/user/dir/file4", "r") as f:    
+        for line in f:
+            assert not CLRF in line
+    
+    with open("ftp://user:123@localhost:21/home/user/dir/file4", "rb") as f:    
+        for line in f:
+            assert B_CLRF in line
--- a/smart_open/ftp.py
+++ b/smart_open/ftp.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Radim Rehurek <me@radimrehurek.com>
+#
+# This code is distributed under the terms and conditions
+# from the MIT License (MIT).
+#
+
+"""Implements I/O streams over FTP.
+"""
+
+import logging
+import urllib.parse
+import smart_open.utils
+from ftplib import FTP, error_reply
+import types
+logger = logging.getLogger(__name__)
+
+SCHEME = "ftp"
+
+"""Supported URL schemes."""
+
+DEFAULT_PORT = 21
+
+URI_EXAMPLES = (
+    "ftp://username@host/path/file",
+    "ftp://username:password@host/path/file",
+    "ftp://username:password@host:port/path/file",
+)
+
+
+def _unquote(text):
+    return text and urllib.parse.unquote(text)
+
+
+def parse_uri(uri_as_string):
+    split_uri = urllib.parse.urlsplit(uri_as_string)
+    assert split_uri.scheme == SCHEME, 'unexpected scheme: %r' % split_uri.scheme
+    return dict(
+        scheme=split_uri.scheme,
+        uri_path=_unquote(split_uri.path),
+        user=_unquote(split_uri.username),
+        host=split_uri.hostname,
+        port=int(split_uri.port or DEFAULT_PORT),
+        password=_unquote(split_uri.password),
+    )
+
+
+def open_uri(uri, mode, transport_params):
+    smart_open.utils.check_kwargs(open, transport_params)
+    parsed_uri = parse_uri(uri)
+    uri_path = parsed_uri.pop("uri_path")
+    parsed_uri.pop("scheme")
+    return open(uri_path, mode, transport_params=transport_params, **parsed_uri)
+
+
+def convert_transport_params_to_args(transport_params):
+    supported_keywords = [
+        "timeout",
+        "source_address",
+        "encoding",
+    ]
+    unsupported_keywords = [k for k in transport_params if k not in supported_keywords]
+    kwargs = {k: v for (k, v) in transport_params.items() if k in supported_keywords}
+
+    if unsupported_keywords:
+        logger.warning(
+            "ignoring unsupported ftp keyword arguments: %r", unsupported_keywords
+        )
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP(**kwargs)
+    try:
+        ftp.connect(hostname, port)
+    except Exception as e:
+        logger.error("Unable to connect to FTP server: try checking the host and port!")
+        raise e
+    try:
+        ftp.login(username, password)
+    except error_reply as e:
+        logger.error("Unable to login to FTP server: try checking the username and password!")
+        raise e
+    return ftp
+
+
+# transport paramaters can include any extra parameters that you want to be passed into FTP_TLS
+def open(
+    path,
+    mode="r",
+    host=None,
+    user=None,
+    password=None,
+    port=DEFAULT_PORT,
+    transport_params=None,
+):
+    if not host:
+        raise ValueError("you must specify the host to connect to")
+    if not user:
+        raise ValueError("you must specify the user")
+    if not transport_params:
+        transport_params = {}
+    conn = _connect(host, user, port, password, transport_params)
+    mode_to_ftp_cmds = {
+        "r": ("RETR", "r"),
+        "rb": ("RETR", "rb"),
+        "w": ("STOR", "w"),
+        "wb": ("STOR", "wb"),
+        "a": ("APPE", "w"),
+        "ab": ("APPE", "wb")
+    }
+    try:
+        ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
+    except KeyError:
+        raise ValueError(f"unsupported mode: {mode!r}")
+    ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
+    socket = conn.transfercmd(f"{ftp_mode} {path}")
+    fobj = socket.makefile(file_obj_mode)
+
+    def full_close(self):
+        self.orig_close()
+        self.socket.close()
+        self.conn.close()
+    fobj.orig_close = fobj.close
+    fobj.socket = socket
+    fobj.conn = conn
+    fobj.close = types.MethodType(full_close, fobj)
+    return fobj
--- a/smart_open/transport.py
+++ b/smart_open/transport.py
@@ -54,15 +54,15 @@ def register_transport(submodule):
     # Save only the last module name piece
     module_name = module_name.rsplit(".")[-1]
 
-    if hasattr(submodule, 'SCHEME'):
+    if hasattr(submodule, "SCHEME"):
         schemes = [submodule.SCHEME]
-    elif hasattr(submodule, 'SCHEMES'):
+    elif hasattr(submodule, "SCHEMES"):
         schemes = submodule.SCHEMES
     else:
-        raise ValueError('%r does not have a .SCHEME or .SCHEMES attribute' % submodule)
+        raise ValueError("%r does not have a .SCHEME or .SCHEMES attribute" % submodule)
 
-    for f in ('open', 'open_uri', 'parse_uri'):
-        assert hasattr(submodule, f), '%r is missing %r' % (submodule, f)
+    for f in ("open", "open_uri", "parse_uri"):
+        assert hasattr(submodule, f), "%r is missing %r" % (submodule, f)
 
     for scheme in schemes:
         assert scheme not in _REGISTRY
@@ -80,7 +80,9 @@ def get_transport(scheme):
     """
     global _ERRORS, _MISSING_DEPS_ERROR, _REGISTRY, SUPPORTED_SCHEMES
     expected = SUPPORTED_SCHEMES
-    readme_url = 'https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst'
+    readme_url = (
+        "https://github.com/RaRe-Technologies/smart_open/blob/master/README.rst"
+    )
     message = (
         "Unable to handle scheme %(scheme)r, expected one of %(expected)r. "
         "Extra dependencies required by %(scheme)r may be missing. "
@@ -94,13 +96,14 @@ def get_transport(scheme):
 
 
 register_transport(smart_open.local_file)
-register_transport('smart_open.azure')
-register_transport('smart_open.gcs')
-register_transport('smart_open.hdfs')
-register_transport('smart_open.http')
-register_transport('smart_open.s3')
-register_transport('smart_open.ssh')
-register_transport('smart_open.webhdfs')
+register_transport("smart_open.azure")
+register_transport("smart_open.ftp")
+register_transport("smart_open.gcs")
+register_transport("smart_open.hdfs")
+register_transport("smart_open.http")
+register_transport("smart_open.s3")
+register_transport("smart_open.ssh")
+register_transport("smart_open.webhdfs")
 
 SUPPORTED_SCHEMES = tuple(sorted(_REGISTRY.keys()))
 """The transport schemes that the local installation of ``smart_open`` supports."""
--- a/ci_helpers/run_integration_tests.py
+++ b/ci_helpers/run_integration_tests.py
@@ -9,6 +9,7 @@ subprocess.check_call(
         'pytest',
         'integration-tests/test_207.py',
         'integration-tests/test_http.py',
+        'integration-tests/test_ftp.py'
     ]
 )
 


smart_open/ftp.py
```diff
@@ -0,0 +1,91 @@
...
+    kwargs = {k: v for (k, v) in transport_params.items() if k in supported_keywords}
+
+    if unsupported_keywords:
+        logger.warning('ignoring unsupported ftp keyword arguments: %r', unsupported_keywords)
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    from ftplib import FTP_TLS
```
mpenkov, line 69: Thank you for the PR. Have a look at how the other submodules are handling dependencies.

- https://github.com/RaRe-Technologies/smart_open/blob/4c6bc38d6b35aa75e5e9b8041bb80f0f956ac486/smart_open/azure.py#L19-L23
- https://github.com/RaRe-Technologies/smart_open/blob/develop/setup.py#L79

mpenkov (changes requested):

smart_open/ftp.py
RachitSharma2001: Thank you for those resources. I have added the following code:
```
def _connect(hostname, username, port, password, transport_params):
    try:
        from ftplib import FTP_TLS
    except ImportError:
        raise Exception("Error: Unable to import FTP_TLS from ftplib library")
```

Let me know if there is anything else that I should add.

mpenkov: No, that's not how we want to handle it. What we want to do io is:

1. Add an extra to setup.py to handle the new dependency (see example above)
2. In the new submodule, import the new dependency at module level, and if it fails, set MISSING_DEPS=True (see example above)

smart_open/ftp.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if unsupported_keywords:
+        logger.warning('ignoring unsupported ftp keyword arguments: %r', unsupported_keywords)
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    try:
+        from ftplib import FTP_TLS
```
petedannemann, line 70: Will this work if the FTP server doesn't support TLS?

smart_open/ftp.py
```diff
@@ -0,0 +1,94 @@
...
+    except ImportError:
+        raise Exception("Error: Unable to import FTP_TLS from ftplib library")
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP_TLS(**kwargs)
+    try:
+        ftp.connect(hostname, port)
+        ftp.sendcmd(f'USER {username}')
+        ftp.sendcmd(f'PASS {password}')
```
petedannemann, line 78: I imagine many users who want to use TLS would like a secure connection, which would require adding `ftp.prot_p()`
https://docs.python.org/3/library/ftplib.html#ftplib.FTP_TLS.prot_p

smart_open/ftp.py
```diff
@@ -0,0 +1,94 @@
...
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    try:
+        from ftplib import FTP_TLS
+    except ImportError:
+        raise Exception("Error: Unable to import FTP_TLS from ftplib library")
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP_TLS(**kwargs)
```
petedannemann, line 74: I think if you change `hostname` to `host` the connecting will take place during the instantiation of this object and we can save all the network calls that occur down below
https://docs.python.org/3/library/ftplib.html#ftplib.FTP.connect

smart_open/ftp.py
RachitSharma2001: That is a great point. Sorry about that, I have changed the implementation to:
`from ftplib import FTP`
The purpose of this PR is to just add functionality for reading from an FTP server, not an FTPS server (with TLS protocol). Thank you for bringing this up. In a next PR, I can add functionality for handling FTPS servers.

smart_open/ftp.py
RachitSharma2001: This is a good point. For now, I am adding functionality for handling reading from a pure FTP server, but I am planning on making a future PR for handling reading from an FTPS server.

smart_open/ftp.py
RachitSharma2001: Thank you for bringing this up. I thought about adding somethign like the following:
`ftp = FTP(hostname=host, username=user, pwd=password, **kwargs)`
Unfortunetly, users may want to specify the specific port, and with the ftplib library for instatiating FTP objects, there is no way to pass in a port parameter (they only read from port 21, as explained [here](https://docs.python.org/3/library/ftplib.html#ftplib.FTP)). So for now, I have just kept the current implementation.

smart_open/ftp.py
RachitSharma2001: Sorry about that, and thank you for bringing up my mistake. It turns out that ftplib is not a pip module and thus does not need to be installed separately (all python interpreters automatically have this library available for import). Thus, I have not added anything to setup.py, as technically there are no new dependencies introduced by my PR. However, I have changed `ftp.py` to import the ftplib library and appropriately set MISSING_DEPS at the module level as opposed to the function level, as you have brought up. Let me know if there are any other issues. And again, sorry for misunderstanding your earlier comment.

smart_open/ftp.py
```diff
@@ -0,0 +1,114 @@
...
+"""Implements I/O streams over FTP.
+"""
+
+import getpass
+import logging
+import urllib.parse
+import smart_open.utils
+
+try:
+    from ftplib import FTP
```
RachitSharma2001, line 18: Added ftplib import at module level as opposed to function level.

smart_open/ftp.py
```diff
@@ -0,0 +1,114 @@
...
+except ImportError:
+    MISSING_DEPS = True
```
RachitSharma2001, line 20: Added ftplib import at module level as opposed to function level.

smart_open/ftp.py
mpenkov: Ah, I thought you're using some external library.

But if ftplib is part of the standard library, then how could it possibly be missing?

RachitSharma2001: That is a good point. I think that the missing dependencies test is unnecessary. I have removed it.

smart_open/ftp.py
```diff
@@ -0,0 +1,110 @@
...
+#
+
+"""Implements I/O streams over FTP.
+"""
+
+import getpass
+import logging
+import urllib.parse
+import smart_open.utils
+from ftplib import FTP
```
RachitSharma2001, line 16: Added ftplib import at module level as opposed to function level. No need to do missing dependencies checks as this module is part of the python standard library.

smart_open/ftp.py
```diff
@@ -0,0 +1,110 @@
...
+    smart_open.utils.check_kwargs(open, transport_params)
+    parsed_uri = parse_uri(uri)
+    uri_path = parsed_uri.pop("uri_path")
+    parsed_uri.pop("scheme")
+    return open(uri_path, mode, transport_params=transport_params, **parsed_uri)
+
+
+def convert_transport_params_to_args(transport_params):
+    supported_keywords = [
+        "keyfile",
```
petedannemann, line 60: I think these are outdated from FTP_TLS

smart_open/tests/test_ftp.py
```diff
@@ -0,0 +1,64 @@
...
+        ftp_obj().sendcmd.assert_has_calls([call("USER user"), call("PASS pass")])
+        ftp_obj().transfercmd.assert_called_with("RETR /dir1/dir2/dir3/file")
+
+    @patch("smart_open.ftp.FTP")
+    def test_open_with_transport_params(self, ftp_obj):
+        ftp_obj().transfercmd().makefile().name = "some/path"
+        smart_open.open(
+            "ftp://user@localhost/dir1/dir2/dir3/file",
+            transport_params={
+                "keyfile": "some_key_file",
```
petedannemann, line 42: Same outdated note as above for these tests

smart_open/ftp.py
RachitSharma2001: Thank you for bringing this up. I have updated the `convert_transport_params_to_args` function to reflect the possible arguments that can be passed into FTP, as specified [here](https://docs.python.org/3/library/ftplib.html#ftplib.FTP).

smart_open/tests/test_ftp.py
RachitSharma2001: Thank you for bringing this up. I have updated the tests to reflect the changes.

smart_open/ftp.py
```diff
@@ -0,0 +1,107 @@
...
+    transport_params=None,
+):
+    if not host:
+        raise ValueError("you must specify the host to connect to")
+    if not user:
+        user = getpass.getuser()
+    if not transport_params:
+        transport_params = {}
+    conn = _connect(host, user, port, password, transport_params)
+    fobj = conn.transfercmd(f"RETR {path}").makefile(mode)
```
petedannemann, line 106: `RETR` is only used for reading a file. I think we should add support for writing to a file on a FTP server as well

smart_open/ftp.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP(**kwargs)
+    try:
+        ftp.connect(hostname, port)
+        ftp.sendcmd(f"USER {username}")
```
petedannemann, line 80: I think we should use the `FTP.login` method instead of two `sendcmd`'s here

smart_open/ftp.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP(**kwargs)
+    try:
+        ftp.connect(hostname, port)
```
petedannemann, line 79: If an exception was thrown during `connect` we would raise it as an exception during logging in, which seems quite confusing for the user. What do we get out of throwing our own exception here? What are the exceptions for a failed `connect` and a failed `login`?

smart_open/tests/test_ftp.py
```diff
@@ -0,0 +1,83 @@
...
+import logging
+import unittest
+from unittest.mock import patch, call
+
+import smart_open
+from smart_open import ftp
+
+
+class FTPOpen(unittest.TestCase):
+    @patch("smart_open.ftp.FTP")
```
petedannemann, line 12: Honestly I'm not sure what these tests really give us. All it is testing is that we are calling certain methods from the ftplib. I'd suggest:

1. Running an FTP server in CI and writing integration tests against it
2. Implementing mocks that imitate functionality of a FTP server. There's some good examples of us doing this [here for GCS](https://github.com/RaRe-Technologies/smart_open/blob/4c6bc38d6b35aa75e5e9b8041bb80f0f956ac486/smart_open/tests/test_gcs.py#L46)

smart_open/ftp.py
RachitSharma2001: For a failed connect the following error is thrown:
`ConnectionRefusedError: [WinError 10061] No connection could be made because the target machine actively refused it`

For a failed login, the following is thrown:
`ftplib.error_perm: 530 Login incorrect.`

I felt that these errors are somewhat confusing and don't really tell the user the real issue, which is that either their host or port is wrongly entered or a ftp server doesn't exist at that host or port (if connect fails) or the username or password is wrong (if login fails).

petedannemann: Those errors are somewhat confusing, but are more informative and specific than anything we could provide. I wouldn't recommend suppressing the exceptions thrown by ftplib

petedannemann: Maybe a best of both worlds situation would be
```python3
try:
    ftp.connect(hostname, port)
except Exception e:
   logger.error("Unable to connect to FTP server: try checking the host and port!")
   raise e
try:
    ftp.login(username, password)
except Exception e:
   logger.error("Unable to login to FTP server: try checking the username and password!")
   raise e
```

piskvorky: +1 on **not** rephrasing the traceback errors. Because this is what users google. We want them to hit the actual support threads of the core libraries, where people have the same issue – and not smart_open, which is just a wrapper.

smart_open/ftp.py
RachitSharma2001: That is a good question. The entire file is not downloaded into memory. In fact, when open is called, it only returns a file like object representing the file on the FTP server, but it will not have downloaded any part of that file into memory. How the file is downloaded depends on what the developer wants to do with the object returned by `open`. For example:

```
# At this point, the file has not been downloaded at all
ftp_file_obj = open("ftp://username:password@host:port/file)

# Now, the file object could be read line by line (this is where the file from the ftp server is actually read)
for line in ftp_file_obj:
    print(line)
```

More specifically: `conn.transfercmd(f"RETR {path}")` returns a [socket object](https://docs.python.org/3/library/socket.html#socket.socket.makefile). When created, this object doesn't download any part of the file. You can see an example of the makefile function being called on a socket object [here](https://github.com/python/cpython/blob/main/Lib/ftplib.py#L463).

petedannemann: Good find and good to know we are streaming the file. In the ftplib.py file you linked it looks like `retrlines` has quite a lot of useful functionality for reading a file that we don't get with our current implementation though such as ensuring that we close the socket you mentioned so we don't leak resources and ensuring that we handle various newlines and CLRF's properly. To me it seems like we should use the higher level `retrlines` method with all of the functionality baked in

.github/workflows/python-package.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - {python: '3.7', os: ubuntu-20.04, moto_server: true}
```
mpenkov, line 108: Why did you do this?

ci_helpers/helpers.sh
```diff
@@ -7,8 +7,17 @@ enable_moto_server(){
   moto_server -p5000 2>/dev/null&
 }
 
+create_ftp_server(){
+  docker run -d -p 21:21 -p 21000-21010:21000-21010 -e USERS="user|123|/home/user/dir" -e ADDRESS=localhost delfer/alpine-ftp-server
```
mpenkov, line 11: Do we really need docker for something like this? We can use a more lightweight solution here: https://pypi.org/project/python-ftp-server/

integration-tests/test_ftp.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import unicode_literals
+
+import logging
+import unittest
+
+from smart_open import open
+
+class FTPTest(unittest.TestCase):
```
mpenkov, line 8: We use pytest, so you don't need a test class.  Just define the test_ functions at module level.

integration-tests/test_ftp.py
```diff
@@ -0,0 +1,54 @@
...
+        appended_content1 = "Added \n to end"
+
+        with open("ftp://user:123@localhost:21/home/user/dir/file", "w") as f:
+            f.write(file_contents)
+            f.close()
+
+        with open("ftp://user:123@localhost:21/home/user/dir/file", "r") as f:
+            read_contents = f.read()
+            assert read_contents == file_contents
+            f.close()
```
mpenkov, line 21: Why do you need to close here? The context manager does that for you.

smart_open/ftp.py
```diff
@@ -0,0 +1,120 @@
...
+    mode_to_ftp_cmds = {
+        "r": ("RETR", "r"),
+        "rb": ("RETR", "rb"),
+        "w": ("STOR", "w"),
+        "wb": ("STOR", "wb"),
+        "a": ("APPE", "w"),
+        "ab": ("APPE", "wb")
+    }
+    if mode not in mode_to_ftp_cmds:
+        raise Exception(f"Unable to support mode {mode}. Only 'r', 'rb', 'w', 'wb', 'a', 'ab' supported")
```
mpenkov, line 117:
```
try:
   ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
except KeyError:
   raise ValueError(f"unsupported mode: {mode})
```

.github/workflows/python-package.yml
RachitSharma2001: Sorry about that, I had accidently deleted this during debugging. I will re-add this on my next commit.

integration-tests/test_ftp.py
RachitSharma2001: Good catch. I have removed it.

ci_helpers/helpers.sh
RachitSharma2001: That is interesting. I have looked into the python-ftp-server. An interesting difference between this server and the docker image used above is that the python ftp server creates an FTP server with all the files within the smart_open directory copied to it (which could lead to unnecessary overhead), whereas the docker image creates an empty ftp server. 

In addition, when writing to the FTP server created by the docker image, the original smart_open directory is not affected. However, with the python-ftp-server, the written files will also be written to the smart_open directory during ci. 

Do you think this behavior be problematic? If so, I can also look at other lightweight ftp servers if we want to avoid using docker at all costs. If not, I will utilize this lighter-weight ftp server.

RachitSharma2001: Also, in terms of another potential local ftp server that could be created, I found [this](https://github.com/nixawk/pentest-wiki/blob/master/System-Services/services/twisted-ftp.md).

smart_open/ftp.py
```diff
@@ -0,0 +1,122 @@
...
+        "wb": ("STOR", "wb"),
+        "a": ("APPE", "w"),
+        "ab": ("APPE", "wb")
+    }
+    try:
+        ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
+    except KeyError:
+        raise ValueError(f"unsupported mode: {mode}")
+    ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
+    fobj = conn.transfercmd(f"{ftp_mode} {path}").makefile(file_obj_mode)
```
petedannemann, line 121: Two comments about this stemming from the [socket.makefile docs](https://docs.python.org/3/library/socket.html#socket.socket.makefile)

> the only supported mode values are 'r' (default), 'w' and 'b'

Do these other modes here actually work? It doesn't seem like they should

> Closing the file object returned by [makefile()]. 
 (https://docs.python.org/3/library/socket.html#socket.socket.makefile) won’t close the original socket unless all other file objects have been closed and [socket.close()](https://docs.python.org/3/library/socket.html#socket.close) has been called on the socket object.

I think we should make sure we are closing the underlying socket returned by `conn.transfercmd` here. On that note, do we also need to make sure we are closing the `conn` object?

RachitSharma2001: 1. Interestingly, append works when we do the following: 
```
fobj = conn.transfercmd("APPE {file}").makefile("w")
fobj.write("something")
```
The result is that "something" is appended to the end of the file. [Here](https://www.serv-u.com/resources/tutorial/appe-stor-stou-retr-list-mlsd-mlst-ftp-command) are examples of ftp commands.

2. Good point. I will work on potentially returning a wrapper around the file object that closes both the underlying socket and the conn object. While in the docs [here](https://docs.python.org/3/library/ftplib.html#ftplib.FTP) it doesn't explicitly say that we should close to the conn object (which is of type ftplib.FTP), I think it is best to still close it, as we never reuse this connection.

RachitSharma2001: I have updated ftp.py to have the following:
```
socket = conn.transfercmd(f"{ftp_mode} {path}")
fobj = socket.makefile(file_obj_mode)
def full_close(self):
    self.orig_close()
    self.socket.close()
    self.conn.close()
fobj.orig_close = fobj.close
fobj.socket = socket
fobj.conn = conn
fobj.close = types.MethodType(full_close, fobj)
```

Now, the ftp connection, the socket, and the file will all be closed.

RachitSharma2001:
@mpenkov @petedannemann 

Just to summarize the changes that I have made:
1. I have added integration tests which bring up an FTP server and tests reading, writing and appending to it.
2. I have created a wrapper around the file object returned by `open("ftp://...")` so that when the file is closed, the underlying socket is closed as well
3. For bringing up the FTP server during ci tests, there doesn't seem to be any viable lightweight solutions, besides perhaps [twisted](https://github.com/nixawk/pentest-wiki/blob/master/System-Services/services/twisted-ftp.md).

Let me know if there is anything that you think should be changed or added.

integration-tests/test_ftp.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import unicode_literals
+
+from smart_open import open
+
+def test_nonbinary():
+    file_contents = "Test Test \n new test \n another tests"
```
petedannemann, line 6: Could we add some tests for CLRF and B_CLRF? I noticed there is quite a lot of explicit handling of these in ftplib
https://github.com/python/cpython/blob/f00645d5dbf4cfa0b8f382c8977724578dff191d/Lib/ftplib.py#L70

smart_open/ftp.py
```diff
@@ -0,0 +1,132 @@
...
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP(**kwargs)
+    try:
+        ftp.connect(hostname, port)
+    except Exception as e:
+        logger.error("Unable to connect to FTP server: try checking the host and port!")
+        raise e
+    try:
+        ftp.login(username, password)
+    except Exception as e:
```
petedannemann, line 85: I think we can be more explicit for this error and use ftp.error_reply
https://github.com/python/cpython/blob/f00645d5dbf4cfa0b8f382c8977724578dff191d/Lib/ftplib.py#L418

smart_open/transport.py
```diff
@@ -100,6 +100,7 @@ def get_transport(scheme):
 register_transport('smart_open.http')
 register_transport('smart_open.s3')
 register_transport('smart_open.ssh')
+register_transport('smart_open.ftp')
```
petedannemann, line 103: Nit but it looks like we should keep the alphabetization here

petedannemann (commented):
Looks very good! Just a few more small comments

integration-tests/test_ftp.py
RachitSharma2001: That is a good point. I have added the tests `test_line_endings_non_binary` and `test_line_endings_binary`. My logic for these tests is that, if you look at retrlines vs retrbinary in ftplib [here](https://github.com/python/cpython/blob/main/Lib/ftplib.py#L421), you will notice that `retrlines` explicitly removes CRLF line endings for each line, but `retrbinary` does not. As a result, in both of the aforementioned tests, I test to make sure that when the file is read in non binary mode, there is no CRLF ending in any line, but when the file is read in binary mode, each line contains the CRLF endings (in order to make sure that the behavior mimics `ftplib`). 

I have tested ftplib `retrlines` vs `retrbinary` and have confirmed that, if you upload a file with CRLF endings, then calling `retrbinary` on that file will result in each line containing a CRLF ending, but when calling `retrlines` each line will not contain those endings.

RachitSharma2001:
@petedannemann Thank you! I have added your recommended changes. Let me know if you have any questions or if anything is confusing.

smart_open/ftp.py
petedannemann: Sorry, I don't think I was clear here. Raising the original exception seems fine, no need to wrap it. What I meant is that we should only be catching` ftp.error_reply` for a failed login, instead of the more general `Exception`.

RachitSharma2001: No problem. I have updated the code accordingly.

smart_open/ftp.py
```diff
@@ -0,0 +1,132 @@
...
+
+    return kwargs
+
+
+def _connect(hostname, username, port, password, transport_params):
+    kwargs = convert_transport_params_to_args(transport_params)
+    ftp = FTP(**kwargs)
+    try:
+        ftp.connect(hostname, port)
+    except error_reply as e:
```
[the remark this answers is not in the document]
petedannemann:
```python3
try: 
    ftp.connect(hostname, port) 
except Exception e: 
    logger.error("Unable to connect to FTP server: try checking the host and port!") 
    raise e 
try: 
    ftp.login(username, password) 
except error_reply as e: 
    logger.error("Unable to login to FTP server: try checking the username and password!") 
    raise e
```

petedannemann (changes requested):

smart_open/ftp.py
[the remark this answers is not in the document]
RachitSharma2001: Sorry about that. I have fixed accordingly.

petedannemann (approved):
Nice work on this!

RachitSharma2001:
@mpenkov Do these changes look good to you?

smart_open/ftp.py
```diff
@@ -0,0 +1,132 @@
...
+)
+
+
+def _unquote(text):
+    return text and urllib.parse.unquote(text)
+
+
+def parse_uri(uri_as_string):
+    split_uri = urllib.parse.urlsplit(uri_as_string)
+    assert split_uri.scheme in SCHEME
```
mpenkov, line 39:
```suggestion
    assert split_uri.scheme == SCHEME, f'
```

smart_open/transport.py
```diff
@@ -17,7 +17,7 @@
 
 logger = logging.getLogger(__name__)
 
-NO_SCHEME = ''
+NO_SCHEME = ""
```
mpenkov, line 20: The changes in this module appear cosmetic and irrelevant, please roll them back.

smart_open/ftp.py
```diff
@@ -0,0 +1,132 @@
...
+    host=None,
+    user=None,
+    password=None,
+    port=DEFAULT_PORT,
+    transport_params=None,
+):
+    if not host:
+        raise ValueError("you must specify the host to connect to")
+    if not user:
+        user = getpass.getuser()
```
mpenkov, line 104: I think it's a bad idea to use getpass here.

smart_open is a library, and there's no guarantee that anyone will be watching standard error for the password prompt. This kind of logic is better handled by whatever application is using the library. So, if they don't specify a password, then error out (same as for hostname)

smart_open/ftp.py
```diff
@@ -0,0 +1,132 @@
...
+        "rb": ("RETR", "rb"),
+        "w": ("STOR", "w"),
+        "wb": ("STOR", "wb"),
+        "a": ("APPE", "w"),
+        "ab": ("APPE", "wb")
+    }
+    try:
+        ftp_mode, file_obj_mode = mode_to_ftp_cmds[mode]
+    except KeyError:
+        raise ValueError(f"unsupported mode: {mode}")
```
mpenkov, line 119:
```suggestion
        raise ValueError(f"unsupported mode: {mode!r}")
```

mpenkov (changes requested):
Looks good, left you some comments.

smart_open/transport.py
RachitSharma2001: Sorry about that, I have removed the change.

RachitSharma2001:
@mpenkov Thank you for the suggestions. I have integrated the changes. Let me know if there is anything else that should be added.

mpenkov (approved):

mpenkov:
Thank you for your work @RachitSharma2001 !!

PR #726, opened 2022-09-28
Allow passing storage_class via gcs blob_properties

ddelange:
#### Title

Allow passing storage_class via gcs blob_properties

#### Motivation

We want to upload files to gcs with a specific storage_class, [similar](https://github.com/ddelange/pypicloud/blob/6edcc4269f1799bdd7b7a49539aa3ca8037b9ff4/pypicloud/storage/s3.py#L210) to s3.

It turns out you can't set a storage_class to a gcs blob object directly ref https://github.com/stevearc/pypicloud/issues/317.

Luckily, there is a [local-only](https://github.com/googleapis/python-storage/blob/8d5a6c3557c48fb885e7e5966d1d906398edc399/google/cloud/storage/_helpers.py#L254-L255) setter for the bucket. This can be used to set the default storage class temporarily, into which the blob will be uploaded.

#### Tests

2. Implement your bug fix.

    pytest smart_open

#### Work in progress

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -408,7 +408,10 @@ class Writer(io.BufferedIOBase):
         if client is None:
             client = google.cloud.storage.Client()
         self._client = client
-        self._blob = self._client.bucket(bucket).blob(blob)  # type: google.cloud.storage.Blob
+        _bucket = self._client.bucket(bucket)
+        if blob_properties and 'storage_class' in blob_properties:
+            _bucket.storage_class = blob_properties.pop('storage_class')
+        self._blob = _bucket.blob(blob)  # type: google.cloud.storage.Blob
         assert min_part_size % _REQUIRED_CHUNK_MULTIPLE == 0, 'min part size must be a multiple of 256KB'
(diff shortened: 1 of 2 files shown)


cadnce:
I dont think this is nessecary? This works just fine for me as-is (setting the storage-class when writing to the object initially) with the following snippet (and an object is created in the archive storage class)

```
from smart_open import gcs

with gcs.open(BUCKET_NAME,
              BLOB_NAME,
              mode='wt',
              blob_properties={
                  'storage_class': 'ARCHIVE',
              }) as fh:
    fh.write('data')
```

That said, from reading the docs i wasnt expecting it to. Based on the bug report in your linked ticket (pypicloud#320) are you sure the issue isn't just with the predefined_acl property?

ddelange:
Hi @cadnce, thanks for your reply, and for confirming this is already working for you! Looks like [source](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/blob.py#L4235) confirms it as well. Closing!

PR #727, opened 2022-09-28
Allow passing predefined_acl via gcs blob_properties

ddelange:
#### Title

Allow passing predefined_acl via gcs blob_properties

#### Motivation

Similar to #726, it should be possible to set an ACL for gcs uploads, as it is for s3.

It turns out you can't set an ACL to a gcs blob object directly ref https://github.com/stevearc/pypicloud/issues/317.

- Blocked by https://github.com/googleapis/python-storage/pull/878
- Blocked by https://github.com/googleapis/python-storage/pull/838

#### Tests

2. Implement your bug fix.

    pytest smart_open

#### Work in progress

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/gcs.py
+++ b/smart_open/gcs.py
@@ -15,6 +15,7 @@ try:
     import google.cloud.exceptions
     import google.cloud.storage
     import google.auth.transport.requests
+    _PREDEFINED_ACL_SUPPORTED = google.cloud.storage.__version__.split('.') >= ['2', '6']
 except ImportError:
     MISSING_DEPS = True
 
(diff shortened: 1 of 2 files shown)


smart_open/gcs.py
```diff
@@ -420,14 +420,20 @@ def __init__(
...
+        kwargs = {}
+
         if blob_properties:
+            if 'predefined_acl' in blob_properties:
+                kwargs['predefined_acl'] = blob_properties.pop('predefined_acl')
             for k, v in blob_properties.items():
                 setattr(self._blob, k, v)
 
+        if 'predefined_acl' in kwargs and google.cloud.storage.__version__.split('.') < ['2', '6', '0']:
+            raise NotImplementedError('GCS "predefined_acl" requires google-cloud-storage>=2.6.0')
```
ddelange, line 432: @mpenkov would this be the preferred way to handle the new feature?

ddelange: not very performant to check this every time. will make a flag on global scope. is `NotImplementedError` ok like this though?

mpenkov:
Is it possible to set the ACL after the upload is complete? If we can do this bypassing smart_open, that would be ideal.

ddelange:
do you mean as a fallback for gcs <= 2.6?
or as primary/only approach?

it was my first thought too. but since i dont use gcs, i had to do a deep dive into the code and have not found a way to set ACL after upload:

there is [`blob.update()`](https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob), but there is no attribute setter for ACL. 

so using predefined_acl kwarg in one of the uplod methods is the only way I could find to achieve blob upload with ACL. all uploads methods have this kwarg for a long time, ONLY the one smart_open uses didn't have it

mpenkov:
I meant as a primary (only) approach.

This looks relevant: https://cloud.google.com/storage/docs/xml-api/put-object-acls

ddelange:
looks like this PR can be superceded by https://github.com/RaRe-Technologies/smart_open/pull/729#issuecomment-1279949567

mpenkov:
OK, thank you. Closing in favor of #729

PR #731, opened 2022-10-11
DAOS plugin for smart_open

sridharbalachandriah:
#### Title

DAOS plugin for smart_open

#### Motivation

Distributed Asynchronous Object Storage (DAOS) is an open source software-defined scale-out object store that provides high bandwidth, low latency, and high I/O operations per second (IOPS) storage containers to HPC applications. This plugin enables python applications using smart_open to connect to DAOS object store through pyDAOS Key-Value interface (https://www.intel.com/content/www/us/en/developer/articles/case-study/unlock-the-power-of-daos-in-python-with-pydaos.html).

If you're adding a new feature, then consider opening a ticket and discussing it with the maintainers before you actually do the hard work.

#### Tests

Unit Tests for DAOS plugin:
1. A generic application with smart_open to open / write / close  and open /read/close to DAOS backend ran successfully.
2. An application with  Ray framework (https://www.ray.io/ , https://github.com/ray-project/ray) using smart_open was written to test the Ray specific smart_open interface. This test also passed with the current pull request with DAOS version 2.2 and 2.3 

#### Work in progress

#### Checklist

#### Workflow

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,9 +40,8 @@ aws_deps = ['boto3']
 gcs_deps = ['google-cloud-storage>=1.31.0']
 azure_deps = ['azure-storage-blob', 'azure-common', 'azure-core']
 http_deps = ['requests']
-daos_deps = ['pydaos']
 
-all_deps = aws_deps + gcs_deps + azure_deps + http_deps + daos_deps
+all_deps = aws_deps + gcs_deps + azure_deps + http_deps
 tests_require = all_deps + [
     'moto[server]',
     'pathlib2',
@@ -81,7 +80,6 @@ setup(
         'all': all_deps,
         'http': http_deps,
         'webhdfs': http_deps,
-        'daos': daos_deps,
     },
     python_requires=">=3.6,<4.0",
 
--- a/setup.py
+++ b/setup.py
@@ -40,8 +40,9 @@ aws_deps = ['boto3']
 gcs_deps = ['google-cloud-storage>=1.31.0']
 azure_deps = ['azure-storage-blob', 'azure-common', 'azure-core']
 http_deps = ['requests']
+daos_deps = ['pydaos']
 
-all_deps = aws_deps + gcs_deps + azure_deps + http_deps
+all_deps = aws_deps + gcs_deps + azure_deps + http_deps + daos_deps
 tests_require = all_deps + [
     'moto[server]',
     'pathlib2',
@@ -80,6 +81,7 @@ setup(
         'all': all_deps,
         'http': http_deps,
         'webhdfs': http_deps,
+        'daos': daos_deps,
     },
     python_requires=">=3.6,<4.0",
 
--- a/smart_open/daos.py
+++ b/smart_open/daos.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2022 Hewlett Packard Enterprise Development LP.
+# Copyright (C) 2022 Radim Rehurek <radim@rare-technologies.com>
+# Author(s): Sridhar Balachandriah, Hewlett Packard Enterprise Development LP.
 #
 # This code is distributed under the terms and conditions
 # from the MIT License (MIT).
@@ -20,8 +21,6 @@ import logging
 import smart_open.utils
 
 try:
-    import sys
-    sys.path.append("/opt/daos/lib64/python3.6/site-packages/")
     from pydaos import (DCont, DObjNotFound)
 except ImportError:
     MISSING_DEPS = True
--- a/smart_open/daos.py
+++ b/smart_open/daos.py
@@ -1,0 +1,311 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2022 Hewlett Packard Enterprise Development LP.
+#
+# This code is distributed under the terms and conditions
+# from the MIT License (MIT).
+#
+
+'''
+Implements reading and writing to/from DAOS
+
+The main entry point is the :func:`~smart_open.daos.open` function.
+
+Uses the pyDAOS KV store under the covers.
+
+'''
+
+import io
+import logging
+import smart_open.utils
+
+try:
+    import sys
+    sys.path.append("/opt/daos/lib64/python3.6/site-packages/")
+    from pydaos import (DCont, DObjNotFound)
+except ImportError:
+    MISSING_DEPS = True
+
+logger = logging.getLogger(__name__)
+
+SCHEME = "daos"
+
+URI_EXAMPLES = (
+    'daos://daos_pool_name/daos_container_name/offset',
+)
+
+
+def parse_uri(uri_as_string):
+    split_uri = smart_open.utils.safe_urlsplit(uri_as_string)
+    assert split_uri.scheme in SCHEME
+
+    uri_path = split_uri.netloc + split_uri.path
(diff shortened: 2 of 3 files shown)


smart_open/daos.py
```diff
@@ -0,0 +1,339 @@
...
+import logging
+import subprocess
+import urllib.parse
+from smart_open import utils
+import smart_open.utils
+import time
+
+try:
+    import sys
+    sys.path.append("/opt/daos/lib64/python3.6/site-packages/")
```
cadnce, line 28: This seems non-ideal. As I understand it python3.7 is the minimum supported for smart open https://github.com/RaRe-Technologies/smart_open/pull/688 ?

sridharbalachandriah: DAOS supports python with pyDAOS(https://www.intel.com/content/www/us/en/developer/articles/case-study/unlock-the-power-of-daos-in-python-with-pydaos.html) using python3.6. This is packaged separately. Not dependent on the smart_open supported version.

mpenkov: Why are you even touching sys.path inside the Python source?

smart_open/daos.py
```diff
@@ -0,0 +1,311 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2022 Hewlett Packard Enterprise Development LP.
```
mpenkov, line 3: @piskvorky We need the copyright transferred to you explicitly, right?

piskvorky: Yes. Managing copyright from a multitude of authors would be a nightmare. Too much legal trouble for zero (open source) gain.

mpenkov: @sridharbalachandriah Please see the other files for examples of what the copyright header should be.

sridharbalachandriah: @mpenkov ,@piskvorky, 

Would the copyright attached below suffice? 

# -*- coding: utf-8 -*-
#
# Copyright (C) 2022 Radim Rehurek <radim@rare-technologies.com>
# Copyright 2022 Hewlett Packard Enterprise Development LP.
#
# This code is distributed under the terms and conditions
# from the MIT License (MIT).
#

piskvorky: No, sorry. We want full control over `smart_open`, and having to coordinate with multiple copyright owners complicates that. We're happy to acknowledge your authorship of course, so a header like `Original auhor(s): X Y Z` is perfectly fine.

smart_open/daos.py
sridharbalachandriah: This is  a requirement when DAOS is not installed from repository packages. Will remove this.

smart_open/daos.py
sridharbalachandriah: @piskvorky , Thank you for the inputs.

Would the content outlined below work?

**
Copyright (C) 2022 Radim Rehurek [radim@rare-technologies.com](mailto:radim@rare-technologies.com)
Original author(s): X Y Z, Hewlett Packard Enterprise Development LP.

This code is distributed under the terms and conditions
from the MIT License (MIT).**

piskvorky: Yes, thanks.

mpenkov (changes requested):
Looks good, I think you also need to add the extra dependencies in setup.py (please see that file for how other backends handle dependencies). Thanks!

sridharbalachandriah:
> Looks good, I think you also need to add the extra dependencies in setup.py (please see that file for how other backends handle dependencies). Thanks!

Thanks, I have added the daos dependency in the setup.py and committed the code.

mpenkov:
Why did you close the PR?

sridharbalachandriah:
> s [ yesterday](#pullrequestreview-1152214593)
I was supposed to close the comment. Not the PR.

setup.py
```diff
@@ -40,9 +40,8 @@ def read(fname):
 gcs_deps = ['google-cloud-storage>=1.31.0']
 azure_deps = ['azure-storage-blob', 'azure-common', 'azure-core']
 http_deps = ['requests']
-daos_deps = ['pydaos']
 
-all_deps = aws_deps + gcs_deps + azure_deps + http_deps + daos_deps
+all_deps = aws_deps + gcs_deps + azure_deps + http_deps
```
mpenkov, line 44: Why did you revert this? The previous commit makes sense to me.

sridharbalachandriah: pydaos as a standalone library is not available. This dependency ends up with errors trying to auto install pydaos. The last set of automated tests failed. pydaos (dependency) gets installed with the DAOS package(rpm).

mpenkov: Ah, I just saw your previous comment. Thank you for explaining this to me (again!).

If pydaos gets installed via an rpm (or otherwise package), then where do the files from the package actually go? I imagine it would be the system's Python, as opposed to the virtual environment. So, that means smart_open would have to be installed into the system's Python as well, right?

sridharbalachandriah: Yes, that's right. By default, pydaos gets installed into system's Python. smart_open installed in the same system's python works fine. 
Example: 
/usr/lib64/python3.6/site-packages/pydaos
/usr/local/lib/python3.6/site-packages/smart_open

If there are different virtual environments created, then one has to link it up appropriately. 
The earlier comment on sys.path.append had the custom path for the pydaos library. pydaos documentation has reference to this. 
sys.path.append("/opt/daos/lib64/python3.6/site-packages/").

mpenkov: @piskvorky What are your thoughts on this? Should we wait for the pydaos thing to be pip-installable before we merge this?

piskvorky: I'll be honest, I don't know what DAOS (besides what's in the PR description), not a user. So my only concern is maintenance and I'm -1 on anything non-standard. We're busy enough as it is, not keen on introducing idiosyncrasies.

mpenkov: @sridharbalachandriah Speaking of tests, how would you test this reliably? I guess locally you have access to the real system, so you can test against that, but what about in CI? Have a look at how some of the other backends are tested: is it possible to **easily** test DAOS in CI?

sridharbalachandriah: @mpenkov , I did have a look at the CI's, a fully functional CI would require to emulate the pydaos(DAOS) functionality. I am checking with the DAOS team for ways to address this.

sridharbalachandriah: @mpenkov , currently DAOS docker container(https://docs.daos.io/v2.2/QSG/docker/) is not published. There is a stand alone mechanism to build DAOS docker container from the rpm, the build has to be  initiated before the DAOS CI tests are invoked. 

Does the smart_open CI infrastructure norms allow for the approach mentioned ?

mpenkov: No, not currently.

I think there are two show-stopper problems for this PR:

1) The underlying library is not pip-installable. Yes, it works under the system Python (some _some_ systems), but using the system Python is generally a bad practice, for several reasons. For example, you cannot expect the system's Python version to be what the user wants. So, it's impossible for people to use this contribution in the general case.
2) It's difficult to test this, so the maintenance burden is also high.

@sridharbalachandriah I really appreciate your effort, but I don't think this contribution is a good fit for smart_open right now.

@piskvorky

mpenkov:
Thank you for your efforts, but this PR is not a good fit for the smart_open project in its current state.

PR #732, opened 2022-10-12
Add flake8 config globally

cadnce:
When making a PR on the project I got caught out by the `flake8` linter and the build failed. Equally, I wasnt sure which formatter was in use by the project.

By configuring flake8 globally this it makes it possible to not get caught out by the CI linting step and ensure the code is formatted correctly on first commit. I guess this is somewhat similar to the changes in #718  but we worried about different aspects of contributing. Both cover different aspects that could be helpful.

I have made the list of errors that are faliling slightly more explicit to just the ones that are failing but equally happy to relax that again if prefered.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,9 @@
+[flake8]
+# E121: Continuation line under-indented for hanging indent
+# E123: Continuation line missing indentation or outdented
+# E125:	Continuation line with same indent as next logical line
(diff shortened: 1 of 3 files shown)


mpenkov (approved):

mpenkov:
Merged, thank you!

PR #734, opened 2022-10-19
Rewrite of s3.Reader class to protect S3 from open range headers (#725)

### LINKED ISSUE #725: [s3] Ranged read http header leaves TO bytes blank, causing S3 filesystems to read full file length
#### Problem description
`smart_open` is effectively crashing SeaweedFS/S3 and Ceph/S3 filesystems when doing many small reads over large files (ex: 32k read on a 4GB file).

On a SeaweedFS/S3 filesystem (also demonstrated on Ceph/S3), using the code shown in the reproduce section below, I am reading a small segment of data (32k in this example) from a large file (4GB in this example). This simple request causes SeaweedFS to move the full file for just a 32k range read. It appears that this is expected behavior based on a reading of the protocol specifications. Notably `boto3` does not trigger the same behavior.

When I run the code and look at the HTTP traffic being generated we see the following GET request:

```
GET /hengenlab/CAF77/Neural_Data/highpass_750/Headstages_256_Channels_int16_2021-02-02_14-28-24.bin HTTP/1.1
Host: seaweed-filer.seaweedfs:8333
Accept-Encoding: identity
Range: bytes=0-
User-Agent: Boto3/1.24.59 Python/3.8.2 Linux/5.4.0-125-generic Botocore/1.27.59
X-Amz-Date: 20220921T204638Z
X-Amz-Content-SHA256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
Authorization: AWS4-HMAC-SHA256 Credential=KL2PPBIZ4OYKR420C28D/20220921/us-east-1/s3/aws4_request, SignedHeaders=host;range;x-amz-content-sha256;x-amz-date, Signature=8009c4fdc85311977066c6988047a72658579e02c02b544fa8d48d8d8b9e8d57
amz-sdk-invocation-id: eace9bc1-a1d1-4244-84ee-1caa164bc294
amz-sdk-request: attempt=1
```

Notably `Range: bytes=0-` is our culprit. My assumption is that `smart_open` intends to open the file for streaming and read data from the stream as dictated by calls to `f.read(...)`.

When performing a ranged read with just `boto3` code the header looks like this:

```
GET /hengenlab/CAF77/Neural_Data/highpass_750/Headstages_256_Channels_int16_2021-02-02_14-28-24.bin?partNumber=1 HTTP/1.1
Host: seaweed-filer.seaweedfs:8333
Accept-Encoding: identity
Range: bytes=1-32768
User-Agent: Boto3/1.24.59 Python/3.8.2 Linux/5.4.0-125-generic Botocore/1.27.59
X-Amz-Date: 20220921T205137Z
X-Amz-Content-SHA256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
Authorization: AWS4-HMAC-SHA256 Credential=KL2PPBIZ4OYKR420C28D/20220921/us-east-1/s3/aws4_request, SignedHeaders=host;range;x-amz-content-sha256;x-amz-date, Signature=5d3398396c217f4a87284479bc6bc947344c256a30552fe98b6057167d7143fb
amz-sdk-invocation-id: 489f165b-cb4d-4ca0-bc49-cc1e70618518
amz-sdk-request: attempt=1
```

Using `boto3` does not cause any issue. With smart_open the SeaweedFS/S3 filesystem is interpreting the lack of a `to-bytes` value as the full file. It is then moving the full (4 GB) data file from a volume server to an S3 server, where just 32k are passed to the end user job. This has the effect of very quickly oversaturating the network. 

The protocol specifications seem to agree that the behavior by SeaweedFS/S3 is the correct way to interpret this Range header. E.g. how can the filesystem know that the user won't need to read the whole file given this header.

> If the last-byte-pos value is absent, or if the value is greater than or equal to the current length of the entity-body, last-byte-pos is taken to be equal to one less than the current length of the entity- body in bytes.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35

This code triggers the full (4 GB) file transfer (internally, not to the end user application) for a small 32k ranged read.
```
import import smart_open

while True:
    with smart_open.open('s3://bucket/path/to/4gb/file.bin', 'rb') as f:
        b = f.read(32768)
        print(f'Read {len(b)}')
```

This `boto3` version of the code does not trigger the same issue:

```
import boto3

while True:
    obj = boto3.resource('s3', endpoint_url='https://swfs-s3.endpoint').Object('bucket', 'path/to/4gb/file.bin')
    stream = obj.get(Range='bytes=1-32768')['Body']
    res = stream.read()
    print(f'Read: {len(res)} bytes')
```

I'd like to open a discussion to determine how to properly interpret the S3 protocol and figure out whether an issue like this should be in the domain of the filesystem(s) or should be changed in `smart_open`.

#### Versions

```
>>> print(platform.platform())
Linux-5.19.0-76051900-generic-x86_64-with-glibc2.17
Python 3.8.10 (default, Jun  4 2021, 15:09:15) 
[GCC 7.5.0]
smart_open 6.0.0
```

davidparks21:
#### Motivation

The `smart_open.s3.Reader` class left the HTTP range header open (e.g. `bytes=0-`). That was done to allow multiple (future) calls to `read` to stream bytes from the same HTTP body to avoid duplicating HTTP headers and connections. The open range header causes many S3 servers to internally move the full file, or a large block of the file, from its storage location to a server handling the S3 request. This was demonstrated on Ceph/S3 and SeaweedFS/S3, see issue #725. In the case of a small read against a large file this can quickly overwhelm the S3 filesystem.

```
- Fixes #725 
```

#### Work Performed

The class `smart_open.s3.Reader` was re-written to ensure the HTTP range headers are always set reasonably to protect the S3 filesystem. The rewrite strikes a balance between optimizing for the open-read-close use case and the open-repeat-read-close use case. See the git issue #725 for a detailed discussion.

#### Tests

Existing unit tests for `smart_open.s3` all pass. A few unit tests were removed for features that were eliminated and a few unit tests were added to cover new features. Some unrelated unit tests were updated because they failed to clean up after themselves.

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -38,7 +38,8 @@ SCHEMES = ("s3", "s3n", 's3u', "s3a")
 DEFAULT_PORT = 443
 DEFAULT_HOST = 's3.amazonaws.com'
 
-DEFAULT_BUFFER_SIZE = 128 * 1024
+DEFAULT_BUFFER_SIZE = 1500
+DEFAULT_STREAM_RANGE = 10485760
 
 URI_EXAMPLES = (
     's3://my_bucket/my_key',
@@ -232,7 +233,7 @@ def open(
     buffer_size=DEFAULT_BUFFER_SIZE,
     min_part_size=DEFAULT_MIN_PART_SIZE,
     multipart_upload=True,
-    defer_seek=False,
+    stream_range=10485760,
     client=None,
     client_kwargs=None,
     writebuffer=None,
@@ -260,11 +261,16 @@ def open(
     version_id: str, optional
         Version of the object, used when reading object.
         If None, will fetch the most recent version.
-    defer_seek: boolean, optional
-        Default: `False`
-        If set to `True` on a file opened for reading, GetObject will not be
-        called until the first seek() or read().
-        Avoids redundant API queries when seeking before reading.
+    stream_range: str, optional
+        Default: 10485760 bytes (10 MB)
+        The stream_range setting limits the size of data that may be streamed through a
+        single HTTP request across multiple read calls, this is an important protection
+        for the S3 server, ensuring the S3 server doesn't get an open-ended byte-range request
+        which can cause it to internally queue up a massive file when only a small bit of it may ultimately be
+        read by the user. Note that the first read call (after opening or seeking) will always set the
+        byte range header to exactly the read size, an optimization for use cases in which a single
+        small read is performed against a large file (example: random reading of small data samples
+        from large files in machine learning contexts).
     client: object, optional
         The S3 client to use when working with boto3.
         If you don't specify this, then smart_open will create a new client for you.
@@ -280,6 +286,7 @@ def open(
         disk IO. If you pass in an open file, then you are responsible for
         cleaning it up after writing completes.
     """
+
     logger.debug('%r', locals())
     if mode not in constants.BINARY_MODES:
         raise NotImplementedError('bad mode: %r expected one of %r' % (mode, constants.BINARY_MODES))
@@ -293,7 +300,7 @@ def open(
             key_id,
             version_id=version_id,
             buffer_size=buffer_size,
-            defer_seek=defer_seek,
+            stream_range=stream_range,
             client=client,
             client_kwargs=client_kwargs,
         )
@@ -338,6 +345,22 @@ def _get(client, bucket, key, version, range_string):
         raise wrapped_error from error
 
 
+def _head(client, bucket, key, version):
+    try:
+        if version:
+            return client.head_object(Bucket=bucket, Key=key, VersionId=version)
+        else:
+            return client.head_object(Bucket=bucket, Key=key)
+    except botocore.client.ClientError as error:
+        wrapped_error = IOError(
+            'unable to access bucket: %r key: %r version: %r error: %s' % (
+                bucket, key, version, error
+            )
+        )
+        wrapped_error.backend_error = error
+        raise wrapped_error from error
+
+
 def _unwrap_ioerror(ioe):
     """Given an IOError from _get, return the 'Error' dictionary from boto."""
     try:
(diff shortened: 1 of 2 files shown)


smart_open/s3.py
```diff
@@ -545,38 +564,47 @@ def __init__(
...
-        self._current_pos = 0
-        self._buffer = smart_open.bytebuffer.ByteBuffer(buffer_size)
+        self._content_length = None
         self._eof = False
+        self._position = 0
+        self._body = None
+        self._buffer = smart_open.bytebuffer.ByteBuffer(buffer_size)
         self._line_terminator = line_terminator
 
+        # the max_stream_size setting limits how much data will be read in a single HTTP request, this is an
```
mpenkov, line 588: Do you mean stream_range? I don't see max_stream_size anywhere.

smart_open/s3.py
```diff
@@ -545,38 +564,47 @@ def __init__(
...
+        # important protection for the S3 server, ensuring the S3 server doesn't get an open-ended byte-range request
+        # which can cause it to internally queue up a massive file when only a small bit of it may ultimately be
+        # read by the user. The variable _stream_range_[from|to] tracks the range of bytes that can be read
+        # from the current request body (e.g. from the same HTTP request). Note that the first read call
+        # will always set the byte range header to exactly the read size, an optimization for uses cases in which a
+        # single small read is performed against a large file (example: random sampling small data samples from
+        # large files in machine learning contexts).
+        self._stream_range = stream_range
+        self._stream_range_from = None  # a None value signifies the first call to `read` where this will be set
```
mpenkov, line 597: I think the current name is confusing, it's too similar to self._stream_range.

```suggestion
        self._stream_position_from = None  # a None value signifies the first call to `read` where this will be set
```

smart_open/s3.py
```diff
@@ -587,28 +615,21 @@ def readable(self):
...
-            self._current_pos = self._raw_reader._content_length
-            return out
 
-        #
-        # Return unused data first
-        #
-        if len(self._buffer) >= size:
-            return self._read_from_buffer(size)
+        # if an absolute size can be calculated we do calculate it to determine if it's available in the buffer already
+        if size < 0:
```
mpenkov, line 620: It's better to avoid nesting conditionals where possible, for the sake of simplicity.

```python
if size < 0 and content_length is None:
   size = -1
elif size < 0:
   size = self._content_length - self._position
```

smart_open/s3.py
```diff
@@ -587,28 +615,21 @@ def readable(self):
...
+            size = size if self._content_length is None else self._content_length - self._position
 
-        #
-        # If the stream is finished, return what we have.
-        #
-        if self._eof:
-            return self._read_from_buffer()
+        # Fill the buffer with at least enough data to satisfy the request
+        if size < 0 or len(self._buffer) < size:
```
mpenkov, line 624: Same thing here.

```python
if size < 0:
    self._fill_buffer(-1)
elif len(self._buffer) < size:
    fill_amount = max(size - len(self._buffer), self._buffer_size)
    self._fill_buffer(fill_amount)
```

smart_open/s3.py
```diff
@@ -657,21 +681,37 @@ def seek(self, offset, whence=constants.WHENCE_START):
...
-
-        self._current_pos = self._raw_reader.seek(offset, whence)
+        if whence not in constants.WHENCE_CHOICES:
+            raise ValueError('invalid whence, expected one of %r' % constants.WHENCE_CHOICES)
+        if whence == constants.WHENCE_END and offset > 0:
+            raise ValueError('offset must be <= 0 when whence == WHENCE_END, got offset: ' + offset)
+
+        if whence == constants.WHENCE_END and self._content_length is None:
+            # WHENCE_END: head request is necessary to determine file length if it's not known yet
+            #             this API call is necessary to return the absolute position as required by io.IOBase
```
mpenkov, line 691:
```suggestion
            # this API call is necessary to return the absolute position as required by io.IOBase
```

smart_open/s3.py
```diff
@@ -657,21 +681,37 @@ def seek(self, offset, whence=constants.WHENCE_START):
...
+            response = _head(self._client, self._bucket, self._key, self._version_id)
+            self._log_retry_attempts(response)
+            self._content_length = int(response['ContentLength'])
+            self._position = self._content_length + offset
+        elif whence == constants.WHENCE_END:
+            # WHENCE_END: we already have file length, no API call needed to compute the absolute position
+            self._position = self._content_length + offset
+        else:
```
mpenkov, line 699: Better to keep things simple here.

```suggestion
        elif whence == constants.WHENCE_START:
            self._position = offset
        elif whence == constants.WHENCE_CURRENT:
            self._position += offset
        else:
             raise ValueError(...)
```

You can also get rid of the whence check at the start of the function this way.

smart_open/s3.py
```diff
@@ -657,21 +681,37 @@ def seek(self, offset, whence=constants.WHENCE_START):
...
+            start = 0 if whence == constants.WHENCE_START else self._position
+            self._position = start + offset
 
         self._buffer.empty()
-        self._eof = self._current_pos == self._raw_reader._content_length
-        return self._current_pos
+        if self._body is not None:
+            self._body.close()
+            self._body = self._stream_range_from = self._stream_range_to = None
+        self._eof = False if self._stream_range_from is None or self._position < self._content_length - 1 else True
```
mpenkov, line 708: Please rewrite this as a block of conditionals.

```python
if self._stream_range_from is None or self._position < self._content_length - 1:
    self._eof = False
else:
    self._eof = True
```

smart_open/s3.py
```diff
@@ -700,20 +740,181 @@ def to_boto3(self, resource):
...
+        reached_eof = True if self._content_length is not None and self._position >= self._content_length else False
+
+        if reached_eof or size == 0:
+            return b''
+
+        if self._body is None:
+            stop = None if size == -1 else self._position + size
+            self._open_body(start=self._position, stop=stop)
+
+        #
```
mpenkov, line 786: This looks like copypasta from below, please refactor.

smart_open/s3.py
```diff
@@ -700,20 +740,181 @@ def to_boto3(self, resource):
...
     def _fill_buffer(self, size=-1):
-        size = max(size, self._buffer._chunk_size)
-        while len(self._buffer) < size and not self._eof:
-            bytes_read = self._buffer.fill(self._raw_reader)
-            if bytes_read == 0:
-                logger.debug('%s: reached EOF while filling buffer', self)
-                self._eof = True
+        # check if existing body range is sufficient to satisfy this request, if not close it so that it re-opens
+        # with an appropriate range.
+        is_stream_limit_before_eof = self._body is not None \
```
mpenkov, line 747: Please use all() to simplify this expression.

```
is_stream_limit_before_eof = all(
    self._body is not None,
    size < 0,
    ...
)
```

this avoids the repetitive "and" and artificial line breaks.

smart_open/s3.py
```diff
@@ -700,20 +740,181 @@ def to_boto3(self, resource):
...
+            and self._stream_range_to is not None \
+            and self._stream_range_to - self._position < size
+        if is_stream_limit_before_eof or is_request_beyond_stream_limit:
+            self._body.close()
+            self._body = self._stream_range_from = self._stream_range_to = None
+
+        # open the HTTP request if it's not open already
+        if self._body is None:
+            start = self._position + len(self._buffer)
+            stop = None if size < 0 else \
```
mpenkov, line 761: Please don't abuse the ternary operator :)

```
if size < 0:
    stop = None
else if self._content_length is None:
    stop = start + size - 1
else:
    stop = start + max(size, self._stream_range) - 1
```

smart_open/s3.py
```diff
@@ -700,20 +740,181 @@ def to_boto3(self, resource):
...
+        if bytes_read == 0:
+            logger.debug('%s: reached EOF while filling buffer', self)
+            self._eof = True
+
+    def _raw_read(self, size=-1):
+        """Internal read from the continuous connection with the remote peer without considering buffering."""
+
+        # If we can figure out that we've read past the EOF, then we can save
+        # an extra API call.
+        reached_eof = True if self._content_length is not None and self._position >= self._content_length else False
```
mpenkov, line 777: Ternary operator.

Also, here's a question: why do we spend so much effort handling the case when the content length is None? I think we ever reach a point where we need it and don't have it, we should just query it.

What do you think?

mpenkov (changes requested):
Left you a ton of comments, please have a look. Thanks!

mpenkov:
@piskvorky There's a ton of work left for this PR, my recommendation is to remove it from the milestone and deal with it after we release 6.3.0

ddelange:
fwiw, we routinely do a `fp.read()` (without size argument) on a ~1GB file and expect that to be a single GET request. If that would be split in 10MB reads internally by smart_open, that would increase the amount of GET requests (and costs) x100. Can the default behaviour of `fp.read()` remain unchanged, i.e. performing only [one single `get_object` call](https://github.com/piskvorky/smart_open/blob/v7.0.5/smart_open/s3.py#L434)?

ddelange:
superceded by #883

PR #735, opened 2022-11-06
Fix avoidable S3 race condition (#693)

### LINKED ISSUE #693: Probably avoidable fatal S3 race condition
#### Problem description

When using `s3.iter_bucket` from my Lambda, I observed the following problem.

A user on my system was deleting an object (valid use case). While that user was deleting it, unrelated to it, `smart_open.s3.iter_bucket` kicked off. Inside that function, the key iterator is created, and then a download is started for each key. The key (of the object about to be deleted) showed up in the result. Once it was time to download the object, it was no longer there, and a 404 was thrown, which proved fatal and an exception was thrown.

While this may seem like a fluke,

a) it was on a system with very little user activity.
b) S3's list operation is (iirc) not strongly consistent and may for a certain duration return keys that have already been deleted.

So it may be more common than it seems. I think that it would be reasonable for `iter_bucket` to skip objects that return a 404 during their download (ie: catch the error and suppress it), and carry on iterating. I think `iter_bucket` would still fulfill its duty this way.

The code:

```python
def get_all_metadata() -> list[dict]:
    bucket = get_s3_bucket()
    prefix = f"{get_s3_path()}/"
    accept_key = lambda key: key.endswith(get_metadata_ext())

    # Not explicitly setting this property to False will cause an error in the Lambda runtime when
    # running iter_bucket. This is a known issue, but it doesn't look like it will be resolved:
    # https://github.com/RaRe-Technologies/smart_open/issues/605

    smart_open.concurrency._MULTIPROCESSING = False

    return [
        json.loads(content)
        for key, content in smart_open.s3.iter_bucket(
            bucket, prefix=prefix, accept_key=accept_key
        )
    ]
```

The error:

```
[ERROR] ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
  File "/var/task/lambdas/filemanager/list.py", line 20, in handler
    metadata_list = filemanager.get_all_metadata()
  File "/var/task/lambdas/lib/filemanager.py", line 78, in get_all_metadata
    return [
  File "/var/task/lambdas/lib/filemanager.py", line 78, in <listcomp>
    return [
  File "/var/task/smart_open/s3.py", line 1192, in iter_bucket
    for key_no, (key, content) in enumerate(result_iterator):
  File "/var/task/smart_open/concurrency.py", line 58, in imap_unordered
    yield future.result()
  File "/var/lang/lib/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/var/lang/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/var/lang/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/var/task/smart_open/s3.py", line 1254, in _download_key
    content_bytes = _download_fileobj(bucket, key_name)
  File "/var/task/smart_open/s3.py", line 1271, in _download_fileobj
    bucket.download_fileobj(key_name, buf)
  File "/var/runtime/boto3/s3/inject.py", line 719, in bucket_download_fileobj
    return self.meta.client.download_fileobj(
  File "/var/runtime/boto3/s3/inject.py", line 679, in download_fileobj
    return future.result()
  File "/var/runtime/s3transfer/futures.py", line 103, in result
    return self._coordinator.result()
  File "/var/runtime/s3transfer/futures.py", line 266, in result
    raise self._exception
  File "/var/runtime/s3transfer/tasks.py", line 269, in _main
    self._submit(transfer_future=transfer_future, **kwargs)
  File "/var/runtime/s3transfer/download.py", line 354, in _submit
    response = client.head_object(
  File "/var/runtime/botocore/client.py", line 391, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/var/runtime/botocore/client.py", line 719, in _make_api_call
    raise error_class(parsed_response, operation_name)
```

#### Versions

This is running on:

* AWS Lambda
* Intel
* Python 3.9
* smart_open 5.2.1

#### Checklist

RachitSharma2001:
#### Motivation

Fixes #693 

As proposed by @ronkorving, it may be common for the s3 `iter_bucket` function to encounter `404 not found` errors due to the s3 `list bucket` command sometimes returning recently deleted keys. So a workaround of this problem is to surpress 404 errors for `iter_bucket`, and instead log them as warnings for the user to assess.

#### Tests

All existing tests related to `iter_bucket` within `s3.py` pass. I also added two new tests: `test_iter_bucket_404` and `test_iter_bucket_non_404`. The first test simulates the scenario with `iter_bucket` for a key throws a 404 error, and expects that error to be suppressed and for computation to carry on for the rest of the keys. The latter test simulates the scenario where a non 404 error occurs, in which case it should not be suppressed.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -1193,7 +1193,7 @@ def iter_bucket(
         while True:
             try:
                 (key, content) = result_iterator.__next__()
-                if True or key_no % 1000 == 0:
+                if key_no % 1000 == 0:
                     logger.info(
                         "yielding key #%i: %s, size %i (total %.1fMB)",
                         key_no, key, len(content), total_size / 1024.0 ** 2
@@ -1204,7 +1204,11 @@ def iter_bucket(
                     # we were asked to output only a limited number of keys => we're done
                     break
             except botocore.exceptions.ClientError as err:
-                # ignore 404 not found errors
+                #
+                # ignore 404 not found errors: they mean the object was deleted
+                # after we listed the contents of the bucket, but before we
+                # downloaded the object.
+                #
                 if not ('Error' in err.response and err.response['Error'].get('Code') == '404'):
                     raise err
             except StopIteration:
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -1189,18 +1189,27 @@ def iter_bucket(
 
     with smart_open.concurrency.create_pool(processes=workers) as pool:
         result_iterator = pool.imap_unordered(download_key, key_iterator)
-        for key_no, (key, content) in enumerate(result_iterator):
-            if True or key_no % 1000 == 0:
-                logger.info(
-                    "yielding key #%i: %s, size %i (total %.1fMB)",
-                    key_no, key, len(content), total_size / 1024.0 ** 2
-                )
-            yield key, content
-            total_size += len(content)
-
-            if key_limit is not None and key_no + 1 >= key_limit:
-                # we were asked to output only a limited number of keys => we're done
+        key_no = 0
+        while True:
+            try:
+                (key, content) = result_iterator.__next__()
+                if True or key_no % 1000 == 0:
+                    logger.info(
+                        "yielding key #%i: %s, size %i (total %.1fMB)",
+                        key_no, key, len(content), total_size / 1024.0 ** 2
+                    )
+                yield key, content
+                total_size += len(content)
+                if key_limit is not None and key_no + 1 >= key_limit:
+                    # we were asked to output only a limited number of keys => we're done
+                    break
+            except botocore.exceptions.ClientError as err:
+                # ignore 404 not found errors
+                if not ('Error' in err.response and err.response['Error'].get('Code') == '404'):
+                    raise err
+            except StopIteration:
                 break
+            key_no += 1
     logger.info("processed %i keys, total size %i" % (key_no + 1, total_size))
 
 
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -681,6 +681,38 @@ class IterBucketTest(unittest.TestCase):
         results = list(smart_open.s3.iter_bucket(BUCKET_NAME))
         self.assertEqual(len(results), 10)
 
+    @pytest.mark.skipif(condition=sys.platform == 'win32', reason="does not run on windows")
+    @pytest.mark.xfail(
+        condition=sys.platform == 'darwin',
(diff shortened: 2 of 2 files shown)


smart_open/s3.py
```diff
@@ -1189,18 +1189,31 @@ def iter_bucket(
...
+                yield key, content
+                total_size += len(content)
+                if key_limit is not None and key_no + 1 >= key_limit:
+                    # we were asked to output only a limited number of keys => we're done
+                    break
+            except botocore.exceptions.ClientError as err:
+                if 'Error' in err.response and err.response['Error'].get('Code') == '404':
+                    logger.warning(
+                        "Encountered '404 Not Found' error for key #%i",
+                        key_no
```
ronkorving, line 1210: Would it perhaps make sense to log the key, rather than the key_no? If anybody does care to see the log, I'm not sure the number would be as helpful in debugging a situation.

An argument could perhaps be made that it's not actually worthy of being logged as a warning. An end-user could read this as something being wrong or broken, when this is completely valid behavior that is to be expected from time to time.

RachitSharma2001: That is a very good point. If a 404 error occurs during iter bucket, it is almost certainly not the users fault, but rather the fault of the internal list keys function. I agree that logging the error would just be confusing to the user. I've removed the warning in my latest commit. Let me know what you think.

smart_open/s3.py
```diff
@@ -1189,18 +1189,27 @@ def iter_bucket(
...
-            yield key, content
-            total_size += len(content)
-
-            if key_limit is not None and key_no + 1 >= key_limit:
-                # we were asked to output only a limited number of keys => we're done
+        key_no = 0
+        while True:
+            try:
+                (key, content) = result_iterator.__next__()
+                if True or key_no % 1000 == 0:
```
ronkorving, line 1196: Any need for the `True or` bit?

smart_open/s3.py
ronkorving: Makes sense to me 👍

smart_open/s3.py
RachitSharma2001: I was also wondering this while I was looking at the code. I am not sure why it is there, I can remove it if its confirmed that it is not needed.

ronkorving: Perhaps @mpenkov knows more about this.

mpenkov (approved):

mpenkov:
Thank you @RachitSharma2001 !

PR #737, opened 2022-11-09
Bring back compression_wrapper(filename) + use case-insensitive extension matching

### LINKED ISSUE #736: Make compression_wrapper accept filename again
The `smart_open.compression.compression_wrapper()` used to accept a `filename` parameter. This is necessary in cases where the input `file_obj` cannot be modified = monkey-patchong a `.name` attribute into it is undesirable.

This feature was removed in 930c897aab96821bd5fff9cb6b0a2f8ab922f51e – @mpenkov do you remember why?

I propose bringing it back, because the alternative is hacky: accessing the protected `compression._COMPRESSOR_REGISTRY` dict directly.

Relatedly, the extension-matching logic is case-sensitive, `.gz` != `.GZ` etc. I don't see much point to this – should be case-insensitive IMO.

piskvorky:
Fixes #736.

- [x] Add test(s), both for the "case-insensitive extension" and for `filename=X` itself.

### DIFF
--- a/smart_open/compression.py
+++ b/smart_open/compression.py
@@ -62,6 +62,7 @@ def register_compressor(ext, callback):
     """
     if not (ext and ext[0] == '.'):
         raise ValueError('ext must be a string starting with ., not %r' % ext)
+    ext = ext.lower()
     if ext in _COMPRESSOR_REGISTRY:
         logger.warning('overriding existing compression handler for %r', ext)
     _COMPRESSOR_REGISTRY[ext] = callback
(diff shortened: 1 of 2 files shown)


piskvorky:
Unclear to me why the CI failed.  I'll try simply re-running the jobs.

piskvorky:
TODO: add test(s), both for the "case-insensitive extension" and for `filename=X` itself.

mpenkov (approved):

PR #739, opened 2022-11-16
Add FTPS support (#33)

RachitSharma2001:
#### Title
Support for reading and writing files directly to/from ftps (ftp with tls)
#### Motivation

I have [previously](https://github.com/RaRe-Technologies/smart_open/pull/723) added support ftp servers (without tls encryption). Now, to add more complete ftp functionality, I have added support for ftp+tls servers. 

```python
from smart_open import open
with open('ftps://user:password@host:port/dir1/dir2/file') as fin:
     for line in fin:
          print(line)
```

#### Tests
I have edited the ftp integration tests to run an ftp and ftps server on different ports and then test reading, writing, and appending to each of them. (specific integration test [here](https://github.com/RachitSharma2001/smart_open/blob/ftps/integration-tests/test_ftp.py), pull up ftp and ftps servers [here](https://github.com/RachitSharma2001/smart_open/blob/ftps/ci_helpers/helpers.sh)).

### DIFF
--- a/ci_helpers/helpers.sh
+++ b/ci_helpers/helpers.sh
@@ -11,6 +11,7 @@ create_ftp_ftps_servers(){
   #
   # Must be run as root
   #
+  HOME_DIR=/home/user
   mkdir $HOME_DIR
   useradd -p $(echo $PASS | openssl passwd -1 -stdin) -d $HOME_DIR $USER
   chown $USER:$USER $HOME_DIR
--- a/ci_helpers/helpers.sh
+++ b/ci_helpers/helpers.sh
@@ -8,9 +8,12 @@ enable_moto_server(){
 }
 
 create_ftp_ftps_servers(){
-  sudo mkdir $HOME_DIR
-  sudo useradd -p $(echo $PASS | openssl passwd -1 -stdin) -d $HOME_DIR $USER
-  sudo chown $USER:$USER $HOME_DIR
+  #
+  # Must be run as root
+  #
+  mkdir $HOME_DIR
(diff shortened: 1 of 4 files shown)


ci_helpers/helpers.sh
```diff
@@ -7,17 +7,27 @@ enable_moto_server(){
...
+    --name my-ftp-ftps-servers \
+    --publish 20-21:20-21/tcp \
+    --publish 40000-40009:40000-40009/tcp \
+    --publish 90:90/tcp \
+    --env USER=user \
+    --env PASS=123 \
+    --env HOME_DIR=/home/user \
+    --env FTP_PORT=21 \
+    --env FTPS_PORT=90 \
+    rachitsharma2001/ftp_ftps
```
piskvorky, line 21: Running external docker images is scary, both from security and also maintenance perspective.

Any wya to launch the FTP(S) server directly from Python here?

piskvorky (commented):
Looks good but can you look into not requiring & running an external Docker image for the testing? Thanks.

ci_helpers/helpers.sh
RachitSharma2001: Thank you for letting me know. I have removed the docker dependency and instead directly set up the ftp and ftps servers on ci using the `vsftpd` linux tool. Let me know if this looks good or if there is anything else I should add or change.

piskvorky (approved):

piskvorky:
Thanks for the quick resolution @RachitSharma2001.

@mpenkov can you please review?

mpenkov:
Thank you for your work @RachitSharma2001 !

mpenkov (approved):

PR #749, opened 2022-12-24
Fix s3.iter_bucket failure when botocore_session passed in

### LINKED ISSUE #670: Using botocore_session parameter of s3.iter_bucket fails
### Problem description

I am attempting to list all files in an S3 bucket, as per the documentation I am providing `smart_open.s3.iter_bucket` with a pre-made `botocore_session`. I have tried turning concurrency off and received a different exception but to no avail.

#### Expected result: 
Works like it does when not providing the `botocore_session` parameter; the keys and contents of the files on S3 are iterated.

##### With `smart_open.concurrency._MULTIPROCESSING = True`
I receive the exception:
```
AttributeError: Can't pickle local object 'lazy_call.<locals>._handler'
```
##### With `smart_open.concurrency._MULTIPROCESSING = False`
 I receive the exception: 
```
RuntimeError: Cannot inject class attribute "upload_file", attribute already exists in class dict.
```

---
From my initial investigation it seems like the creation of many new `boto3` sessions / resources from the same `botocore` session in `smart_open.s3._download_key` is the issue here, which might be a limitation/issue with the boto3 library.

### Steps/code to reproduce the problem

```python
import botocore
from smart_open import s3, concurrency

bucket_name = '[A_REAL_BUCKET]'
concurrency._MULTIPROCESSING = False

my_session = botocore.session.Session()
bucket_contents = list(s3.iter_bucket(bucket_name=bucket_name,
                                      botocore_session=my_session,
                                      workers=1))
```

Running the above results in the below traceback:
```
/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/bin/python /home/tyto/.config/JetBrains/PyCharm2021.2/scratches/scratch.py
11:52:28.649 INFO:smart_open.concurrency: creating concurrent futures pool with 1 workers
11:52:28.662 INFO:botocore.credentials: Found credentials in shared credentials file: ~/.aws/credentials
  File "/home/tyto/.config/JetBrains/PyCharm2021.2/scratches/scratch.py", line 18, in <module>
    bucket_contents = list(s3.iter_bucket(bucket_name=bucket_name,
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/smart_open/s3.py", line 1192, in iter_bucket
    for key_no, (key, content) in enumerate(result_iterator):
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/smart_open/concurrency.py", line 58, in imap_unordered
    yield future.result()
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 438, in result
    return self.__get_result()
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 390, in __get_result
    raise self._exception
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/smart_open/s3.py", line 1246, in _download_key
    s3 = session.resource('s3')
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/boto3/session.py", line 396, in resource
    client = self.client(
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/boto3/session.py", line 270, in client
    return self._session.create_client(
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/session.py", line 841, in create_client
    client = client_creator.create_client(
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/client.py", line 84, in create_client
    cls = self._create_client_class(service_name, service_model)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/client.py", line 115, in _create_client_class
    self._event_emitter.emit(
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/hooks.py", line 357, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/hooks.py", line 228, in emit
    return self._emit(event_name, kwargs)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/boto3/utils.py", line 63, in _handler
    return getattr(module, function_name)(**kwargs)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/boto3/s3/inject.py", line 22, in inject_s3_transfer_methods
    utils.inject_attribute(class_attributes, 'upload_file', upload_file)
  File "/home/tyto/.cache/pypoetry/virtualenvs/s3-collector-combiner-UhjNKLw0-py3.9/lib/python3.9/site-packages/boto3/utils.py", line 70, in inject_attribute
    raise RuntimeError(
RuntimeError: Cannot inject class attribute "upload_file", attribute already exists in class dict.

Process finished with exit code 1

```

Note that with `_MULTIPROCESSING = True` a different traceback occurs. For now I'm putting that down to multiprocessing returning exceptions in another way but I'm happy to provide the output if required.

### Versions

```
Linux-5.13.0-7620-generic-x86_64-with-glibc2.33
Python 3.9.5 (default, May 11 2021, 08:20:37) 
[GCC 10.3.0]
smart_open 5.2.1
```

### Checklist

RachitSharma2001:
#### Motivation
Fixes #670 

As shown by @SootyOwl, when a user declares their own `botocore_session` object and passes it into `s3.iter_bucket`, one of two errors occur:
1. With **smart_open.concurrency._MULTIPROCESSING = True**:
   `AttributeError: Can't pickle local object 'lazy_call.<locals>._handler`
2. With **smart_open.concurrency._MULTIPROCESSING = False**:
   `RuntimeError: Cannot inject class attribute "upload_file", attribute already exists in class dict.`

As explained [here](https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror), the reason the first error occurs is that the multiprocessing module performs pickling on objects and requires those objects to be global, not local.

As explained in the original raised issue, the reason the second error occurs is that `_list_bucket` and `_download_key` both creates `boto3.session.Session` objects out of the passed in `botocore_session`, which is not allowed by the boto3 library.

The proposed changes address both issues by creating a global session object within `iter_bucket` that `_list_bucket` and `_download_key` can access.
 
#### Tests
All existing tests related to `iter_bucket` within `s3.py` pass. I also added two new tests: `test_iter_bucket_passed_in_session_multiprocessing_false` and `test_iter_bucket_passed_in_session_multiprocessing_true`. These test the two previously failing situations.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -1176,17 +1176,18 @@ def iter_bucket(
         pass
 
     total_size, key_no = 0, -1
+    global session
+    session = boto3.session.Session(**session_kwargs)
     key_iterator = _list_bucket(
         bucket_name,
         prefix=prefix,
         accept_key=accept_key,
-        **session_kwargs)
+        create_session=False)
     download_key = functools.partial(
         _download_key,
         bucket_name=bucket_name,
         retries=retries,
-        **session_kwargs)
-
+        create_session=False)
     with smart_open.concurrency.create_pool(processes=workers) as pool:
         result_iterator = pool.imap_unordered(download_key, key_iterator)
         key_no = 0
@@ -1221,9 +1222,12 @@ def _list_bucket(
         bucket_name,
         prefix='',
         accept_key=lambda k: True,
+        create_session=True,
         **session_kwargs):
-    session = boto3.session.Session(**session_kwargs)
-    client = session.client('s3')
+    if create_session:
+        client = boto3.session.Session(**session_kwargs).client('s3')
+    else:
+        client = session.client('s3')
     ctoken = None
 
     while True:
@@ -1248,15 +1252,17 @@ def _list_bucket(
             break
 
 
-def _download_key(key_name, bucket_name=None, retries=3, **session_kwargs):
+def _download_key(key_name, bucket_name=None, retries=3, create_session=True, **session_kwargs):
     if bucket_name is None:
         raise ValueError('bucket_name may not be None')
 
     #
     # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-or-multiprocessing-with-resources
     #
-    session = boto3.session.Session(**session_kwargs)
-    s3 = session.resource('s3')
+    if create_session:
+        s3 = boto3.session.Session(**session_kwargs).resource('s3')
+    else:
+        s3 = session.resource('s3')
     bucket = s3.Bucket(bucket_name)
 
     # Sometimes, https://github.com/boto/boto/issues/2409 can happen
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -686,6 +686,34 @@ class IterBucketTest(unittest.TestCase):
         condition=sys.platform == 'darwin',
         reason="MacOS uses spawn rather than fork for multiprocessing",
     )
+    def test_iter_bucket_passed_in_session_multiprocessing_true(self):
+        populate_bucket()
+        smart_open.concurrency._MULTIPROCESSING = True
+        my_session = botocore.session.Session()
+        results = list(smart_open.s3.iter_bucket(bucket_name=BUCKET_NAME,
+                                      botocore_session=my_session,
+                                      workers=1))
+        self.assertEqual(len(results), 10)
+
+    @pytest.mark.skipif(condition=sys.platform == 'win32', reason="does not run on windows")
+    @pytest.mark.xfail(
+        condition=sys.platform == 'darwin',
+        reason="MacOS uses spawn rather than fork for multiprocessing",
+    )
+    def test_iter_bucket_passed_in_session_multiprocessing_false(self):
+        populate_bucket()
+        smart_open.concurrency._MULTIPROCESSING = False
+        my_session = botocore.session.Session()
+        results = list(smart_open.s3.iter_bucket(bucket_name=BUCKET_NAME,
+                                      botocore_session=my_session,
+                                      workers=1))
+        self.assertEqual(len(results), 10)
+
+    @pytest.mark.skipif(condition=sys.platform == 'win32', reason="does not run on windows")
+    @pytest.mark.xfail(
+        condition=sys.platform == 'darwin',
+        reason="MacOS uses spawn rather than fork for multiprocessing",
+    )
     def test_iter_bucket_404(self):
         populate_bucket()
 


mpenkov:
Thank you for this carefully prepared PR.

I think you may be misunderstanding the post on StackOverflow. Multiprocessing doesn't require objects to be global. It only requires them to be pickle-able. For functions, this means they must be defined at model scope (I think this is what you refer to as "global"). For objects (e.g. an boto3 session), there is no such restriction.

With the above in mind, I think it's possible to fix this problem without using globals: just pass the session object instead of the session_kwargs.

mpenkov:
Thank you for this carefully prepared PR.

I think you may be misunderstanding the StackOverflow post. Multiprocessing does not require things to be global. It only requires them to be pickle-able. For functions, this means defined at module scope (I think this is what you refer to as global). For objects (like boto3 sessions) there is no such requirement.

Is there a reason why we're not passing the entire session object to iter_bucket? I think that would make it possible to avoid using globals here.

mpenkov:
> Is there a reason why we're not passing the entire session object to iter_bucket?

I remembered why we're doing this. The [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#multithreading-or-multiprocessing-with-sessions) advise to do so:

> It's recommended to create a new Session object for each thread or process

So it looks like their recommendation is part of the problem here. We could go against it (and pass the Session across the process boundary like I suggested above). There aren't really any good alternatives here, because passing the Client (instead of the Session) across a process boundary is forbidden:

> Multi-Processing: While clients are thread-safe, they cannot be shared across processes due to their networking implementation. Doing so may lead to incorrect response ordering when calling services.

I've re-read the original issue:

> As explained in the original raised issue, the reason the second error occurs is that _list_bucket and _download_key both creates boto3.session.Session objects out of the passed in botocore_session, which is not allowed by the boto3 library.

... and now I'm unsure if the use case is valid. If the botocore.Session cannot be safely passed across a subprocess boundary (effectively cloning it) and re-used, there isn't much we can do. Going against Boto3 recommendations and passing a boto3.Session across the subprocess boundary (whether as a global or otherwise) also sounds like a bad idea to me.

RachitSharma2001:
@mpenkov That is a good point. So from what I am thinking, either we don't allow the user to pass in a custom boto3 Session object at all (and throw an error or warning if they do), or (if it is possible to reuse that session object) allow them to pass it in only if `smart_open.concurrency._MULTIPROCESSING = False`, and throw an error if the session object is passed but `smart_open.concurrency._MULTIPROCESSING = True`. For either case, I could update this pr to make the necessary changes if needed.

mpenkov:
I think a warning is sufficient. Also, in this particular case, we don't care about *thread* safety - we care about *process* safety.

The other tricky thing to consider is how this process issue interplays with the MULTIPROCESSING flag. If that flag is set to False, then the game changes significantly, and using the single session is OK. So we may have to split the implementation into two:

1. multiprocessing implementation, process-safety caveats apply
2. single-process implementation, process-safety caveats do not apply

It's a bit of a pain...  I'm not sure what the best way to handle this is.

RachitSharma2001:
I have an idea of how to handle this but it does seem a bit messy:
1. Give `_download_key` and `_list_bucket` two additional arguments: `boto3_session` which by default is **None**, and `**session_kwargs`.
2. Then, in `iter_bucket`, check if `botocore_session` exists in `session_kwargs`. If so, create a `boto3 session` object out of it, and pass it in to `_download_key` and `_list_bucket` (thus avoiding the 'upload_file' error that occurs when multiprocessing is **False**)
3. But if no `botocore_session` object is passed in, then we can just call `_download_key` and `_list_bucket` just with `**session_kwargs`, and they will both individually create their own boto3 session objects (thus avoiding any pickling error that may occur when multiprocessing is true yet no botocore session object is passed in)
4. and if the user passes in `botocore_session` to `iter_bucket` but multiprocessing is true, we give a warning 

This would help avoid the 'upload file' error that occurs when multiprocessing is False while also avoiding any unnecessary errors when multiprocessing is True (except, of course, when `botocore_session` is passed in). 

We could also just add a botocore_session argument to iter_bucket itself. Let me know if this idea sounds good or if there could be a better way to do this.

mpenkov:
Yeah, that _does_ sound a bit messy.  There _should_ be a better way. Give me a little while to think about it.

mpenkov:
I'm going to close this, because the use case described in the original issue has a work-around (disable multiprocessing).

Thank you for your work. In this particular case, supporting the original use case does not justify increasing the complexity of the iter_bucket function even further. I had a think about what the best solution here would be, but couldn't come up with anything.

If I end up refactoring iter_bucket in the future, I'll take this work into consideration.

PR #752, opened 2023-01-04
Fix #684 Abort S3 MultipartUpload if exception is raised

### LINKED ISSUE #684: S3 Multipart uploads comitted on exception when using compression via context manager
#### Problem description

- We're trying to upload gzipped objects to S3, using multipart upload
- We're using a context manager to open a gzipped output file in S3, and writing to the file. When the writing throws an exception we expect the multipart upload to be terminated
- The multipart upload is commited

This is because when using a context manager around a compressed file, the `MultipartWriter` `__exit__` method will not be called on https://github.com/RaRe-Technologies/smart_open/blob/59d3a6079b523c030c78a3935622cf94405ce052/smart_open/s3.py#L938-L942

We've seen two resulting behaviours from this issue:

1. Empty or incomplete multipart uploads are written, this seems to be as expected when `__exit__` is not called correctly
2. Corrupted gzipped files being written. We're less sure how this occured

```python
import smart_open

with smart_open.open(
    s3.s3_uri(output_bucket, "test-file.gz"),
    "wb",
) as fout:
    for data in ["valid-data", Exception("An exception")]:
		if isinstance(data, Exception)
			raise data
    	fout.write((doc + "\n").encode("utf-8"))
```

#### Versions

```python
>>> import platform, sys, smart_open
m.platform())
print("Python", sys.version)
print("smart_open", smart_open.__version__)>>> print(platform.platform())
Linux-5.10.60.1-microsoft-standard-WSL2-x86_64-with-debian-bullseye-sid
Python 3.7.11 (default, Jan 13 2022, 14:48:06)
[GCC 9.3.0]
smart_open 5.2.1
```

lociko:
#### Motivation

- Fixes #684
AWS Supports multipart upload of a file. I big file split by chunks uploaded one by one and on S3 concatenated again.
In our case, if an exception is raised while processing one of the parts we have to abort uploading to avoid corrupted file creation.

#### Tests

`test_write_gz_with_error`

### DIFF
--- a/smart_open/compression.py
+++ b/smart_open/compression.py
@@ -8,6 +8,9 @@
 """Implements the compression layer of the ``smart_open`` library."""
 import logging
 import os.path
+import sys
+
+from smart_open.s3 import MultipartWriter
 
 logger = logging.getLogger(__name__)
 
@@ -83,9 +86,15 @@ def tweak_close(outer, inner):
         try:
             outer_close()
         finally:
-            if inner:
-                inner, fp = None, inner
-                fp.close()
+            ex, _, _ = sys.exc_info()
+
+            # https://github.com/RaRe-Technologies/smart_open/issues/684
+            # If exception is raised multipart writer should be terminated to abort file uploading to S3
+            # If not incomplete file will be saved in bucket
+            if type(inner) == MultipartWriter and ex:
+                inner.terminate()
+            elif inner:
+                inner.close()
 
     outer.close = close_both
 
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -554,7 +554,6 @@ class MultipartWriterTest(unittest.TestCase):
 
             assert actual == contents
 
-
     def test_write_gz_with_error(self):
         """Does s3 multipart chunking work correctly if exception is raised for compression file?"""
 
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -555,6 +555,34 @@ class MultipartWriterTest(unittest.TestCase):
             assert actual == contents
 
 
+    def test_write_gz_with_error(self):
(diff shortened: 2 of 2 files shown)


smart_open/compression.py
```diff
@@ -83,9 +86,15 @@ def close_both(*args):
...
         finally:
-            if inner:
-                inner, fp = None, inner
-                fp.close()
+            ex, _, _ = sys.exc_info()
+
+            # https://github.com/RaRe-Technologies/smart_open/issues/684
+            # If exception is raised multipart writer should be terminated to abort file uploading to S3
+            # If not incomplete file will be saved in bucket
+            if type(inner) == MultipartWriter and ex:
```
ddelange, line 94: Hi 👋 can this be generalized such that it doesn't just handle S3, but also the other adapters like Azure?

Azure exception handling ref https://github.com/RaRe-Technologies/smart_open/pull/783

ddelange: e.g. by refactoring `tweak_close` to become `tweak_exit`, that patches `.__exit__()` instead of `.close()`?

ddelange: opened https://github.com/RaRe-Technologies/smart_open/pull/786

leeprevost:
What do I need to do to install this? I'm running into situation where larger transfers fail with this error:

`botocore.exceptions.ClientError('An error occurred (EntityTooLarge) when calling the UploadPart operation: Your proposed upload exceeds the maximum allowed size')`

ddelange:
- this PR was superceded by https://github.com/piskvorky/smart_open/pull/786
- large S3 uploads was fixed in https://github.com/piskvorky/smart_open/pull/796
- both are unreleased so for now you can `pip install 'smart_open @ https://github.com/piskvorky/smart_open/archive/02b8be2577821c1a6d30e6321c07aa6dbc690a2f.zip'`

mpenkov:
Closing, thank you

PR #753, opened 2023-01-05
Make s3u protocol have the intended non-SSL behaviour

fosslinux:
Currently https is hardcoded in when s3* protocol is used. However in the documentation it states s3u is the non-SSL version, but this appears unimplemented.

When s3u is used, use http rather than https.

#### Tests

Could a maintainer please advise how (if?) a test should be written for this change.
I don't believe vanilla AWS S3 supports unsecured http.

I am not able to run pytest on this PR as I don't have access to an AWS S3 bucket (I am making this change so I can use smart_open with my minio installation).

#### Checklist

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -88,6 +88,10 @@ def parse_uri(uri_as_string):
     #
     split_uri = smart_open.utils.safe_urlsplit(uri_as_string)
     assert split_uri.scheme in SCHEMES
+    if split_uri.scheme == "s3u":
+        ssl_mode = ""
+    else:
+        ssl_mode = "s"
 
     port = DEFAULT_PORT
     host = DEFAULT_HOST
@@ -130,6 +134,7 @@ def parse_uri(uri_as_string):
         ordinary_calling_format=ordinary_calling_format,
         access_id=access_id,
         access_secret=access_secret,
+        ssl_mode=ssl_mode,
     )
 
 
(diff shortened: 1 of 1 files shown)


smart_open/s3.py
```diff
@@ -130,6 +134,7 @@ def parse_uri(uri_as_string):
         ordinary_calling_format=ordinary_calling_format,
         access_id=access_id,
         access_secret=access_secret,
+        ssl_mode=ssl_mode,
```
mpenkov, line 137: You don't need to pass this as a separate parameter, because it's already included in the URL scheme.

fosslinux: How do you mean? Would you prefer the essence of the `ssl_mode` logic to be moved down into `_consolidate_params`?

mpenkov: Yes, exactly.

smart_open/s3.py
```diff
@@ -184,7 +189,7 @@ def inject(**kwargs):
         )
         uri.update(host=None)
     elif uri['host'] != DEFAULT_HOST:
-        inject(endpoint_url='https://%(host)s:%(port)d' % uri)
+        inject(endpoint_url='http%(ssl_mode)s://%(host)s:%(port)d' % uri)
```
mpenkov, line 192:
```suggestion
        if uri['scheme'] == 's3u':
            scheme = 'https'
        else:
            scheme = 'http'
        inject(endpoint_url=scheme + '://%(host)s:%(port)d' % uri)
```

fosslinux: Understood.

mpenkov:
Thank you for your work!

PR #756, opened 2023-01-18
Add required import for example to work

jensenbox:
If a person were to simply copy this code block it would use the built in `open` and would not work. Adding in the correct import makes this block a bit easier for a simple copy paste.


piskvorky (approved):

mpenkov (approved):

mpenkov:
Thank you @jensenbox  !

PR #767, opened 2023-04-06
Avoid overfilling buffer when reading from Azure

ronreiter:
#### Title

when read size > chunk size, return read size and not chunk size.

```
bugfix: read(MAX_READ_SIZE) returns read(CHUNK_SIZE) when read size > chunk size
```

#### Motivation

The read function expects to return the maximum size it can read from a stream and not the chunk size. This causes serious bugs when reading with a parameter such as x.read(MAX_READ_SIZE), which does not expect you to re-read the buffer until you get to MAX_READ_SIZE.

This bug does not happen when the read size is not specified.

If you're fixing a bug, link to the issue number like so:

```
- Fixes #{issue_number}
```

If you're adding a new feature, then consider opening a ticket and discussing it with the maintainers before you actually do the hard work.

#### Tests

2. Implement your bug fix.

    pytest smart_open

#### Work in progress

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/azure.py
+++ b/smart_open/azure.py
@@ -306,7 +306,7 @@ class Reader(io.BufferedIOBase):
         if self._position == self._size:
             return self._read_from_buffer()
 
-        self._fill_buffer()
+        self._fill_buffer(size)
         return self._read_from_buffer(size)
 
     def read1(self, size=-1):


mpenkov (approved):

mpenkov:
Thank you @ronreiter !

PR #786, opened 2023-10-02
Propagate __exit__ call to underlying filestream

### LINKED ISSUE #684: S3 Multipart uploads comitted on exception when using compression via context manager

ddelange:
#### Title

Propagate `__exit__` call to underlying filestream. 

No more need for `tweak_close` function: the double wrapping will now cause three calls to `__exit__`, which in turn will call `close`, first on the outermost filestream, working inward.

If a third party compression handler calls `close` on the inner filestream, the inner `__exit__` call will do a second call to `close` on the inner filestream, which will be ignored.

#### Motivation

Currently, the compression mechanic has a side-effect that `__exit__` is no longer called on the original filestream. Consequently, eg boto3 abort multipart upload is not called, causing lingering blobs that will be billed. 

- Supercedes https://github.com/RaRe-Technologies/smart_open/pull/752
- Fixes #684

#### Tests

✅ 

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -911,12 +911,14 @@ multipart upload may fail")
         """Cancel the underlying multipart upload."""
         if self._upload_id is None:
             return
+        logger.debug('%s: terminating multipart upload', self)
         self._client.abort_multipart_upload(
             Bucket=self._bucket,
             Key=self._key,
             UploadId=self._upload_id,
         )
         self._upload_id = None
+        logger.debug('%s: terminated multipart upload', self)
 
     def to_boto3(self, resource):
         """Create an **independent** `boto3.s3.Object` instance that points to
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -909,7 +909,8 @@ multipart upload may fail")
 
     def terminate(self):
         """Cancel the underlying multipart upload."""
-        assert self._upload_id, "no multipart upload in progress"
+        if self._upload_id is None:
+            return
         self._client.abort_multipart_upload(
             Bucket=self._bucket,
             Key=self._key,
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -826,6 +826,7 @@ multipart upload may fail")
         if self._buf.tell():
             self._upload_next_part()
 
+        logger.debug('%s: completing multipart upload', self)
         if self._total_bytes and self._upload_id:
             partial = functools.partial(
                 self._client.complete_multipart_upload,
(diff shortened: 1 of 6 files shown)


smart_open/s3.py
```diff
@@ -909,7 +909,8 @@ def write(self, b):
 
     def terminate(self):
         """Cancel the underlying multipart upload."""
-        assert self._upload_id, "no multipart upload in progress"
+        if self._upload_id is None:
+            return
```
ddelange, line 913: terminate() called before a multipart upload was initiated:

```
self = <smart_open.tests.test_smart_open.S3OpenTest testMethod=test_write_bad_encoding_strict>

    @moto.mock_s3
    def test_write_bad_encoding_strict(self):
        """Should open the file for writing with the correct encoding."""
        s3 = _resource('s3')
        s3.create_bucket(Bucket='bucket')
        key = "s3://bucket/key.txt"
        text = u'欲しい気持ちが成長しすぎて'
    
        with self.assertRaises(UnicodeEncodeError):
            with smart_open.open(key, 'w', encoding='koi8-r', errors='strict') as fout:
>               fout.write(text)
```

mpenkov (approved):
LGTM, thank you

PR #789, opened 2023-11-03
Fix `KeyError: 'ContentRange'` when received full content from S3

### LINKED ISSUE #686: Missing `ContentRange` in the response from AWS
#### Problem description

We run the following code to stream avro files directly from s3. As of a few weeks ago, `smart-open` complains that the response from AWS is missing the `ContentRange` key ([here](https://github.com/RaRe-Technologies/smart_open/blob/develop/smart_open/s3.py#L464)).

```
from fastavro import reader
from smart_open import open as smart_open
import boto3

session = boto3.session.Session()
client = session.client('s3')
transport_params = dict(client=client, session=session)

record_meta = dict(
    file_name="s3://bucket-name/filename.avro",
    offset=4500,
)

with smart_open(record_meta["file_name"], "rb", transport_params=transport_params) as file:
    avro_reader = reader(file)
    next(avro_reader)
    file.seek(record_meta['offset'])
    record = next(avro_reader)
```

But according to [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html), the response body should contain a range. Is this an issue that needs to be raised on the AWS side, or should something be fixed on this end?

#### Versions

```
Darwin-20.6.0-x86_64-i386-64bit
Python 3.7.9 (v3.7.9:13c94747c7, Aug 15 2020, 01:31:08) 
[Clang 6.0 (clang-600.0.57)]
smart_open 4.2.0
```

messense:
#### Motivation

https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#partial_request_responses

> If range requests are not supported, an [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) OK status is sent back and the entire response body is transmitted.

In case there is a proxy between s3 and end user, range requests may not be supported.

Fixes #686

#### Checklist

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -486,9 +486,17 @@ class _SeekableRawReader(object):
                 self,
                 response['ResponseMetadata']['RetryAttempts'],
             )
-            _, start, stop, length = smart_open.utils.parse_content_range(response['ContentRange'])
+            # range request may not always return partial content, see
+            # https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#partial_request_responses
+            status_code = response['ResponseMetadata']['HTTPStatusCode']
+            if status_code == 206:
+                # partial content
+                _, start, stop, length = smart_open.utils.parse_content_range(response['ContentRange'])
+                self._position = start
+            elif status_code == 200:
+                # full content
+                length = response["ContentLength"]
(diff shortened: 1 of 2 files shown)


mpenkov:
Thank you!

PR #790, opened 2023-11-04
Add support for SSH connection via aliases from `~/.ssh/config`

wbeardall:
#### Motivation

For commonly-used connections, machines with pre-existing SSH configurations, and general convenience, we find it useful to be able to infer (or interpolate, as you please) a full connection configuration from the host alias in the standard SSH config file.

We suggest that configurations as specified as such should only be used as default values, and if alternatives username and port are provided as part of the file URI, these should override pre-existing values. For example, `ssh://new-user@host-alias:<new-port>/...` should override the pre-defined port and username in the SSH configuration file. 

#### Implementation Details

As currently implemented, the connection configuration interpolation supports the following configuration options, which are passed to `paramiko.client.SSHClient.connect`:

- hostname
- port
- username
- key_filename
- timeout
- compress
- gss_auth
- gss_kex
- gss_deleg_creds
- gss_trust_dns

As an example, the following configuration would be fully and properly utilised when called with `smart_open.open("ssh://another-host/...")`

```
Host another-host
  HostName another-host-domain.com
  User another-user
  Port 2345
  IdentityFile /path/to/key/file
  ConnectTimeout 20
  Compression yes
  GSSAPIAuthentication no
  GSSAPIKeyExchange no
  GSSAPIDelegateCredentials no
  GSSAPITrustDns no
```

#### Tests

Additional unit tests have been added in `smart_open/tests/test_ssh.py`, following the existing test pattern.

#### Checklist

### DIFF
--- a/smart_open/ssh.py
+++ b/smart_open/ssh.py
@@ -23,6 +23,7 @@ Similarly, from a command line::
 """
 
 import getpass
+import os
 import logging
 import urllib.parse
 
@@ -52,9 +53,22 @@ URI_EXAMPLES = (
     'sftp://username@host/path/file',
 )
 
+#
+# Global storage for SSH config files.
+#
+_SSH_CONFIG_FILES = [os.path.expanduser("~/.ssh/config")]
+
 
 def _unquote(text):
     return text and urllib.parse.unquote(text)
+
+
+def _str2bool(string):
+    if string == "no":
+        return False
+    if string == "yes":
+        return True
+    raise ValueError(f"Expected 'yes' / 'no', got {string}.")
 
 
 def parse_uri(uri_as_string):
@@ -65,7 +79,7 @@ def parse_uri(uri_as_string):
         uri_path=_unquote(split_uri.path),
         user=_unquote(split_uri.username),
         host=split_uri.hostname,
-        port=int(split_uri.port or DEFAULT_PORT),
+        port=int(split_uri.port) if split_uri.port else None,
         password=_unquote(split_uri.password),
     )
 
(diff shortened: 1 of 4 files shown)


smart_open/ssh.py
```diff
@@ -65,7 +79,7 @@ def parse_uri(uri_as_string):
         uri_path=_unquote(split_uri.path),
         user=_unquote(split_uri.username),
         host=split_uri.hostname,
-        port=int(split_uri.port or DEFAULT_PORT),
+        port=int(split_uri.port) if split_uri.port else None,
```
mpenkov, line 82: Why are we setting port to None instead of the default here?

smart_open/ssh.py
```diff
@@ -90,7 +104,74 @@ def _connect_ssh(hostname, username, port, password, transport_params):
...
+    # - GSS configuration
+    for config_filename in _SSH_CONFIG_FILES:
+        if os.path.exists(config_filename):
+            try:
+                cfg = paramiko.SSHConfig.from_path(config_filename)
+            except PermissionError:
+                continue
+            if host in cfg.get_hostnames():
+                cfg = cfg.lookup(host)
+                host = cfg["hostname"]
```
mpenkov, line 139: Why are you overwriting the host variable, and not using it further down below?

mpenkov: Never mind, I see what's happening here.

mpenkov (approved):
Thank you!

smart_open/ssh.py
wbeardall: If I recall correctly, this is to ensure that non-default ports from configuration files are loaded correctly. We think it best that the parse_uri function should not inject additional information, such as a default port, to a URI, as this is the job of the configuration parser. This is important for cases where a non-default port is specified in the config file for a connection, but not in the URI as-provided. If parse_uri injects the default here, then the non-default port from the config will be ignored, which is not what we want. We only want to override ports specified in the config if explicitly provided by the user as part of the URI.

PR #791, opened 2023-11-24
Fix __str__ method in SinglepartWriter

### LINKED ISSUE #671: S3 SinglePartWriter doesn't work
#### Problem description
When you try to write a file to s3 using a single part upload, `SinglePartWriter.close()` throws: ` AttributeError: 'SinglepartWriter' object has no attribute '_object'`. The issue is in line 1012 of `smart_open.s3`:

```
logger.debug("%s: direct upload finished", self)
```

The problem is `SinglePartWriter.__str__()` has a reference to `self._object.bucket_name` and `self._object.key`. I'm assuming those are supposed to be `self._bucket` and `self._key` like everywhere else?

Just a simple write to s3 where `multipart_upload` transport kwarg is set to `False`.

#### Versions
Linux-4.19.104-microsoft-standard-x86_64-with-glibc2.31
python 3.9.5 (default, Sep 14 2021, 17:25:17)
[GCC 9.3.0]
smart_open 5.2.1

ThosRTanner:
#### Title

Fix __str__ method in SinglepartWriter

#### Motivation

- Fixes #671 

The bug here means that whenever you open an s3 connection and switch multipart writing off because the file is small, although the write completes successfully, you get an exception thrown when the code logs a debug message saying it has finished. This makes it look like your transfer has failed.

#### Tests

Added test

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -675,7 +675,7 @@ class SinglepartWriterTest(unittest.TestCase):
     def test_str(self):
         """Check the stringifying works!"""
         fout = smart_open.s3.open(BUCKET_NAME, 'key', 'wb', multipart_upload=False)
-        fout.write(text)
+        fout.write(b"text")
         fout.flush()
         fout.close()
         assert str(fout) != ''
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -672,6 +672,14 @@ class SinglepartWriterTest(unittest.TestCase):
 
             assert actual == contents
 
+    def test_str(self):
+        """Check the stringifying works!"""
+        fout = smart_open.s3.open(BUCKET_NAME, 'key', 'wb', multipart_upload=False)
+        fout.write(text)
+        fout.flush()
+        fout.close()
+        assert str(fout) != ''
+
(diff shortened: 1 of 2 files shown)


cadnce:
Nice. LGTM!

smart_open/tests/test_s3.py
```diff
@@ -672,6 +672,14 @@ def test_writebuffer(self):
 
             assert actual == contents
 
+    def test_str(self):
+        """Check the stringifying works!"""
+        fout = smart_open.s3.open(BUCKET_NAME, 'key', 'wb', multipart_upload=False)
+        fout.write(b"text")
+        fout.flush()
+        fout.close()
+        assert str(fout) != ''
```
mpenkov, line 681: Why are we checking against the empty string here? Why not just check against what the expected value should be?

mpenkov: Also, is writing, flushing, closing explicitly necessary? If not, then it shouldn't be part of the test, and a context manager will be enough.

mpenkov:
@ThosRTanner Could you please enable maintainer edits for this PR?

mpenkov:
OK, thank you!

PR #793, opened 2023-12-05
Secure the connection using SSL when connecting to the FTPS server

wammaster:
#### Title

Secure the connection using SSL when connecting to the FTPS server

#### Motivation

The connection to the FTPS server was not secure enough because, by default, the FTP library doesn't use SSL certificates. This PR fixes it by forcing the use of the SSL certificate if we connect to a secure server.

#### Tests

No tests to add or update.

#### Checklist

#### Workflow

### DIFF
--- a/smart_open/ftp.py
+++ b/smart_open/ftp.py
@@ -10,6 +10,7 @@
 """
 
 import logging
+import ssl
 import urllib.parse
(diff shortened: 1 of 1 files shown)


mpenkov:
Thank you!

PR #825, opened 2024-06-17
make http handler take an optional requests.Session

arondaniel:
#### Title
Makes http handler take an optional requests.Session object

#### Motivation

```
- Fixes piskvorky/smart_open#792 
```
Allows passing in a custom requests.Session object via smart_open.open transport_params. The Session object can support OAUTH2 authentication.

#### Tests

Added test for setting session attribute.

### DIFF
--- a/smart_open/http.py
+++ b/smart_open/http.py
@@ -50,7 +50,7 @@ def open_uri(uri, mode, transport_params):
 
 
 def open(uri, mode, kerberos=False, user=None, password=None, cert=None,
-         headers=None, timeout=None, buffer_size=DEFAULT_BUFFER_SIZE):
+         headers=None, timeout=None, session=None, buffer_size=DEFAULT_BUFFER_SIZE):
     """Implement streamed reader from a web site.
 
     Supports Kerberos and Basic HTTP authentication.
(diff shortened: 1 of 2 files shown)


ddelange:
@arondaniel can you pull latest develop? it contains some changes on CI that touch rewuests/urllib3

smart_open/http.py
```diff
@@ -227,6 +240,8 @@ def __init__(self, url, mode='r', buffer_size=DEFAULT_BUFFER_SIZE,
         """
         self.url = url
 
+        self.session = session
```
ddelange, line 243: Like this, you don't need the new if statement below, as you can always call upon `self.session`:
```suggestion
        self.session = session or requests
```

smart_open/tests/test_http.py
```diff
@@ -161,6 +161,14 @@ def test_timeout_attribute(self):
         assert hasattr(reader, 'timeout')
         assert reader.timeout == timeout
 
+    @responses.activate
+    def test_session_attribute(self):
+        session = requests.Session()
+        responses.add_callback(responses.GET, URL, callback=request_callback)
+        reader = smart_open.open(URL, "rb", transport_params={'session': session})
+        assert hasattr(reader, 'session')
+        assert reader.session == session
```
ddelange, line 170: can you also call `reader.read()` so the code you touched is covered when a session is passed

mpenkov:
Thank you @arondaniel and @ddelange !

PR #838, opened 2024-09-24
Not able to upload a compressed S3 file using close()

### LINKED ISSUE #837: Not able to upload a compressed S3 file using close()/without context manager
#### Problem description

- What are you trying to achieve?

We are using smart_open to compress and upload files to s3 without a context manager:
e.g.
```python
hook = S3Hook()
s3_params = {
    'client': hook.get_session().client('s3'),
}
f = smart_open.open('s3://bucket/file.gz', 'w', transport_params=s3_params)
f.write('Some content')
f.close()
```

- What is the expected result?
The expected result (and the result previous version 7.0.0) is to have the `MultipartWriter.close()` been called to perform a S3 complete multipart upload and see the file uploaded into s3

- What are you seeing instead?
The file is not uploaded and `MultipartWriter.close()` is not called.

Adding the following tests to test_s3.py (`class MultipartWriterTest`) shows our problem:
```python
    def test_write_gz_using_context_manager(self):
        """Does s3 multipart upload create a compressed file using context manager?"""
        contents = b'get ready for a surprise'
        with smart_open.open(
                f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz',
                mode="wb",
                transport_params={
                    "multipart_upload": True,
                    "min_part_size": 10,
                }
        ) as fout:
            fout.write(contents)

        with smart_open.open(f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz', 'rb') as fin:
            actual = fin.read()

        assert actual == contents

    def test_write_gz_not_using_context_manager(self):
        """Does s3 multipart upload create a compressed file not using context manager but close()?"""
        contents = b'get ready for a surprise'
        fout = smart_open.open(
            f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz',
            mode="wb",
            transport_params={
                "multipart_upload": True,
                "min_part_size": 10,
            }
        )
        fout.write(contents)
        fout.close()

        with smart_open.open(f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz', 'rb') as fin:
            actual = fin.read()

        assert actual == contents
```
#### Versions
This looks to be affecting from v7.0.0 to current version

Please provide the output of:
```
Linux-6.5.7-arch1-1-x86_64-with-glibc2.38
Python 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801]
smart_open 7.0.4
```
Probably related to this PR https://github.com/piskvorky/smart_open/pull/786 here:

https://github.com/piskvorky/smart_open/pull/786/files#diff-ba9d790d5607c344a97e86f15b3706586deb24d52d52cdf70b9776e3f3412b69L103

#### Recommendation

Probably closing both, outer and inner in `FileLikeProxy` will help:
```python

class FileLikeProxy(wrapt.ObjectProxy):
    __inner = ...  # initialized before wrapt disallows __setattr__ on certain objects
...
    def __next__(self):
        return self.__wrapped__.__next__()

    def close(self):
        try:
            self.__wrapped__.close()
        finally:
            self.__inner.close()


```

I can open a PR if that helps.

Thanks for reading!!

#### Checklist

- [ x ] Described the problem clearly
- [ x ] Provided a minimal reproducible example, including any required data
- [ x ] Provided the version numbers of the relevant software

jbarragan-bridge:
#### Motivation

We are using smart_open to compress and upload files to s3 without a context manager but since version 7.0.0 the following scenario is NOT longer working because we are using `close()` call without a context manager:
e.g.

```python
hook = S3Hook()
s3_params = {
    'client': hook.get_session().client('s3'),
}
f = smart_open.open('s3://bucket/file.gz', 'w', transport_params=s3_params)
f.write('Some content')
f.close()
```

The problem is that the close() call is only hitting the outer stream and NOT the inner stream. More information can be found in the [issue](https://github.com/piskvorky/smart_open/issues/837) information.

```
- Fixes #837
```

#### Tests

2 unit tests were added to shows how this change address the compressed file upload during the close() call.
#### Checklist

- [ x ] Picked a concise, informative and complete title
- [ x ] Clearly explained the motivation behind the PR
- [ x ] Linked to any existing issues that your PR will be solving
- [ x ] Included tests for any new functionality
- [ x ] Checked that all unit tests pass

### DIFF
--- a/smart_open/utils.py
+++ b/smart_open/utils.py
@@ -226,7 +226,7 @@ class FileLikeProxy(wrapt.ObjectProxy):
 
     def close(self):
         try:
-            self.__wrapped__.close()
+            return self.__wrapped__.close()
         finally:
             if self.__inner != self.__wrapped__:  # Don't close again if inner and wrapped are the same
                 self.__inner.close()
--- a/smart_open/utils.py
+++ b/smart_open/utils.py
@@ -223,3 +223,10 @@ class FileLikeProxy(wrapt.ObjectProxy):
 
     def __next__(self):
         return self.__wrapped__.__next__()
+
+    def close(self):
+        try:
+            self.__wrapped__.close()
+        finally:
+            if self.__inner != self.__wrapped__:  # Don't close again if inner and wrapped are the same
+                self.__inner.close()
--- a/smart_open/tests/test_s3.py
+++ b/smart_open/tests/test_s3.py
@@ -634,6 +634,43 @@ class MultipartWriterTest(unittest.TestCase):
 
             assert actual == contents
 
+    def test_write_gz_using_context_manager(self):
+        """Does s3 multipart upload create a compressed file using context manager?"""
+        contents = b'get ready for a surprise'
+        with smart_open.open(
+                f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz',
+                mode="wb",
+                transport_params={
+                    "multipart_upload": True,
+                    "min_part_size": 10,
+                }
+        ) as fout:
+            fout.write(contents)
+
+        with smart_open.open(f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz', 'rb') as fin:
+            actual = fin.read()
+
+        assert actual == contents
+
+    def test_write_gz_not_using_context_manager(self):
+        """Does s3 multipart upload create a compressed file not using context manager but close()?"""
+        contents = b'get ready for a surprise'
+        fout = smart_open.open(
+            f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz',
+            mode="wb",
+            transport_params={
+                "multipart_upload": True,
+                "min_part_size": 10,
+            }
+        )
+        fout.write(contents)
+        fout.close()
+
+        with smart_open.open(f's3://{BUCKET_NAME}/{WRITE_KEY_NAME}.gz', 'rb') as fin:
+            actual = fin.read()
+
+        assert actual == contents
+
     def test_write_gz_with_error(self):
         """Does s3 multipart upload abort for a failed compressed file upload?"""
         with self.assertRaises(ValueError):


smart_open/utils.py
```diff
@@ -223,3 +223,10 @@ def __exit__(self, *args, **kwargs):
 
     def __next__(self):
         return self.__wrapped__.__next__()
+
+    def close(self):
+        try:
+            self.__wrapped__.close()
+        finally:
+            if self.__inner != self.__wrapped__:  # Don't close again if inner and wrapped are the same
```
ddelange, line 231: I don't think there is a scenario where they are the same, plus calling close a second time should be a no-op. I think we can do without this conditional (unless there are tests failing without this conditional?)

smart_open/utils.py
```diff
@@ -223,3 +223,10 @@ def __exit__(self, *args, **kwargs):
 
     def __next__(self):
         return self.__wrapped__.__next__()
+
+    def close(self):
+        try:
+            self.__wrapped__.close()
```
ddelange, line 229:
```suggestion
            return super().close()
```
analogous to `__exit__` above (maybe some wrapped classes might return a boolean or something)

smart_open/utils.py
jbarragan-bridge: Thanks for checking the code @ddelange . This following test fails if I remove this conditional:

https://github.com/jbarragan-bridge/smart_open/blob/f435238348ce6ab66bc2cc8a3ab13a5a4595c145/smart_open/tests/test_smart_open.py#L1409-L1445 :
```python 
assert len(responses.calls) == 4  # it returns 6 without the conditional
```

I don't know much about `hdfs`; I saw in debug mode that this scenario creates a `FileLikeProxy` instance where the `wrapper` and `inner` are the same; duplicating the `close()` call.

Should adjust the unit test instead?

smart_open/utils.py
jbarragan-bridge: Hi @ddelange ,

I am trying the `super().close()` call but the unit test returns an error:
```python
    def close(self):
        try:
            # return self.__wrapped__.close()
>           return super().close()
E           AttributeError: 'super' object has no attribute 'close'

../utils.py:230: AttributeError
```

I noticed that you have `__wrapper__.next()` in the class method above. Probably because `wrapt.ObjectProxy` does not implement it; similar scenario with the class method `close()`).

I am implementing the return part in case `close()` return something in different implementation. Thanks!!

ddelange: alright, thanks for checking!

smart_open/utils.py
ddelange: good to know, something to look into but not related to this PR. let's leave it as is :)

ddelange (approved):
@mpenkov an important fix, can you cut a bugfix release with this one included?

jbarragan-bridge:
> @mpenkov an important fix, can you cut a bugfix release with this one included?

Thanks @mpenkov . I was out for a couple of days. Any updates on this change? Do you need me to take a different approach? Thank you!!

mpenkov (approved):

mpenkov:
Thank you @jbarragan-bridge . I will make a release in the next day or so.

mpenkov:
https://pypi.org/project/smart-open/7.0.5/

jbarragan-bridge:
> https://pypi.org/project/smart-open/7.0.5/

Thank you so much @mpenkov and @ddelange!! We are moving to version 7.0.5 this week. I will keep you posted with our results.

Issue #712, opened 2022-08-09
http module - buffer does not work?


grubberr:
Hello,

As for me `smart_open` http module can improve buffering, please look on code sample:

```python
import smart_open
import pandas as pd
import http.client as http_client

http_client.HTTPConnection.debuglevel = 1

fp = smart_open.open("https://github.com/airbytehq/airbyte/files/9280856/test.xlsx", mode="rb")
df = pd.read_excel(fp)
print(df)
```

```bash
$ ./test.py | grep airbytehq/airbyte/files/9280856/test.xlsx
send: b'GET /airbytehq/airbyte/files/9280856/test.xlsx HTTP/1.1\r\nHost: github.com\r\nUser-Agent: python-requests/2.28.1\r\nAccept-Encoding: identity\r\nAccept: */*\r\nConnection: keep-alive\r\n\r\n'
send: b'GET /airbytehq/airbyte/files/9280856/test.xlsx HTTP/1.1\r\nHost: github.com\r\nUser-Agent: python-requests/2.28.1\r\nAccept-Encoding: identity\r\nAccept: */*\r\nConnection: keep-alive\r\nrange: bytes=0-\r\n\r\n'
send: b'GET /airbytehq/airbyte/files/9280856/test.xlsx HTTP/1.1\r\nHost: github.com\r\nUser-Agent: python-requests/2.28.1\r\nAccept-Encoding: identity\r\nAccept: */*\r\nConnection: keep-alive\r\nrange: bytes=8685-\r\n\r\n'
send: b'GET /airbytehq/airbyte/files/9280856/test.xlsx HTTP/1.1\r\nHost: github.com\r\nUser-Agent: python-requests/2.28.1\r\nAccept-Encoding: identity\r\nAccept: */*\r\nConnection: keep-alive\r\nrange: bytes=8665-\r\n\r\n'
send: b'GET /airbytehq/airbyte/files/9280856/test.xlsx HTTP/1.1\r\nHost: github.com\r\nUser-Agent: python-requests/2.28.1\r\nAccept-Encoding: identity\r\nAccept: */*\r\nConnection: keep-alive\r\nrange: bytes=8045-\r\n\r\n'
[...]

PR #883, opened 2025-09-16
Support optional chunked reads from S3

### LINKED ISSUE #725: [s3] Ranged read http header leaves TO bytes blank, causing S3 filesystems to read full file length

batterseapower:
This commit adds a new `range_chunk_size` parameter to the S3 reader that allows reading large files in smaller chunks. This is useful when you only need to read small portions of large S3 files, as it prevents S3-compatible storage systems from queueing up the entire file internally.

Key changes:
- Added range_chunk_size parameter to smart_open.s3.open() and related classes
- Modified _SeekableRawReader to support chunked reading with proper boundary handling

When range_chunk_size is None (default), behavior is unchanged - single request for the whole file to minimize per-request costs on S3.

> Please **pick a concise, informative and complete title** for your PR.
> 
> The title is important because it will appear in [our change log](https://github.com/RaRe-Technologies/smart_open/blob/master/CHANGELOG.md).

### Motivation

> Please explain the motivation behind this PR.
> 
> If you're fixing a bug, link to the issue using a [supported keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) like "Fixes #{issue_number}".
> 
> If you're adding a new feature, then consider opening a ticket and discussing it with the maintainers before you actually do the hard work.

Fixes #725, uses an alternative approach to the stale PR #734

### Tests

> If you're fixing a bug, consider [test-driven development](https://en.wikipedia.org/wiki/Test-driven_development):
> 
> 1. Create a unit test that demonstrates the bug. The test should **fail**.
> 2. Implement your bug fix.
> 3. The test you created should now **pass**.
> 
> If you're implementing a new feature, include unit tests for it.
> 
> Make sure all existing unit tests pass.
> You can run them locally using:
> 
>     pytest tests
> 
> If there are any failures, please fix them before creating the PR (or mark it as WIP, see below).

- Added comprehensive test coverage including adversarial testing for retry logic
- Improved error handling for edge cases (negative offsets, empty files, etc.)

### Work in progress

> If you're still working on your PR, mark the PR as [draft PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request).
> 
> We'll skip reviewing it for the time being.
> 
> Once it's ready, mark the PR as "ready for review", and ping one of the maintainers (e.g. mpenkov).

### Checklist

> Before you mark the PR as "ready for review", please make sure you have:

- [x] Run `python update_helptext.py` in case there are API changes

### Workflow

> Please avoid rebasing and force-pushing to the branch of the PR once a review is in progress.
> 
> Rebasing can make your commits look a bit cleaner, but it also makes life more difficult from the reviewer, because they are no longer able to distinguish between code that has already been reviewed, and unreviewed code.

### DIFF
--- a/smart_open/s3.py
+++ b/smart_open/s3.py
@@ -675,7 +675,7 @@ class _SeekableRawReader(object):
         if self._body is None:
             # This is necessary to initialize _content_length for the very first read() after __init__().
             self._open_body()
-        
+
         if self._position >= self._content_length:
             return b''
 


ddelange:
Hi @batterseapower 👋 

Thanks for your efforts! 

Does this supercede https://github.com/piskvorky/smart_open/pull/734 and thus fix https://github.com/piskvorky/smart_open/issues/725? 

Please have a read, I wasn't a maintainer at the time but from this diff it looks like you're in pretty deep so I'd love to have a second opinion from you on the tradeoffs they're discussing.

ddelange:
fwiw, I'm not opposed to releasing smart_open v8.0.0 with a behaviour change here if we can simplify + improve the code as opposed to adding complexity to the already complex s3 module. 

e.g. the retry logic has improved in boto3 over the years and i feel we could simplify the s3 module a lot in a single sweep here
```py
client = boto3.session.Session().client("s3", config=botocore.client.Config(
    max_pool_connections=64,  # thread-safe client ref https://github.com/boto/boto3/blob/1.38.41/docs/source/guide/clients.rst?plain=1#L111
    tcp_keepalive=True,  # re-use across GET requests without too much overhead
    retries={"max_attempts": 6, "mode": "adaptive"}  # exponential backoff built-in nowadays
))  
```

it would be great if the entire `fp.read()` does a single GET regardless of current position, but anything else, like iterating over lines, can be chunked by default imo. boto3 also has [`iter_lines`](https://github.com/boto/botocore/blob/1.40.31/botocore/response.py#L150-L163) which is chunked by default (but on a single HTTP request iirc)

a related/adjacent issue is our buffering issue: https://github.com/piskvorky/smart_open/issues/712 which also calls for simplification

batterseapower:
> Does this supercede #734 and thus fix #725?

Yes, that's the idea!

I think #712 is out of scope for now - this would require some kind of more elaborate caching mechanism. I'm just trying to do the most minimal change that fixes #725.

> it would be great if the entire `fp.read()` does a single GET regardless of current position, but anything else, like iterating over lines, can be chunked by default imo. 

This wouldn't be hard to implement, but I'm wary about changing the default behaviour without a strong reason. It's hard to say whether switching to chunked by default for downstream users would break them in some way, given the diversity of use cases and s3-compatible APIs that could be in use.

smart_open/s3.py
```diff
@@ -605,31 +681,68 @@ def read(self, size=-1):
...
+                try:
+                    if size == -1:
+                        binary = self._body.read()
+                    else:
+                        binary = self._body.read(size - bytes_read)
+                except (
+                    ConnectionResetError,
+                    botocore.exceptions.BotoCoreError,
+                    urllib3.exceptions.HTTPError,
+                ) as err:
```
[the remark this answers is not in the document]
batterseapower: Possibly, yes. The current approach is fairly readable though. We aren't doing anything very complicated with e.g. exponential backoff so no strong reason to migrate to tenacity IMHO (though I like tenacity and use it quite a bit)

ddelange:
> Essentially analogous to multipart_upload=True which toggles between SinglepartWriter and MultipartWriter.

If these are going to be living next to each other, how about extending that analogy and refactor to have a `SinglerequestReader` + `MultirequestReader`, instead of the current Reader + _SeekableRawReader?

Just thinking here how to keep this module maintainable and readable.

smart_open/s3.py
```diff
@@ -605,31 +695,68 @@ def read(self, size=-1):
...
-                logger.warning(
-                    '%s: caught %r while reading %d bytes, sleeping %ds before retry',
-                    self,
-                    err,
-                    size,
-                    seconds,
-                )
-                time.sleep(seconds)
+        binary_collected = []
+        bytes_read = 0  # == sum(map(len, binary_collected))
```
ddelange, line 699: could use `binary_collected = bytearray()` to which you can efficiently `+=`, see also https://github.com/piskvorky/smart_open/pull/711

it also has a `len` so you can get rid of bytes_read

smart_open/s3.py
```diff
@@ -738,8 +737,7 @@ def read(self, size=-1):
 
             if binary:
                 self._position += len(binary)
-                binary_collected.append(binary)
-                bytes_read += len(binary)
+                binary_collected += binary
```
piskvorky, line 740: Not a fan. The previous version was more explicit & simpler IMO.

Is it really guaranteed `bytearray +=` is efficient and doesn't keep re-allocating internally?

piskvorky: So, logarithmic reallocations. Not as good as the original `join()` which is constant, but not terrible I guess.

ddelange: according to [this article](https://pythonspeed.com/articles/bytesio-reduce-memory-usage/), our best bet seems to be `BytesIO.get_value()`

smart_open/s3.py
```diff
@@ -605,31 +695,66 @@ def read(self, size=-1):
...
+
+            if binary:
                 self._position += len(binary)
-                return binary
+                binary_collected += binary
+            else:
+                # By construction we always try to read >= 1 bytes, so if we see this
+                # we know that we are at the end of the current _body.
+                # here just in case.
+                if self._chunk_end is None or is_fresh_chunk:
```
ddelange, line 745: both `_chunk_end` and `is_fresh_chunk` variables only exist for this check. judging by the code comments, this _might_ not be necessary at all.

maybe we can construct moto tests to prove that we can avoid this added complexity?

ddelange:
how about this simplification, eliminating `is_fresh_chunk` and `self._chunk_end` (also from `__init__`)?

- Replaces while True with explicit exit conditions
- Removes for-else construct
- Uses BytesIO instead of bytearray

```py
def read(self, size=-1):
    """Read from the continuous connection with the remote peer."""
    inf = float('inf')
    retry_errors = (
        ConnectionResetError,
        botocore.exceptions.BotoCoreError,
        urllib3.exceptions.HTTPError,
    )

    if size < -1:
        raise ValueError(f'size must be >= -1, got {size}')

    if size == -1:
        size = inf

    binary_collected = BytesIO()

    while binary_collected.tell() < size and self._position < self._content_length:
        if self._body is None:
            self._open_body()

        binary = None
        for seconds in [1, 2, 4, 8, 16]:
            try:
                if size == inf:
                    binary = self._body.read()
                else:
                    binary = self._body.read(size - binary_collected.tell())
                break  # success - exit retry loop
            except retry_errors as err:
                logger.warning(
                    '%s: caught %r while reading %d bytes, sleeping %ds before retry',
                    self,
                    err,
                    size,
                    seconds,
                )
                time.sleep(seconds)
                self._open_body()

        if binary is None:
            raise IOError('%s: failed to read after %d attempts' % (self, 5))

        if not binary:
            break  # end of stream - no more data available

        self._position += len(binary)
        binary_collected.write(binary)

    return binary_collected.getvalue()
```

batterseapower:
I don't think that CI failure is real ("The action 'Start vsftpd' has timed out after 2 minutes.")

Issue #518, opened 2020-07-05
Add type annotations to code


mpenkov:

DeflateAwning:
Specifically, the public-facing API is where type hinting would be most useful!

ddelange:
starting point would be typing.IO (base), typing.BytesIO and typing.TextIO (subclasses) ([cpython source](https://github.com/python/cpython/blob/v3.13.7/Lib/typing.py#L3460))

ddelange:
another working example is xopen ([src](https://github.com/pycompression/xopen/blob/v2.0.2/src/xopen/__init__.py#L747-L787))

ddelange:
planning to pair this with a linting/type-checking/ci polish

probably lean on uv/ruff/ty and pre-commit. this repo just went live and it uses uv check https://github.com/astral-sh/ty-pre-commit

PR #885, opened 2025-09-29
feat: add type hinting to `s3.open()` and main `open()`

DeflateAwning:
### Motivation

Related to #518. Not a complete fix, but makes some progress.

### Tests

CI should really get upgraded and have `ruff` for checking and `pyright` for type validations.


ddelange (commented):
> CI should really get upgraded and have ruff for checking and pyright for type validations

I would love an astral stack upgrade! They also built a type checker that I expect will supercede mypy/pyright over time: https://github.com/astral-sh/ty

Overall it would be great to have some CI to cover this PR before merging.

Big bonus would be to differentiate r/w/a return types (see docstring I mentioned in the issue, although I don't know a project that has implemented this properly).

ddelange:
superseded by #948, thanks for your effort @DeflateAwning!

DeflateAwning:
Wonderful, happy to see it complete one way or another!

PR #921, opened 2026-05-28
Add support for lz4 compression

muxiaofan:
### Motivation

A simple change that adds support for lz4 compression, since this format is quite popular.

### Verification
```
.venv/bin/python -m pytest tests/test_compression.py -q
........................                                                                                                                                                                                                                                                            [100%]
24 passed in 0.40s
```

### Checklist

> Before you mark the PR as "ready for review", please make sure you have:

- [x] Run `python update_helptext.py` in case there are API changes

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,7 +117,7 @@ You can install one or more of these dependencies explicitly using optional depe
 
 .. code-block:: sh
 
-    pip install 'smart_open[s3,gcs,azure,http,webhdfs,ssh,zst]'
+    pip install 'smart_open[s3,gcs,azure,http,webhdfs,ssh,zst,lz4]'
 
 Or, if you don't mind installing a large number of third party libraries, you can install all dependencies using:
 
@@ -238,6 +238,7 @@ The supported values for this parameter are:
 - ``.gz``
 - ``.xz``
 - ``.zst``
+- ``.lz4``
 
 By default, ``smart_open`` automatically (de)compresses the file if the filename ends with one of these extensions.
 `See also <https://github.com/piskvorky/smart_open/blob/master/smart_open/compression.py>`__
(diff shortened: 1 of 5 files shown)


tests/test_compression.py
```diff
@@ -57,3 +57,18 @@ def label(thing, name):
...
+
+@pytest.mark.parametrize(
+    'compression,filename',
+    [
+        ('infer_from_extension', 'file.lz4'),
+        ('infer_from_extension', 'file.LZ4'),
+        ('.lz4', 'file.lz4'),
+    ]
+)
+def test_compression_wrapper_read_lz4(compression, filename):
```
ddelange: this test is not needed as we already cover the mechanics using `test_compression_extensions` and `test_compression_wrapper_read`.

you can fold the permutations you added here into `test_compression_wrapper_read`

README.rst
```diff
@@ -238,6 +238,7 @@ The supported values for this parameter are:
 - ``.gz``
 - ``.xz``
 - ``.zst``
+- ``.lz4``
```
ddelange: can you move this up two lines analogous to help.txt?

tests/test_compression.py
muxiaofan: I kept it separate from `test_compression_wrapper_read` so that `pytest.importorskip` will only skip this test when `lz4` is not available. If you are okay with it, I can directly import `lz4` and merge this test case into `test_compression_wrapper_read` like you suggested.

ddelange: yes its fine, and no need for explicit lz4 import analogous to zstd:

- the `register_compressor` call registers lz4 unconditionally
- github CI installs `[all]` extras
- `test_compression_extensions` runs through all registered compressors unconditionally
- import of the compression lib happens at runtime inside the wrapper function

so you can just add lz4 to the parametrize of `test_compression_wrapper_read` just like the others and we should be good!

muxiaofan: both comments resolved. please take another look, thanks!

ddelange (approved):
many thanks!

